PR #794, opened 2015-01-11
fix #793 leaked labelCalcDiv used to calculate label dimensions

### LINKED ISSUE #793: Cytoscape Leaking DOM?
Each time a cytoscape instance is created and destroyed, there are remnant DOM nodes added to the body that are not removed. They seem to correspond to the last node in the graph, "George" in this case.

image

Here's the plunker: http://run.plnkr.co/plunks/4vKxA6Lqlqm7DxiuE8At/ and code is at http://plnkr.co/edit/4vKxA6Lqlqm7DxiuE8At?p=info

markov00:
Hi, this PR fix the leaked div reported on #793 that is used to calculate the label dimensions.
This div is appended to the body at the init and it's never removed or reused during the cy.destroy().

I'm not sure if this is the best way to solve this issue. Maybe it can be better to add the labelCalcDiv into the current cytoscape container (not into the body) and remove everything during the destroy changing the line https://github.com/cytoscape/cytoscape.js/blob/master/src/extensions/renderer.canvas.coord-ele-math.js#L659 with the following:

      this.data.container.appendChild( div );

### DIFF
--- a/src/extensions/renderer.canvas.define-and-init-etc.js
+++ b/src/extensions/renderer.canvas.define-and-init-etc.js
@@ -177,6 +177,10 @@
     if( this.removeObserver ){
       this.removeObserver.disconnect();
     }
+
+    if( this.labelCalcDiv ){
+      document.body.removeChild(this.labelCalcDiv);
+    }
   };
 
   


maxkfranz:
Nice; thanks

PR #829, opened 2015-02-11
Use gender-neutral pronouns in documentation

xc8tlik:
It's not immediately obvious from the diff formatting but this pull request just changes "his" to "their".


maxkfranz:
This is good.  My habit has been to use the rule that a person of unknown sex should be referred to as "he", but "they" would be better.

Unfortunately, your PR changes only the resultant HTML file rather than the source Markdown files from which it is generated.  So while your idea is a good one, I can't merge the pull request just now.  I'll see if I can change this if I have time, but if you want to do the find-replace across the `.md` files, that would be much appreciated!

maxkfranz:
Closing; I'll do this on the search files for 2.4.  Thanks

Issue #831, opened 2015-02-12
Add shadow properties


MissChocoe:
Canvas allow to draw shadow. Cytoscape could use that power. Proposed properties (which are the one available in the canvas, color being transferred to rgba to reflect opacity):

For node/edges:
shadow-blur
shadow-opacity
shadow-color
shadow-offset-x
shadow-offset-y

For labels:
text-shadow-blur
text-shadow-opacity
text-shadow-color
text-shadow-offset-x
text-shadow-offset-y

I have already this all done, just need to transfer to github and make a pull request.

maxkfranz:
It could be, but then again, I have avoided adding that because it doesn't add much utility over the border and text outline and it adds considerable rendering expense -- meaning that people who want to use large graphs wouldn't like it.

People tend to gravitate towards fancy styles like that even though they're often prohibitively expensive, so I've avoided adding blurs and shadows thus far.

Maybe expensive styles should be marked with the red asterisk or similar in the docs.  Most devs ought to avoid them, because performance is so often a high concern.

maxkfranz:
Merged; looking into issue you noted now

MissChocoe:
@maxkfranz Thanks for your fix!

maxkfranz:
Closing for now; let me know if there is anything else.  Thanks again!

MissChocoe:
@maxkfranz  My pleasure! When do you plan on close v2.4? I'll try to download the latest changes to do some testing before the deadline.

maxkfranz:
ASAP so some issues may have to be postponed to a later milestone...

PR #833, opened 2015-02-12
#831 Added shadow properties for node, edge and text

MissChocoe:
I've also updated the documentation with a note on performance issue when using shadow-blur.

This is my first contribution to this project so any comments are welcome.


MissChocoe:
There is one issue that I didn't know how to fix: when zooming out, the shadow becomes bigger.

PR #835, opened 2015-02-15
Added feature to rotated label: now it can be centered to line and have ...

ktei:
@maxkfranz Our project recently required a feature that the rotated labels have a filled background. I basically implemented it and wanted to contribute it back to cytoscape.js. So previously we have autorotate label, but now I added another option autorotate-with-background. It looks like this once it's turned on:
capture

Let me know how you think of this. Feel free to merge it if you want to.


maxkfranz:
This looks really nice, but maybe it would be better to work from #838 instead as that is very similar but has more options over styling?  Thanks

ktei:
@maxkfranz Yeah, I'll do that.

maxkfranz:
Closing because of separate PR

PR #837, opened 2015-02-16
Log correct variable in map example.

skasio:
Hi - this updates the the example for `eles.map()` so that it logs the defined variable.

Regards,

Szymon


maxkfranz:
Thanks

Issue #825, opened 2015-02-09
Allow to add label background color


MissChocoe:
This could be useful if some labels are overlapping, label with an opaque background color would be easier to read.

maxkfranz:
That makes sense.  Proposed properties:

* `text-background-color` : With the specified colour, places a rectangular background on the label text (no background is used by default or when specified with the value `none`).
* `text-background-opacity` : Specifies the opacity of the label text background colour (fully opaque with a value of `1` by default).

Note that this will have performance concerns as it will increase the number of shapes drawn per element.  However, the expense can be minimised by using sqaure rectangles rather than ones with rounded corners.  What do you think?  If rounded corners are a requirement, then it may merit an additional property.

MissChocoe:
To reduce performance issue, you could put the opacity at 0 by default and then not draw the shape in that case. And I'm sure at some point someone will want rounded corner ;) Might as well put it as an option (but used regular rectangle by default). Properties name make sense.

MissChocoe:
I can actually implement it. :)

MissChocoe:
I've created a pull request with this feature. I decided to let go of the round rectangle to simplify rendering of all that but added the border in addition to background color. So here's the full list of new properties: 

 * **`text-background-color`** : A color to apply on the text background, may be `none` or a valid color.
 * **`text-background-opacity`** : The opacity of the label background.
 * **`text-border-width`** : The border width to put around the label.
 * **`text-border-style`** : The style of the border around the label; may be `solid`, `dotted`, `dashed`, or `double`.
 * **`text-border-color`** : The color of the border around the label.

MissChocoe:
One more property that could be added is text-margin. For now, I've fixed it to 4px so the text is not glue to the border. Maybe we should let user control that.

jonorthwash:
Is this what `"text-background-opacity": 1, "text-background-color": "#ffffff"` is supposed to look like?
![2017-11-19-20 44 50_001](https://user-images.githubusercontent.com/963849/32998595-8d55b51e-cd6a-11e7-961e-9aa156087e50.png)
I'm happy to file a bug report if something is wrong, but that seems to be heavily discouraged if there's any chance I'm just being being stupid, so I decided to assume I was stupid this time.

If the answer is just "ask on stackexchange", that's fine too.
&lt;/stupidity&gt;

maxkfranz:
@jonorthwash It's working fine for me on master (3.2.6) with those two properties:

[...]

PR #849, opened 2015-03-06
added a bit code based on commit #825 so that the rotated label text looks bet...

ktei:
@maxkfranz I checked out #825 and it seems the author had implemented almost what I had done and also did much more. So I just changed a bit based on her work to make the text background look better when label text is rotated. Can you have a look?

Cheers

### DIFF
--- a/src/extensions/renderer.canvas.drawing-label-text.js
+++ b/src/extensions/renderer.canvas.drawing-label-text.js
@@ -184,6 +184,22 @@
     return text;
   };
 
+  function roundRect(ctx, x, y, width, height, radius) {
+    var radius = radius || 5;
+    ctx.beginPath();
+    ctx.moveTo(x + radius, y);
+    ctx.lineTo(x + width - radius, y);
+    ctx.quadraticCurveTo(x + width, y, x + width, y + radius);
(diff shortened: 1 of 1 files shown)


maxkfranz:
Thanks!  It looks really good.  Do you think it makes sense to centre the labels on edges (as it is with the background) all the time?  Or is there a different usecase that I'm not thinking of?

ktei:
@maxkfranz I think maybe let's just make it easy:
If the 'autorotate' is switched on, then we always center the labels on edges. It is true that however if the user does not declare any background, the labels will be strikethrough be edges, but I think most of the time, people will want background.

MissChocoe:
@ktei That would be the best option. text-border-radius would be the property to addd to do so

maxkfranz:
Maybe a `shape: rectangle | roundrectangle` would be best.  In future, if we allow click events on labels, having a large border radius would complicate the hit tests -- whereas if the `roundrectangle` case has a small, unchangeable border radius then we can assume rectangles for the hit tests.

Thoughts?

MissChocoe:
@maxkfranz totally agree with you, shape property would be enough. 

By the way, I'm working with the unstable version and it seems to go smootly!

PR #863, opened 2015-03-11
#849 Added a text-background-shape property

MissChocoe:
So rounded corners are now optional for the label background. My editor removed some spaces on empty lines, sorry about that!


maxkfranz:
Thanks; and I don't really care about the formatting too much as long as things are human readable.

PR #961, opened 2015-06-03
Add length property to springy edge

baso10:
https://github.com/dhotson/springy/issues/39
The length can be set on edge objects. ie edge.data.length
If the length property exists it will be used in the physics calculations (default length is 1).


maxkfranz:
Great idea, but it should probably be functional instead of assuming a particular data field, i.e.

```js
length: options.edgeLength.call( edge, edge )
```

baso10:
I have included also function call, but kept simple data field. In my case I know length when I am creating edge and there is no need for function calls.

PR #980, opened 2015-06-18
Fix canvas resize issue

megawebmaster:
This issue is calling resize() on cytoscape object when container was
resized. Due to how browsers works with this kind of changes we need to
add a milisecond timeout to properly get container size.

Source:
http://stackoverflow.com/questions/2643396/get-width-of-element-after-resizing-it

Real change is lines 100-1005, all the other stuff is automatic cleaning of trailing whitespaces (I hope you don't mind ;) )

### DIFF
--- a/src/extensions/renderer.canvas.drawing-redraw.js
+++ b/src/extensions/renderer.canvas.drawing-redraw.js
@@ -6,7 +6,7 @@
 
   // var isFirefox = typeof InstallTrigger !== 'undefined';
 
-  CRp.getPixelRatio = function(){ 
+  CRp.getPixelRatio = function(){
     var context = this.data.contexts[0];
 
     if( this.forcedPixelRatio != null ){
@@ -55,7 +55,7 @@
 
   CRp.fillStyle = function(context, r, g, b, a){
     context.fillStyle = 'rgba(' + r + ',' + g + ',' + b + ',' + a + ')';
-    
+
     // turn off for now, seems context does its own caching
 
     // var cache = this.paintCache(context);
@@ -69,7 +69,7 @@
 
   CRp.strokeStyle = function(context, r, g, b, a){
     context.strokeStyle = 'rgba(' + r + ',' + g + ',' + b + ',' + a + ')';
-    
+
     // turn off for now, seems context does its own caching
 
     // var cache = this.paintCache(context);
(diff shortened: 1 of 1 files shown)


maxkfranz:
Do you have a testcase that has issues with the current setup?

The PR is much appreciated, but a timeout may not work here.  Calls to `.matchCanvasSize()` need to be guaranteed to finish before subsequent calls to `.redraw()` -- else `.canvasWidth` etc will be incorrect during the redraw.

I think the case may be a bit different here with respect to reading the CSS values.  Usually, reads can be made OK at any time.  But writes can invalidate subsequent reads -- or cause multiple reflows in the HTML model.  Since we read at the start and then only write after that, I think we should be OK as-is.  Does that make sense?

megawebmaster:
Hi,

It might be weird, but it seems everything is working fine even without this change, I have updated Cytoscape thrice (2.4.2, then master, then master with my fix and then again 2.4.2) and it began to work properly. I'm sorry to bother you.

P.S. I'm sure yesterday I had this bug and this was the way to solve it!

megawebmaster:
It seems the problem is back. Some images:

Opened column on the left side:
open_bf

Closed before fix:
closed_bf

Closed after fix:
closed_af

As you can see before fix images are chopped of, BUT the graph is operable in that "blind spot" (i.e. I can pan the graph). I tried to create a codepen for this, but Cytoscape worked, although without showing a thing (canvases were right there). It might be the issue with iframes though.

As you said about `redraw()` function I changed pull request to add safer solution.

maxkfranz:
Have you tried calling `cy.resize()` *after* you change the cy div?

You could put the call to `cy.resize()` in a timeout.

megawebmaster:
`cy.resize()` is called after resizing of the div, but the thing is browser itself is doing width recalculation. I even looked through the code and tried to see what's going on with width and height of the cy div, but it really was lower when calling to `resize()`. Also I think putting `resize()` call in a timeout is a bit dirty workaround. 

Just curious - why do you want to avoid this 1ms delay in calling `resize()` function?

maxkfranz:
The delay should not be on `cy.resize()`, because it removes control from the developer.  The call would be async such that the dev could never guarantee the order of events.

It's far better for the dev to be able to specify a sync or async `cy.resize()` by making that call either directly or with his own `setTimeout()`.

PR #982, opened 2015-06-19
Fix canvas resize issue

megawebmaster:
This issue is calling resize() on cytoscape object when container was
resized. Due to how browsers works with this kind of changes we need to
add a milisecond timeout to properly get container size.

Source:
http://stackoverflow.com/questions/2643396/get-width-of-element-after-resizing-it

Some images:

Opened column on the left side:
open_bf

Closed before fix:
closed_bf

Closed after fix:
closed_af

As you can see before fix images are chopped of, BUT the graph is operable in that "blind spot" (i.e. I can pan the graph). I tried to create a codepen for this, but Cytoscape worked, although without showing a thing (canvases were right there). It might be the issue with iframes though.

Safer solution to: https://github.com/cytoscape/cytoscape.js/pull/980

### DIFF
--- a/src/extensions/renderer.canvas.define-and-init-etc.js
+++ b/src/extensions/renderer.canvas.define-and-init-etc.js
@@ -18,54 +18,54 @@
   CanvasRenderer.MOTIONBLUR_BUFFER_NODE = 1;
   CanvasRenderer.MOTIONBLUR_BUFFER_DRAG = 2;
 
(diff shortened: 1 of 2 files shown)


maxkfranz:
Closing for now re. https://github.com/cytoscape/cytoscape.js/pull/980#issuecomment-114576137

PR #1008, opened 2015-07-13
Grid layout : condensed & padded

sylvain-hamel:
I have an issue with the grid mode. I use it with `columns : 1, fit: true`. If I only have 2-3 nodes and the canvas is 700px height then I get too much spacing between the nodes and the graph become hard to read. This PR adds a feature to compute the height of the highest cell and then use this value as the *row* height.

What do you think of the idea? 
Should I do the same for columns  (using `autoCellWidth`)?
Do you like the name of the option? Can you think of a better name?

### DIFF
--- a/src/extensions/layout.grid.js
+++ b/src/extensions/layout.grid.js
@@ -118,15 +118,30 @@
       var cellWidth = bb.w / cols;
       var cellHeight = bb.h / rows;
 
-      // cellHeight fits the tallest node
-      if (options.autoCellHeight) {
-        var maxHeight = 0;
+      // option to set all cell sizes to the size of the biggest cell
+      if (options.autoCellSize) {
+        cellWidth = 0;
+        cellHeight = 0;
+        var cellPaddingX = 20;
+        var cellPaddingY = 20;
+
         for (var i = 0; i < nodes.length; i++) {
           var node = nodes[i];
-          var h = node.boundingBox({ includeNodes: true, includeLabels: true }).h;
-          maxHeight = Math.max(maxHeight, h);
+
+          // During position calculation, requesting for bouding box will fail when node does not have position, fix 0, 0
+          if (!node._private.position.x) {
+            node._private.position.x = 0;
+          }
+
+          if (!node._private.position.y) {
+            node._private.position.y = 0;
+          }
+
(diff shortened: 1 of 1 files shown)


sylvain-hamel:
I did not add any padding. Maybe `10px` of padding would be better. Would you see the padding as an hardcoded value or do you think an option is necessary for that?

maxkfranz:
There's a later call to `cy.fit( padding )`, so I don't think that's an issue.  A few things, though:

(1) All new features should be PRs on `unstable`.  `master` only gets fixes for minor releases.

(2) Doesn't this work only with one column?

(3) Couldn't you use the existing options, like `boundingBox`, to make the nodes closer together?  A smaller `boundingBox` will put the nodes closer together and zoomed in more.

(4) The layouts all try to use up all the available space.  This depends on screen size and your style.  It may also help to make your nodes etc bigger in your stylesheet.

sylvain-hamel:
Hi,

(1) Sorry about the wrong branch. This is only to get the discussion started. I'll rebase and add the doc if we get to that point.

(2) No. If you have several nodes and one of the node is really high and all the others are short. Then the cell high of all rows and all columns will be equal to the highest.

pic0


(3) `boundingBox` does not prevent overlap :
pic1

(4) In some cases using up all the available space is not optimal. For instance the grid layout produces this:

pic3

But I think that this is more readable:

pic4


----

What do you think or all this? Will you accept this feature if I clean up the PR?

sylvain-hamel:
@maxkfranz I have improved the code a little and I have renamed the option to `autoCellSize` because it now supports both `height` and `width`. Do you plan on accepting this PR. Otherwise we are going to fork the `grid` layout to make our own.
Thanks

maxkfranz:
This PR still isn't on `unstable` so I can't merge.

That aside, I've made a separate commit that doesn't conflict with the previous options and has more modular options.  Please review.

sylvain-hamel:
Hi, I like the result and I like how it extends the current option instead of creating a new option. Next time I'll send PRs on `unstable`. Thanks

maxkfranz:
OK; closing for now

Let me know if anything else comes up regarding this

Thanks!

PR #1071, opened 2015-09-03
When attempting to register JQuery, check for jQuery property on window

areel:
When attempting to register JQuery, check for window property called "jQuery" rather than "jquery".

I believe this is the correct fix for the error I receive on latest version.
line 579 in cytoscape.js fails with "require not defined error", I believe we get here because
line 550 window[name] is undefined when name contains "jquery".


maxkfranz:
The call before the one you modified checks `jQuery` on `window`.  The second check is for CommonJS and AMD systems.

areel:
­I am returning to this codebase after some months, it was functional at that stage. 

It is JSPM based.

I updated the applications dependencies. 

A lot of new versions (cytoscape included) was pulled.

So the error I am seeing raised within cytoscape could well be due to external structural changes.

 

What I am seeing is that $ and “jQuery” are defined on window not “jquery”.

 

The code you referred checks for $ (as far as I can see) which would succeed and return, execution then continues to run the code that I changed which checks on window for jquery and fails so it continues on to look for AMD and CommonJS, and is forced to examine require on the last check which is undefined and generates the error I mentioned.

 

Regards

 

Aidan

maxkfranz:
The first call checks `window.jQuery`.  The second checks the *package* with name `jquery` via `require()`.  With the external register API now available, it would be simplest to remove the second check: https://github.com/cytoscape/cytoscape.js/issues/1073

PR #1123, opened 2015-10-13
Fix infinite loop in some JavaScript implementations

mmillman:
Fixes infinite for...in loop that happens with some JavaScript implementations, caused by adding keys to an object while iterating over its keys.

I found two working fixes. (I went with the first.):

1. Change exprs to an array.
I went with this one since it looked like exprs was being modified to be used like an array immediately after initialization anyway.

2. Don't modify exprs, and instead, build up a temp object while iterating over exprs, and then after the iteration is complete, assign exprs to that temp object.
This solution would be fitting if there is a reason to keep exprs as an object.

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -133,8 +133,9 @@
       // - the current query is stored in self[i] --- you can use the reference to `this` in the populate function;
       // - you need to check the query objects in Selector.filter() for it actually filter properly, but that's pretty straight forward
       // - when you add something here, also add to Selector.toString()
-      var exprs = {
-        group: {
+      var exprs = [
+        {
+          name: group,
           query: true,
           regex: '(node|edge|\\*)',
           populate: function( group ){
@@ -142,7 +143,8 @@
           }
         },
 
-        state: {
+        {
+          name: state,
           query: true,
           // NB: if one colon selector is a substring of another from its start, place the longer one first
           // e.g. :foobar|:foo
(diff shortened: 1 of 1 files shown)


maxkfranz:
Thanks; the object was probably used in a different way before and wasn't refactored

I've made a separate commit, because the PR fails under Travis tests.

Snapshot : http://js.cytoscape.org/download/cytoscape.js-snapshot-6bac8d5c5e-1444757559523.zip

mmillman:
Great, thank you!

Also, just a small thing, I noticed you left out the "var j = 0", so I just wanted to note that without it, there may be accidental assignment to the global object in future refactoring, since there's a for-loop in that function that does not redeclare j, and just does "for (j...)". It currently does not assign to the global object because of other for-loops declaring j with "for (var j...)" and hoisting, but if all the for-loops were converted to the non-redeclaring style, that may cause trouble down the road. Since the (re)declaration style is inconsistent, that predeclaration can act like a "style safeguard".

mmillman:
And regarding the tests, they are failing locally as well.

maxkfranz:
Tests seem to be OK on Travis: https://travis-ci.org/cytoscape/cytoscape.js/builds/85168207

maxkfranz:
Added `var`; better to have it as may use babel w/ `let` in future

mmillman:
Ah, whoops, tests were failing because I missed the quotes around the 'name' values in the exprs object in my commit. Looks like you didn't copy over that error, which explains the earlier Travis test failure and current test pass.

PR #1155, opened 2015-11-10
Fix for Thread fnAs breaking minified functions

yurynix:
Changed regex to cover the minified functions case.

https://github.com/cytoscape/cytoscape.js/blob/e75a707415564d5a20928b6ae480cfb616a2e29b/src/thread.js#L425

Consider the case where:
```
fn === "function cellCentroid(e){for(var n,t,o,i=e.halfedges,a=0,r=0,s=0,c=0;c<i.length;++c)n=i[c].getEndpoint(),t=i[c].getStartpoint(),a+=n.x*t.y,a-=n.y*t.x,o=n.x*t.y-t.x*n.y,r+=(n.x+t.x)*o,s+=(n.y+t.y)*o;return a/=2,o=6*a,{x:r/o,y:s/o}}"

name === "cellCentroid"
```
( Code from spread layout )

The result is a broken with syntax error code:
```
"function cellCentroid(var n,t,o,i=e.halfedges,a=0,r=0,s=0,c=0;c<i.length;++c)n=i[c].getEndpoint(),t=i[c].getStartpoint(),a+=n.x*t.y,a-=n.y*t.x,o=n.x*t.y-t.x*n.y,r+=(n.x+t.x)*o,s+=(n.y+t.y)*o;return a/=2,o=6*a,{x:r/o,y:s/o}}"
```

Because what matched on the regular expression is 'function cellCentroid(e){for('

Maybe change the regex to something like:
``/function[\s\S]+?\(/`` - match at least one space/no-space char in a non-greedy way?

It fixes it for that case and some others I've tested.

### DIFF
--- a/src/thread.js
+++ b/src/thread.js
@@ -422,7 +422,7 @@ util.extend(thdfn, {
 // turns a stringified function into a (re)named function
 var fnAs = function( fn, name ){
   var fnStr = fn.toString();
-  fnStr = fnStr.replace(/function\s*\S*\s*\(/, 'function ' + name + '(');
+  fnStr = fnStr.replace(/function[\s\S]+?\(/, 'function ' + name + '(');
 
   return fnStr;
 };


yurynix:
The error shown in console when using cytoscape (2.5.0) with spread layout is:
Uncaught SyntaxError: Unexpected token var 
...and some blob url with random GUID.

maxkfranz:
I'll take a look at this [Weaver](https://github.com/maxkfranz/weaver) issue when I'm back and have regular online access on the week of the 24th

maxkfranz:
Weaver has been updated with a fix (non-greedy match).  The updated code on weaver/master has been pulled into cytoscape via `gulp weaver`.

Snapshot: http://js.cytoscape.org/download/cytoscape.js-snapshot-a3fcbe2638-1448481076888.zip

yurynix:
Tnx! =)

Issue #789, opened 2015-01-06
Fixed element size (nodes,edges,font) regardless of zoom


markov00:
Is there a way to specify a fixed size for each element regardless of the zoom? on #780 I've found a way to have a fixed node size that you can test [here](http://jsbin.com/hapasuyoni/2/edit?js,output): 
but it has some problems:
- the labels are redrawn only when you are selecting a node or an edge
- the arrow size depend on the initial edge width, but if this width change the edge arrow doesn't change its dimension
- edges curves: we can't control the bezier control point so if the edges are relatively very near each other, the curve is very big.
  I'm asking if there is a different way to create a graph with fixed element size regardless of the zoom level. 
  Think of a big graph that you want to explore like the [tokyo railways demo](http://jsbin.com/gist/f64e811fc3311414e083?js,output), if you zoom in  everything becomes overlapped and unreadable. I think this feature can be very useful.

markov00:
I absolutely agree: my approach was just a workaround, but implementing this at renderer level will be much more efficient and easier for developers.

ktei:
Bump up this. I currently set font-size and node radius based on the count of visible nodes, because when there are only a few nodes, without any adjustment, all nodes and labels become very very big. I can work around with this by tweaking zoom level and style dynamically.

It'd be nice to see that: if a max node radius is specified, then regardless of how many nodes are there, their size won't go beyond the max node radius (which potentially means that the cytoscape would choose a proper zoom level for the graph when it's loaded...?)

Also, it'd be nice to see that: if a max text size is specified, then regardless of how many nodes are there, all the text's size won't go beyond the max value?

I fully understand that at the moment, cytoscape is just trying to fully fill the canvas on loading, but it doesn't look very good when you just have a handful nodes, but each one of them is so large with very big font size.

Thoughts?

maxkfranz:
@ktei For your case, just don't use the `fit` option.  You can instead call `cy.zoom( yourPreferredZoom )` and `cy.center()`.

maxkfranz:
This would actually require some other model on the viewport, like a stretch rather than a zoom.

swtyler:
@maxkfranz Any updates on this? Do you have any suggestions for fixed size labels? Here is an example scenario:

[...]

Issue #1140, opened 2015-10-28
Expose zoom gesture events


maxkfranz:
Expose mouse wheel and pinch events such that external extensions can replace zoom behaviour while using the same gestures and normalised values

Joseph-Gage:
@maxkfranz Was this ever implemented? I cannot find it in the docs.

PR #1258, opened 2016-02-19
Extension point in preparation for issue #789

janblok:
In order to make an extension possible as discussed in issue #789, its needed to access or override the zoom factor. It also allows for changing the constants used there.

### DIFF
--- a/src/extensions/renderer/base/coord-ele-math.js
+++ b/src/extensions/renderer/base/coord-ele-math.js
@@ -31,17 +31,29 @@ BRp.invalidateContainerClientCoordsCache = function(){
   this.containerBB = null;
 };
 
+BRp.getEdgeThreshold = function(isTouch, zoom) {
+	return (isTouch ? 24 : 8) / zoom;
+}
+
+BRp.getNodeThreshold = function(isTouch, zoom) {
+	return (isTouch ? 8 : 2) / zoom;
+}
+
+BRp.getLabelThreshold = function(isTouch, zoom) {
+	return (isTouch ? 8 : 2) / zoom;
+}
+
 // Find nearest element
 BRp.findNearestElement = function(x, y, visibleElementsOnly, isTouch){
   var self = this;
   var r = this;
   var eles = r.getCachedZSortedEles();
   var near = [];
-  var zoom = r.cy.zoom();
   var hasCompounds = r.cy.hasCompoundNodes();
-  var edgeThreshold = (isTouch ? 24 : 8) / zoom;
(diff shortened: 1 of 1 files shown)


janblok:
An example of fixed size node/edge extension:
[cytoscape-fixed-element-size-extension.js.txt](https://github.com/cytoscape/cytoscape.js/files/138117/cytoscape-fixed-element-size-extension.js.txt)

maxkfranz:
That looks like it's referencing a lot of internal, renderer-private data/APIs.  All of that is private and subject to change at any time.  It wouldn't work well as an extension.

Fixed element sizing doesn't make sense, because a zoom operation is a scale operation.

As mentioned in the issue towards the end, it would make sense to have fixed size elements only if you replace the zoom with your own logic.  This means that the extension doesn't use the Cytoscape.js internals for zoom and pan -- it would have to rewrite them from scratch and disable the built-in zoom and pan features in Cytoscape.js.

janblok:
It only overrides drawNodeText, drawNode, drawEdge and in order to do so properly, these needs the suggested minor pull request code change. 
The change containing getLabelThreshold, getNodeThreshold and getEdgeThreshold makes it possible to make "thresholds" based on other constants and different zoom factor.

Without this pull request it's impossible to make a proper working fixed scale solution today...

maxkfranz:
This PR isn't necessary and as stated earlier it is predicated on modifying private data.  It doesn't matter how much private data you modify -- using a little is just as bad as using everything.  If you use private data, your code will break in future versions.  I *guarantee* it would be broken in 2.7.

If you want the simple implementation, use what's been mentioned earlier with modifying style on zoom in the other issue.

If you want a more complex implementation, write a more complex extension.  Disable native zooming.  Implement a stretch feature that changes node positions on gestures normally used for zooming, like wheel and pinch.

janblok:
We can definitely have a look at modding the style during zoom, but the pull request is for the threshold!
And the threshold is not part of the style or anything else...and without modding the threshold nothing workable is possible today as you indicated.

maxkfranz:
As I said before, it matters that you're relying on changing/overriding *private* renderer functions/data -- not how much data or which particular data.  You linked to example code that overrides many functions in the renderer.  None of the renderer is public, none of it is documented publicly, and none of it is guaranteed to remain consistent between versions.  

I've already described to you two good solutions for implementing this today.

If you want the second, more performant solution without rewriting/copying pinch and wheel gestures in your extension, then consider a PR for #1140, as noted in #789.

PR #1263, opened 2016-02-22
Fix bug preventing edge label wrap

jaschaephraim:


maxkfranz:
Great!  It may be better to check `== null`, as the value might be set to `null` if it's invalidated.

I'll be able to merge this in within the next few days.  If you could update the check in the meanwhile, that would be great.  Thanks!

jaschaephraim:
In the case I checked it was `undefined`, so how about checking for falsy:

```javascript
if( rscratch.labelWrapKey && rscratch.labelWrapKey === rscratch.labelKey ){
```

maxkfranz:
Perfect

PR #1265, opened 2016-02-24
Fix typos

pra85:
Remove extra `the`
`posisition` → `position` (4 instances)

### DIFF
--- a/documentation/docmaker.json
+++ b/documentation/docmaker.json
@@ -1460,14 +1460,14 @@
                 },
 
                 {
-                  "descr": "Get the value of a specified rendered posisition dimension.",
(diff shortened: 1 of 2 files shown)


maxkfranz:
Thanks for catching these typos!

Would you be able to modify the source files for the docs instead?  The documentation gets built, and you've modified the file that comes out of the build rather than the sources.

The documentation is specified in the `docmaker.json` file and the linked `.md` files.  Thanks

pra85:
@maxkfranz Apologies, didn't know that `index.html` was built. I have updated the PR to do changes in the sources

maxkfranz:
Merged in.  Thanks!

PR #1269, opened 2016-02-25
Fix taphold and cxttap problems under Chrome for Windows.

AlexJF:
# Changes
Use desktopTapThreshold for all mouse drag functionality.

# Details
Related external bug: https://bugs.chromium.org/p/chromium/issues/detail?id=161464

Because of the aforementioned bug in Chromium (which apparently only affects Windows users), all events that rely on there not being a `mousemove` event, such as `taphold` and `cxttap` (these were the only 2 I tested), fail to trigger.

As pointed out in the external bug report, a simple solution and good practice when dealing with these click + move interactions is to only consider a drag after a certain movement threshold. Such a threshold already exists (`desktopTapThreshold`) but is only applied when testing dragging of nodes. Because cancellation of holds and right taps occurs outside of this test, the problem manifests.

# Comments
 For the fix, testing against a threshold of 0 (`mousemove` event with no actual movement) would be sufficient but, in my opinion, the `desktopTapThreshold` should be applied to any type of mouse drag movement. Most importantly, doing so, also applies `desktopTapThreshold` to right-click taps.

I haven't tested the touch-related events so perhaps they suffer from the same problem.

### DIFF
--- a/src/extensions/renderer/base/load-listeners.js
+++ b/src/extensions/renderer/base/load-listeners.js
@@ -480,7 +480,11 @@ BRp.load = function() {
 
     var multSelKeyDown = isMultSelKeyDown( e );
 
-    r.hoverData.tapholdCancelled = true;
+    var isOverThresholdDrag = rdist2 >= r.desktopTapThreshold2;
+
+    if (isOverThresholdDrag) {
+      r.hoverData.tapholdCancelled = true;
+    }
 
     var updateDragDelta = function(){
       var dragDelta = r.hoverData.dragDelta = r.hoverData.dragDelta || [];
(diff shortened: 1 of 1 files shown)


maxkfranz:
Thanks for the heads up!

Unfortunately, I don't think this can be accepted as-is.  The check for the distance threshold wraps a lot of other checks and state continuation logic.  Instead of preventing all logic, the threshold should only prevent the logic for *starting* a gesture -- like `cxtdrag`.  This should probably happen on a case-by-case basis rather than a wrapped overall strategy.

AlexJF:
Changed it to a more conservative approach. Retested, cxttap and taphold work as expected as do all drags I tested.

maxkfranz:
Great!  I'll review/test and look at porting this to touch on Monday

maxkfranz:
This looks great; I'll merge this in and review the touch case to make sure it's consistent with this update

PR #1387, opened 2016-05-25
Add new option zorderStrict to allow for overriding default nodes atop edges behavior

mbeynon:
I need strict zordering because I'm annotating a base network diagram with high level flows and cannot have the high level flow arrows drawn under the nodes of the graph.

The default value is false, which is backwards compatible with all existing uses.  The option is read in the collection comparator to enforce the ordering.

gulp test runs clean.

Mike.


maxkfranz:
@mbeynon 

A few things:

(1) Having a global state to determine styling behaviour will conflict with cached style calculations, and it adds coupling --- all styling goes in the stylesheet.

(2) Having a global option precludes using the nodes-on-top logic on a per-element basis.

(3) You've modified the files in the `dist` folder.  Those aren't the source files.  They won't get tested by mocha, and they'll just get overridden whenever a build is made.

(4) New features should go in `unstable`.  New features do not get added to patch releases (`master`).

I don't think an option like this is practical because of (1) and (2).  The only way I see this working is a new style property for edges like `z-order-auto-below-nodes: yes | no`.  Though a property like that should probably be more generic, because there's also compound node auto ordering to consider.

mbeynon:
Appreciate the feedback.  I'm motivated to find an agreeable approach to avoid living on a fork.

I don't consider the global option as being styling.  Rather it controls the hard-coded nodes-on-top logic heuristic deep in the collection sort code.  Elements already have a `z-index` style that should affect when it's drawn, but it's ignored in some cases.  That behavior seems wrong.  Fixing it by adding another style to actually use the existing style feels redundant.  My global option was meant to support backward compatibility while changing the logic to be more in line with what I expect `z-index` means.

How does the global option conflicts with cached style calculations?  If the global option is set at init, then all cached style calculations are based on the collection zsort order.  Don't see how that's any different to what you have now -- cached style calculations and a particular collection zsort ordering.  All this PR does is change the arbitrary collection ordering.  I must be missing something.

The current logic ignores `z-index` unless the elements are in the same depth of the compound node tree structure.  Sort of breaking ties.  The odd part is this happens not just for the same compound node, but any that happen to be at the same compound node depth.  It that intentional?

Here's what I would expect:
1.  If no `z-index` is provided, then primary sort is compound node depth (or perhaps this should be the bfs total order?) and secondary sort is nodes on top of edges.
2.  If a `z-index` is provided for a compound node, then all children inherit it by adding it to their potential `z-index` value.  (this may be too complicated to implement or reason about.)
3.  If a `z-index` is provided for a node or edge, then use it.  For ties, use nodes on top of edges.

Thoughts?

Note that I updated the dist files because I found it required when using `npm install` to the github url of my forked version.  Appears to only copy from dist on install.  Happy to do another way if you have a better suggestion.

maxkfranz:
I think the conflict with the caching may be only in `unstable` so you wouldn't have noticed it in `master`.  In any case, this feature ought to be a style property if it affects the visual styling of elements.

A `z-index` must be relative to a parent for compound nodes, and this must be automatic -- else compound nodes would be near unusable:  You'd have to manually specify a `z-index` perfectly for every descendent and descendent-touching edge.  

Consider someone who has a usecase just like yours but in a compound graph.  They couldn't do it reasonably with a global option.

`z-index` in Cy.js behaves just like it does in HTML/CSS:  The value is relative to the parent, and so you can only meaningful comparison of the values is with siblings.  Setting a second CSS property with a different value allows for an element to be considered an orphan w.r.t. `z-index`.

CSS has an implicit way to specify whether a z-index is relative to the parent or not, and it's a bit inflexible.  We would do better to be explicit about it with a property like `z-order-auto-below-nodes: yes | no` or perhaps `z-index-compare: auto | orphan | someOtherFutureValue ...`

maxkfranz:
>The current logic ignores z-index unless the elements are in the same depth of the compound node tree structure. Sort of breaking ties. The odd part is this happens not just for the same compound node, but any that happen to be at the same compound node depth. It that intentional?

Because `z-index` only matters between siblings, the comparison result doesn't matter.  This keeps the calculation cheaper and faster.

maxkfranz:
> Note that I updated the dist files because I found it required when using npm install to the github url of my forked version. Appears to only copy from dist on install. Happy to do another way if you have a better suggestion.

If you use `unstable`, the `main` in `package.json` is `index.js` rather than `dist/cytoscape.js`.  No `dist` needed for npm.

We'll be moving `dist` (for bower etc.) to another repo soon to reduce the source repo size.  The `dist` was previously used in `package.json` just for consistency.

mbeynon:
My use case is actually a compound graph that is used to visualize a network diagram.  Subnets are compound nodes that contain hosts.  Network devices are the routers and firewalls that connect the subnets.

To make it clear what I'm trying to do, here is a screenshot that shows one level of subnet breakdown and some annotations and overlays.  In the future there could be more levels of compound nodes in the baseline network diagram.  Annotations will also include nodes with dynamic placement to animate different behavior along an arc from source node to dest node, so z-index of such overlay nodes to be above all other compound node children is needed.  Annotations and overlay arrows need to be on top of the baseline graph nodes and edges regardless of the depth.

screen shot 2016-06-06 at 9 58 27 am

Single node annotations are easy, because they are relative to a given node.  Edges can be handled given the discussion we had earlier.  Dynamically placed nodes for animated behavior may also require a similar style.

Let me know your take, so I can ensure PR agreement at least on the approach.  I'll then rework a PR against unstable.

I also need two edge labels drawn at the proper side of the connection (to label interface addresses), but I've not started that PR yet.  That change is larger because the notion of multiple labels breaks much of the code I looked at.  May require some kind of separator in the single label string and a flag to say split it up and draw at the ends.

maxkfranz:
I think `z-index-compare: auto | orphan | orphan-manual` would suffice then.

- `auto` : ordered as today (default)
- `orphan` : consider the element as though it had no parent
- `orphan-manual` : consider the element as though it had no parent and only compare `z-index` values (i.e. no edges before nodes precedence etc.)

maxkfranz:
I think `node, edge { z-index-compare: orphan-manual; }` would be equivalent to your `zorderStrict: true`, right?

mbeynon:
Yes, I think it would be equivalent.

mbeynon:
Since it's per-node and per-edge, need to consider conflicts of policy.  How would it work for an edge { z-index-compare: orphan-manual; } between two nodes with { z-index-compare: auto; } and say { z-index-compare: orphan; } ?

maxkfranz:
`auto` and `orphan` don't conflict, because `orphan` just assumes a compound depth of 0.  That means those two can just be compared as normal.

`orphan-manual` would also be depth 0, so its `z-index` would only be compared with other depth 0 elements.  I think you could have a conflict for:

- edge `a` : `z-index: 1; z-index-compare: auto;`
- edge `b` : `z-index: 2; z-index-compare: orphan-manual;`
- node `c` : `z-index: 1; z-index-compare: auto;`

Because the sort doesn't compare every pair of elements, the ordering is either `a => b => c` or `a => c => b` -- depending on whether `b` gets compared to `a` or `c`.

In that case, maybe `z-index-compare: auto | orphan | top-manual` would be better.  `top-manual` would assume depth -1 to avoid conflict.

mbeynon:
So `top-manual` would always be drawn on top?  The `-1` would have to be special cased, or it would draw before depth 0 and positive depth values for compound nodes.  Seems that `infinity` is how it would be treated.

maxkfranz:
The sorting is based on the (depth, ele, z) tuple, in left-to-right pirority like a number.  Each "digit" in the tuple is sorted ascendingly.

So yes, `top-manual` would need the highest depth in order to be the last drawn depth.  

Since then `orphan` is analogous to bottom, I could see someone else needing `top` (not manual).  Maybe we should use `Number.MAX_SAFE_INTEGER` for `top-manual` and `Number.MAX_SAFE_INTEGER - 1` for `top`.  I can't imagine someone having 9,007,199,254,740,991 depths...

maxkfranz:
Or maybe `orphan` should be `0` and `bottom` should be `-1`...

mbeynon:
Still some edge cases to consider.  To handle different `z-index-compare` property per-element, we should find a way to compute tuple values for a and b, and numerically compare them to sort.  Can short circuit them for efficiency.

Given (depth, ele, z-index) to describe sort order ...

1. For depth, we have a `z-index-compare` property of:
`bottom` is `-1`
`orphan` is `0`
`auto` is depth of compound node `0+`, or just `0` if not a compound node
`top` is `Number.MAX_SAFE_INTEGER - 1`
`top-manual` is `Number.MAX_SAFE_INTEGER`

2. For ele:
for `auto` use `0` for edges or `1` for nodes
otherwise use `1`

3. For z-index:
if specified, use the provided value, otherwise use `0`

Q: current logic broke ties between two nodes or two edges with indicies in the core.  Do we drop this, or include somehow?  If used, should it be a 4th item in the tuple?  Would not want it to compete with a user-specified z-index.

maxkfranz:
This looks right.

For (2) : Don't need a number, as edge < node suffices.

For (Q) : Yes, ties must be broken with the index, else the sorting might not be stable.

mbeynon:
edge < node suffices if `auto` for both, but need to consider other `z-index-compare` values.  What I listed was intended to solve combinations in a generic way to handle `auto` and other values in the same sort comparator call.

If we leave it as always edge < node, it will interfere with `top-manual` and explicit setting of a `z-index` value to purposely place an edge over a node.

maxkfranz:
The edge/node "digit"/comparison should just be ignored for the manual case

mbeynon:
That's equivalent to what I described above by assigning manual a `1` for the edge/node digit.  Assigning values for this digit forces a total ordering of elements which is required for a stable sort.

I will implement and submit a new PR as soon as I can.

maxkfranz:
Yes, you're right.  I think I should have been clearer by what I mean by the manual case.

If we have `bottom`, `orphan`, `auto`, `top`, and `top-manual` -- then I assume `top-manual` is the only manual case and the other cases have the default edges under nodes rule.

Perhaps for flexbility we should just separate the concerns:

- One property for depth : `z-compound-depth: bottom | orphan | auto | top`
- One property for edges under nodes: `z-index-compare: auto | manual`

@mbeynon What do you think?

mbeynon:
I was starting to think the name `z-index-compare` wasn't a good fit for all the depth concerns.  I like this much better and will be easier to read the code.

mbeynon:
This PR replaced with #1409

Issue #783, opened 2014-12-29
Edge (rendering) approach angle to/from a node


nmorse:
I would like to specify (or hint at) the angle of approach of an edge coming into (or out of) a node.
Possibly edge attributes `source_angle` and `target_angle` would either control or hint this to the renderer.

maxkfranz:
Good idea, though this could apply only to bezier curve edges.  Would that suit your usecase?

nmorse:
Yes, only bezier curve edges, that makes sense and is within my usecase. I also understand that this may cause some odd visual arrangements if nodes are not placed accordingly. But if it helps here is an example: 
![screenshot](https://cloud.githubusercontent.com/assets/25401/5601624/f360b012-92d9-11e4-9ad6-8427a837e65e.png)
I have a self edge that stems out of the top and I was hoping for a side or downward (visual) "flow".

maxkfranz:
That's a good point.  State machines are an important usecase.  Adding to 2.4 milestone

maxkfranz:
Removing from 2.4 milestone due to time constraints.  2.4 has been in the pipeline for a while now...

KulvinderSen:
Hi maxkfranz,
  
  I have attached an image below. I am getting layout shown in image part A) but what I need is layout B). Can you please point me toward the right direction that how can I achieve this?

![cystoscape](https://cloud.githubusercontent.com/assets/12406102/7580982/09c02816-f89b-11e4-9af1-53cc21438427.png)

maxkfranz:
@nmorse For your case, couldn't you reverse the arrow direction on the loops to get the effect you want?

tmullen71:
This is something I'd like to be able to do as well (also for state machines).

Lenndev:
So, this is still on the wishlist?

tmullen71:
I'd still definitely like it.

maxkfranz:
This is similar to the more general issue : https://github.com/cytoscape/cytoscape.js/issues/1219

That issue is being considered for 2.8, but its requirements need to be clarified.  In both that issue and this one, there are invalid/undesirable rendering results to consider:  It's easy to create bad results if you can specify position and/or angle.  

Further, this wouldn't work well with freeform graphs.  It would work with only heavily restricted or static graphs.

There are already a lot of high priority items (like performance-related enhancements) for 2.8, so  I'm not sure that issue will make it into 2.8.

tmullen71:
[...]

PR #1391, opened 2016-05-31
added loop direction options

tonymullen:
I've added options for having looping edges extend from the node in different directions. The direction is controlled by the 'loop-direction' css selector on edges. Possible options are 'north', 'south', 'east', 'west', 'northeast', 'southeast', 'southwest', and 'northwest' which is the default and is identical to previous looping behavior. 

Addresses issue #783


machow:
I swapped this branch in for the current version of cytoscape. It threw this error in the browser:

coord-ele-math.js:1331 Uncaught ReferenceError: eStyle is not defined

It's clear eStyle is never defined in that file (but I don't know how cytoscape does globals)

maxkfranz:
@tonymullen This looks great!  I wonder if we should also allow a manual angle value.  That may also help to consolidate the conditionals into a single formula

maxkfranz:
@machow Yes, I don't know about this `eStyle`.  All style values should be read via the internal `ele.pstyle()` (i.e. parsed style) function

tonymullen:
@maxfranz I'll give some thought to how a manual angle should work. 

I guess the control-point step could be made to come into play only when angles were identical. Should there be minimal increments, or would we want to allow arbitrarily closely positioned loops? 

Also, should the loop's internal angle be controllable? Currently north, south, east and west are narrower than the diagonal loops, but this could also be styled.

The direction angle style could allow any value and just calculate the angle by modulo 360. The loop angle would probably need to be constrained to something reasonable (<=90 or so). Maybe just treat any value larger than 90 as 90?

maxkfranz:
I was thinking just to specify the start angle (i.e. angle of the edge as it leaves the node from the beginning / source end).  The loop could just go 90 degrees from there.  I suppose there could be a second property to control the direction: clockwise or counterclockwise.

tonymullen:
Personally, I find 90 degrees to be too broad a sweep for my needs, so I like the idea of being able to set it, or have it default to something smaller, at least for NSEW options.

Setting an angle for the leaving edge seems more confusing to me than setting an angle for the overall direction of the loop (i.e. the angle that bisects the leaving and re-entering edge). So, e.g., 90 degrees would make the loop point straight up (or down, I forget). The loop's leaving and entering angles will always be symmetrical, so this seems like the simplest way to think of loop direction to me. 

I hadn't thought about counterclockwise vs clockwise, but that shouldn't be too difficult to implement. Something like 'loop-sweep-direction' with values 'cw' and 'ccw'.  

So I think I'd propose something like the following four properties: 

'loop-direction' (north, south, etc.)
'loop-direction-angle' (specific angle where 90 corresponds with north, etc. would override 'loop-direction')
'loop-sweep-direction' (cw, ccw)
'loop-sweep-angle' (default 90, min 1, max 135 (arbitrary but sane)).

maxkfranz:
Actually, we'd only need `loop-sweep` and not `-direction` and `-angle`.  The direction can be inferred by the sign.

maxkfranz:
@tonymullen So I think we'd have 

- `loop-direction` : `north`, `south`, or an angle (values like `north` would be converted to their corresponding angles)
- `loop-sweep` : the direction (sign) and amount (angle) the loop sweeps -- can be rad or deg (I think that's already built into the parser)

I think we should have just one default `loop-sweep` value to keep things simple.  Whatever the old loops sweep to should probably be the default so we're backwards compatible.

Thanks for organising this!

maxkfranz:
I really don't know whether `loop-direction` should denote the angle of the source or target parts of the loop.  Maybe it should denote the angle in the middle?  That would make converting values like `north` much more straightforward (e.g. `north` => 12 o'clock).

tonymullen:
Yes, I like the idea of having loop-direction denote the angle in the middle. And getting rid of the compass directions would work fine if that were the case. 

Also, good point about sweep direction being represented by the sign. 

I'll sit down and see about implementing these two properties soon.

tonymullen:
This is coming along fine so far. The only thing I'm unsure of is how to handle the stepping of overlapping edges. It seems to me that loops should step up (increase in size) only if they overlap directly (i.e. have the same direction and sweep angles). The most straightforward way to do this is to check all the loops on the node and step up from the largest one with identical angles. Let me know if you have other ideas on this.

tonymullen:
I think I've resolved the stepping question. This seems to be working as intended.

maxkfranz:
This looks good!  The only thing I think that needs to be revised code wise is the style property types.  They should both be `t.angle`.  That way you can specify `90deg` or `1.57rad`.

Numerical values all get normalised to their preferred units for convenienc, e.g. `ele.pstyle('loop-sweep').pfValue` would be in radians regardless of what units the user specified.

tonymullen:
Still working on getting the deg/rad issues completely sorted out.

maxkfranz:
You should just need to read `.pstyle( someProp ).pfValue` in [lines like this](https://github.com/cytoscape/cytoscape.js/pull/1391/commits/8900fd5ce60732933a79e8fe789bb58b5542e2ca#diff-436f2c18c50462e41ddc306440430eb1R1343) to get the value in radians.  Since you put a default value in the stylesheet, you don't need the `? : ` default value then.

tonymullen:
Sorry, just saw your comment. The code seems to be working now but I'm going to clean it up a bit more with reference to your comment.

maxkfranz:
Great; merged.  Thanks, Tony

PR #1453, opened 2016-07-12
Add 'ellipsize' options to text-wrap

floriandejonckheere:
Add the ability to draw single-line labels that are ellipsized instead of wrapped.

### DIFF
--- a/documentation/md/style.md
+++ b/documentation/md/style.md
@@ -306,8 +306,8 @@ Basic font styling:
 
 Wrapping text:
 
- * **`text-wrap`** : A wrapping style to apply to the label text; may be `none` for no wrapping (including manual newlines: `\n`) or `wrap` for manual and/or autowrapping.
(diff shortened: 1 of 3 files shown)


maxkfranz:
Great!

I made some minor revisions:

- Refactor
 - No `.split()` needed -- better performance
 - Check next char before append -- no need for tolerance and makes looping over the chars a bit more readable
- Rename value from "ellipsize" to "ellipsis"
 - [More like CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow)
 - Better for International English (i.e. -ize versus -ise)

aindlq:
I wonder if it is possible to cherry-pick this onto current master, or it uses something from unstable?

maxkfranz:
All new features go in the unstable branch for the next version.  The next version with new features will be 3.1

If you want to use anything from the unstable branch, you may as well just build snapshots from the unstable branch.

PR #1464, opened 2016-07-22
Add the new property of edges: distance-from-node

soomin-maeng:

### DIFF
--- a/src/extensions/renderer/canvas/drawing-edges.js
+++ b/src/extensions/renderer/canvas/drawing-edges.js
@@ -7,7 +7,7 @@ CRp.drawEdge = function( context, edge, shiftToOriginWithBb, drawLabel, drawOver
   var usePaths = this.usePaths();
 
   // if bezier ctrl pts can not be calculated, then die
-  if( rs.badLine || isNaN(rs.allpts[0]) ){ // isNaN in case edge is impossible and browser bugs (e.g. safari)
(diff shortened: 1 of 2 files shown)


maxkfranz:
@soomin-maeng This looks interesting!  A few notes:

(1) All PRs for new features should go on the unstable branch.  The master branch is only for bugfixes.

(2) You're doing calculations in the draw code.  That needs to be changed.  Note that the draw functions are relatively "dumb" -- they just read values and call the corresponding paint functions.  This is to keep the draw functions fast, because they can be called more often than rendered style calculation functions.  Further, you'll cause bugs with more complex edges with the change as you have it now.

Please refactor your calculations to go in the edge endpoint calculation function: https://github.com/cytoscape/cytoscape.js/blob/736edb00648f9f5d1460c532f61e6a71aa3b78dc/src/extensions/renderer/base/coord-ele-math.js#L1992'

I think you can just add the extra distance to the two offsets (for both start and end):

- https://github.com/cytoscape/cytoscape.js/blob/736edb00648f9f5d1460c532f61e6a71aa3b78dc/src/extensions/renderer/base/coord-ele-math.js#L2043
- https://github.com/cytoscape/cytoscape.js/blob/736edb00648f9f5d1460c532f61e6a71aa3b78dc/src/extensions/renderer/base/coord-ele-math.js#L2045

soomin-maeng:
Thanks for the suggestion!
I reuploaded it!

PR #1469, opened 2016-07-26
Added a new edge style property: distance-from-node

soomin-maeng:
capture1
capture2


soomin-maeng:
Please close this pull request.
I am sorry for the inconvenience...

PR #1470, opened 2016-07-26
Added a new edge style property: distance-from-node

soomin-maeng:
capture1
capture2


maxkfranz:
Merged with modification to allow separate specifications for the source side and the target side.

PR #1482, opened 2016-08-05
New line-style 'break': creates a gap in the middle of edges

soomin-maeng:
capture1
capture2

The gap in the middle of edges allow the user to read the labels more easily.
I tried this property with segment edge style, but it only works great when the zigzag section is symmetrical.

### DIFF
--- a/src/extensions/renderer/canvas/drawing-edges.js
+++ b/src/extensions/renderer/canvas/drawing-edges.js
@@ -107,6 +107,27 @@ CRp.drawEdgePath = function( edge, context, pts, type, width ){
       rs.pathCache = path;
     }
   }
+  
+  function bezierLength(x1, y1, x2, y2, x3, y3) {
+    //formula source : http://www.malczak.linuxpl.com/blog/quadratic-bezier-curve-length/
+    var a = {
+      x: x1 - 2 * x2 + x3,
+      y: y1 - 2 * y2 + y3
+    },
+    b = {
+      x: 2 * (x2 - x1),
+      y: 2 * (y2 - y1)
+    };
+    var A = 4 * (a.x * a.x + a.y * a.y);
+    var B = 4 * (a.x * b.x + a.y * b.y);
+    var C = b.x * b.x + b.y * b.y;
+    var A32 = Math.pow(A, 3/2);
+    var A12 = Math.sqrt(A);
+    var ABC12 = Math.sqrt(A+B+C);
+    var C12 = Math.sqrt(C);
+    return 1 / (8 * A32) * (4 * A32 * ABC12 + 2 * A12 * B * (ABC12 - C12) + 
+                           (4 * C * A - B * B) * Math.abs(Math.log((2 * A12 + B / A12 + 2 * ABC12) / (B / A12 + 2 * C12))));
+  }
 
   if( canvasCxt.setLineDash ){ // for very outofdate browsers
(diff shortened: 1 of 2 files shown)


maxkfranz:
Thanks for the pull request!

Have you tried using a background colour on labels, though?  That would give the same effect with a bit better coverage as edge orientation changes -- and it would be less computationally expensive as well.

soomin-maeng:
Hi,

I tried the background option, but the display for other edges became weird as the edges become thicker.

Thanks for the quick reply!

maxkfranz:
Do you have an example stylesheet and screenshot or a live example (like JSbin)?

It seems to work pretty well when I tried it:

screen shot 2016-08-09 at 2 04 25 pm

It also works nicely with autorotate:

screen shot 2016-08-09 at 2 05 21 pm

While I really appreciate your pull request, the performance of the background colour would be much faster.  The background also accommodates different edge orientations.  So, it would be great if the background could work for this usecase.

soomin-maeng:
Thank you for the prompt reply.
I did try the text-background-color property.
But, I encountered a number of problems.

The first problem I had with the background colour is that it starts to break other lines when two edges become close enough.

with break,
capture1
capture1a


with background color,
capture2a
capture2


The second problem is that when the edges become thick, the option with background color is not sufficient to do the same job as the line style 'break' does unless the border of the test gets calculated for that width of an edge. Also, even with the calculation, the first problem is not solved.

with break,
capture3
capture3a


with background colour,
capture4a
capture4

Lastly, when there are lots of thick edges, it becomes too obvious that the lines are not cut but rather covered with a box due their uneven cut near the labels.

maxkfranz:
If you're just looking for legibility and to minimise edge overlap issues, have you tried using a text outline?  It's the same style used on nodes in your example, but you could use white (or whatever colour you like) for edges.

e.g.

```
edge {
  text-outline-color: white;
  text-outline-width: 3px;
  text-outline-opacity: 1;
}
```

soomin-maeng:
Yes, but it creates same problems as it was with background colour.
Also, the cut in the edges are even stranger than than the cut with the background colour.

capture5
capture5a
capture6
capture6a

maxkfranz:
Have you tried something more similar to the node style?

```js
edge {
  color: #fff;
  line-color: #888;
  text-outline-color: #888; /* same colour as edge */
  text-outline-width: 3px;
  text-outline-opacity: 1;
}
```

Or have you tried 1px with the white outline?  Or try a thicker outline on thicker edges.  You may need to tweak the style a bit

maxkfranz:
If we want to merge this, there are some changes that would need to be made:

(1) Calculations (especially expensive ones like bezier length) should be done in the renderer's routine for recalculating the rendered style for elements (only when dirty).  Doing calculations on every draw is much more expensive, especially for large graphs.

(2) We should probably take into account more than text height in calculating the gap, because if the edge isn't straight up and down w.r.t. the text, then the gap isn't big enough.  Or maybe it would be better to keep it simple and just let the dev specify with a style property how big the gap should be (e.g. 20px for a 20px size font, maybe 30px if the dev wants extra spacing).

soomin-maeng:
Thank you for the suggestions.
I will come back to this when I get a better idea to solve the problem.

maxkfranz:
@soomin-maeng OK, make sure to let me know if you need any help

Issue #1499, opened 2016-08-21
null reference to renderer in v2.7.8


jiangliu:
With v2.7.8, it triggers following violation after reinitializing cytoscape. And a simple protection could fix the issue by checking "if (renderer)" before calling "renderer.notify( params );" in function notify at line 7245 in file cytoscape.js. Error messages as below.
Thanks!

Cannot read property 'notify' of null
隐藏详细
TypeError: Cannot read property 'notify' of null at corefn.notify (http://127.0.0.1:8080/x5/UI2/v_8c3422a0b2d148ae8a2fd8f22767054dl_zh_CNs_desktopd_pc/etrack/common/plugin/cytoscape/cytoscape.js:7246:13) at elesfn.rtrigger (http://127.0.0.1:8080/x5/UI2/v_8c3422a0b2d148ae8a2fd8f22767054dl_zh_CNs_desktopd_pc/etrack/common/plugin/cytoscape/cytoscape.js:3644:15) at elesfn.updateStyle (http://127.0.0.1:8080/x5/UI2/v_8c3422a0b2d148ae8a2fd8f22767054dl_zh_CNs_desktopd_pc/etrack/common/plugin/cytoscape/cytoscape.js:5090:16) at null.select (http://127.0.0.1:8080/x5/UI2/v_8c3422a0b2d148ae8a2fd8f22767054dl_zh_CNs_desktopd_pc/etrack/common/plugin/cytoscape/cytoscape.js:5402:19) at mouseupHandler (http://127.0.0.1:8080/x5/UI2/v_8c3422a0b2d148ae8a2fd8f22767054dl_zh_CNs_desktopd_pc/etrack/common/plugin/cytoscape/cytoscape.js:15138:20)

maxkfranz:
Are you trying to reuse a `.destroy()`ed instance?  Once an instance is destroyed, it can't be used anymore.

Would you please paste your re-init code here?

jiangliu:
I never call .destroy:) Just call showSpaceCy.prototype.showRoom() and alike repeatly.
	showSpaceCy.prototype.showRoom = function(row) {
		var self = this, eles = this.generateRoomEles(row);
		if (this.uiInfo.autoSize) {
			this.$container.height(this.yOffset * 2 + "px");
			this.xPanBy = 0;
			this.yPanBy = 0;
			this.uiInfo.padding = 10;
		} else {
			this.xPanBy = -30;
			this.yPanBy = 25;
		}
[...]

PR #1527, opened 2016-09-19
add 'destroy' event & provide info if already destroyed

EyMaddis:
Adds a `destroy` event which is triggered right **after** the instance was destroyed.
Allows external loops (like animated layouts) to listen to calls of `.destroy()`. Should provide the basis for future fixes for #1499 and related issues.

In addition to the event this PR also adds an `isDestroyed()` method (similar to `isReady()`) to check the state. This also allows for `.destroy()` to get called multiple times, as they can be safely ignored.


maxkfranz:
Looks great; thanks!

PR #1573, opened 2016-10-08
Animated dashed line Unstable

etienne69:
Hi,
I'd like to animate a dashed line like in Cytoscape software (Marquee equal line type for example). The aim is to represents a movement.
We told me it could me doable if there was a "line-style-offset" style property.
Thanks !


maxkfranz:
@etienne69 Thanks for the pull request!  Would you be able to apply your change cleanly on top of the current unstable branch?  It should just be a fast-forward operation so that Github can take in the change.  Thanks!

catobus:
Has this feature ever been implemented? I do cannot find any references to `line-style-offset` style property. 

Thanks!

maxkfranz:
No, it would be in the milestones if it were: https://github.com/cytoscape/cytoscape.js/milestones

catobus:
Do you have a suggestion how to achieve such [marching ant](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineDashOffset) effect on edges? 

Thanks!

maxkfranz:
Introduce the previously mentioned offset visual style property and then use the animation API on it.

PR #1587, opened 2016-10-25
update timing.js

mmaclach:
Invalid calling object fix for ie10/ie11


maxkfranz:
Thanks for the information!  I made a slightly different change with direct calls only -- which should be faster.

maxkfranz:
https://github.com/cytoscape/cytoscape.js/issues/1609

mmaclach:
sounds great, thanks

PR #1592, opened 2016-11-01
betweenness-centrality.js: fixed "max"(undefined) was never greater a…

XiaNi:
betweenness-centrality.js: fixed "max"(undefined by default) was never greater or smaller then any number

betweenness-centrality.js: betweennessNormalized() if max is zero - return zero
closeness-centrality.js: closeness() if maxCloseness is zero - return zero


maxkfranz:
@XiaNi Looks good.  Thanks!

Issue #884, opened 2015-03-26
`arrow-scale: multiplier` to control arrow size with a multiplier


rwu823:
In current  feature set the edge's CSS `width` able to adjust line and arrow's size. But how to adjust the arrow's size only?

ktei:
@rwu823 I don't think there is such setting at the moment, but you can simply implement it yourself and even do a pull request ; )

Find this file: src/extensions/renderer.canvas.arrow-shape.js and then find this line:
```javascript
draw: function(context, size, angle, translation) {
```
(around line 341, there are multiple draw functions, so make sure you find the arrow draw function)

Then you can change the `size`. I customized it to fit my project like this: `size = 10`, which might not be your requirement, but you got the idea. After that, remember to rebuild cytoscape using `gulp build`.

maxkfranz:
Currently, there's a bug making this not possible.  I added the feature to the temporary `arrowsize` branch, since the feature itself is simple.  However, the bug makes tapping edge arrows impossible for larger arrows.

If you can fix the branch, it could be merged to unstable for 2.4.  If not, it will have to wait for some later version.

maxkfranz:
The issue has been fixed for 2.5, so it should be possible to add this for 2.6.  Adding to milestone

jelmerjellema:
I implemented this for our project in 3 changes to the 2.7.9 source tree. It just adds a arrow-resize: number style property for general resizing of arrows. I use it as:
`edge {
  arrow-resize: 2.5
}`

So it just gives me bigger arrows all along. 

I see no issues with selecting the edge.

Should I push this somewhere?  It quite trivial anyway.

maxkfranz:
@jelmerjellema Yes, that would be great!  Please make a PR on the `unstable` branch, which is used for new versions.  Maybe it should be called `arrow-scale` rather than `size` or `resize` for clarity?  Thanks

jelmerjellema:
I'm sorry @maxkfranz. This slipped my attention. As a bitbucket mercurial user I don't know how to create a PR or what to put in it. Can I just send a patch or something and work from there. Love to help!

maxkfranz:
@jelmerjellema If you [fork the repo](https://github.com/cytoscape/cytoscape.js/issues/884#fork-destination-box), any changes that you push to your fork's git repo can be made into a pull request on Github.  You just click the pull request button on your repo's page.  [(More info)](https://help.github.com/articles/creating-a-pull-request-from-a-fork/)

[...]

PR #1606, opened 2016-11-18
`arrow-scale: multiplier` to control arrow size with a multiplier #884

metincansiper:
Please see #884


maxkfranz:
This looks like a good start, but other things will be broken with this change.

Some notes:

(1) If a default `arrow-scale` value is defined, you don't need `|| 1`.  And having `|| 1` effectively means the default value is duplicated.

(2) This change only affects rendering.  You won't be able to click arrows, for example.  Instead, the signature of `BaseRenderer.prototype.getArrowWidth()` should be changed to `.getArrowWidth( edgeWidth, scale )` and every call to `.getArrowWidth()` in the project needs to be updated to use the two parameters.

Issue #1607, opened 2016-11-18
`background-image-crossorigin`


maxkfranz:
`background-image-crossorigin: anonymous | use-credentials`

Control the `crossorigin` attribute on the image via a style property.  See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin

`anonymous` would be the default value. 

All browsers but IE11 behave properly with `anonymous`, but `use-credentials` is sometimes necessary for IE.

Ref : #1504

PR #1608, opened 2016-11-18
Add `background-image-crossorigin` style

jharwig:
Allows the crossorigin attribute on image requests to be changed using styles. Fixes https://github.com/cytoscape/cytoscape.js/issues/1607

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin
https://github.com/cytoscape/cytoscape.js/issues/1607

### DIFF
--- a/documentation/md/style.md
+++ b/documentation/md/style.md
@@ -184,6 +184,7 @@ Compound parent:
 A background image may be applied to a node's body:
 
  * **`background-image`** : The URL that points to the image that should be used as the node's background.  PNG, JPG, and SVG are supported formats.  You may use a [data URI](https://en.wikipedia.org/wiki/Data_URI_scheme) to use embedded images, thereby saving a HTTP request.
+ * ** `background-image-crossorigin`**: All images are loaded with a [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin) attribute which may be `anonymous` or `use-credentials`. The default is set to `anonymous`.
  * **`background-image-opacity`** : The opacity of the background image.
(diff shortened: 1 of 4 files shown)


jharwig:
https://github.com/cytoscape/cytoscape.js/issues/1607

maxkfranz:
Normally, new features go on the unstable branch.  Given this works around a browser's bug, let's just put this in 2.7.12

maxkfranz:
Thanks!

jharwig:
Given that I needed the fix, thanks for pushing it through 👍

xandris:
Hey, this doesn't appear to work correctly. The PR does this:

```javascript
    var bgImgCrossOrigin = node.pstyle( 'background-image-crossorigin' );

    // get image, and if not loaded then ask to redraw when later loaded
    image = this.getCachedImage( url, bgImgCrossOrigin, function() {
```

and then treats `bgImgCrossOrigin` like a string. The trouble is that `node.pstyle` returns an object instead. This works for me:

```javascript
    var bgImgCrossOrigin = (node.pstyle( 'background-image-crossorigin' )||{}).value;

    // get image, and if not loaded then ask to redraw when later loaded
    image = this.getCachedImage( url, bgImgCrossOrigin, function() {
```

unless I'm doing something wrong in my stylesheet JSON:

```json
      "css": {
        "background-color": "rgb(124, 130, 140)",
        "background-image": "/resources/src/img/question.svg",
        "background-image-crossorigin": "use-credentials"
      }
```

Did the behavior of `pstyle` change since this PR was merged? I can work around it but it's ugly.

maxkfranz:
Please see https://github.com/cytoscape/cytoscape.js/issues/1798

PR #1621, opened 2016-11-23
Remove falsy checks in place of empty string equality.

### LINKED ISSUE #1620: Allow non-string values to be specified as string property values (e.g. `label`)
Coerce non-string values to strings when specified to string style property values.  This will make it easier for devs who specify objects that can be `.toString()`ed.

OP follows:

---

When setting a label of anything falsy, does not render node labels. I came into a scenario where my node label was 0. Unfortunately, the label did not render due to the falsy check here.

https://github.com/cytoscape/cytoscape.js/blob/06e11b9c6792622d4d96be3da8b3001423465bff/src/extensions/renderer/canvas/drawing-label-text.js#L39

Simply removing the falsy check and checking for equality on an empty string resolved my issue. Making a PR now. Haven't dived into the source heavily so not sure what other side effects this may cause outside the scope of the function.

tbaeg:
Fixes #1620

### DIFF
--- a/src/extensions/renderer/canvas/drawing-label-text.js
+++ b/src/extensions/renderer/canvas/drawing-label-text.js
@@ -36,7 +36,7 @@ CRp.drawElementText = function( context, ele, force ){
   if( ele.isNode() ){
     var label = ele.pstyle( 'label' );
 
-    if( !label || !label.value ){ return; }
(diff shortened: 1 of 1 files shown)


maxkfranz:
Resolved by #1620

Issue #1614, opened 2016-11-21
`min-width`, `min-width-bias-left`, `min-width-bias-right`, etc.


maxkfranz:
`min-width` : Specifies the minimum size of the node's width.

`min-width-bias-left` : Specifies the portion of the additional width (in `%`) from `min-width` that is allocated on the left side of the node.

`min-width-bias-right` : Specifies the portion of the additional width (in `%`) from `min-width` that is allocated on the right side of the node.

If less than 100% is allocated between the left and right biases, then the weighted average of the total is used so that the total is effectively 100%.

The same properties apply for height, top, and bottom.

maxkfranz:
Defining a style property : https://github.com/cytoscape/cytoscape.js/blob/unstable/src/style/properties.js

Specifying compound size `ele.updateCompoundBounds()`:
https://github.com/cytoscape/cytoscape.js/blob/unstable/src/style/properties.js

PR #1630, opened 2016-11-30
`min-width`, `min-height` properties for updating compound bounds. #1614

d2fong:
- adds optional min-width, min-height properties for compound nodes
- computes autoWidth based on maximum of the boundingBox and min-width/height values
- adds commented out properties to the init script in the debug app

@maxkfranz @metincansiper @ugurdogrusoz 
Related issues: 
https://github.com/cytoscape/cytoscape.js/issues/1614
https://github.com/iVis-at-Bilkent/chise.js/issues/66

**Testing**

I have tested the following cases on the debug app:

width:
* left-bias > right-bias
* right-bias > left-bias
* left-bias = right-bias
* the above cases applied with left-bias+right-bias > 100
* the above cases applied with left-bias+right-bias < 100
* case when padding is not 0
* case when left-bias + right-bias = 0

height:
* the same cases applied mentioned above applied to height and its associated properties

events:
* clicking outside of the compound bound does not select the compound
* clicking inside or on the compound boundary selects the compound
* selecting edges and child nodes in the compound selects the child/edge
* creating nodes inside the compound adjusts the compound correctly

**Demo**

You can view a demo of the changes here:

https://www.youtube.com/watch?v=1wwLdkjLlsU

### DIFF
--- a/debug/init.js
+++ b/debug/init.js
@@ -148,6 +148,19 @@ $(function(){
         .css({
           'curve-style': 'bezier'
         })
+      .selector('node#n1')
+        .css({
+          // 'min-width': 400,
+          // 'min-width-bias-left': 30,
+          // 'min-width-bias-right': 60,
+          // 'padding-left': 0,
+          // 'padding-right': 0,
+          // 'padding-bottom': 0,
+          // 'padding-top': 0,
+          // 'min-height': 400,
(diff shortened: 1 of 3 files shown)


maxkfranz:
Ref : #1614

Issue #1634, opened 2016-12-03
Breadthfirst layout fails for secondary call on subset of elements


markgoffin:
This bug will show itself if a breadthfirst `layout()` is called on a subset of elements which have already been breadthfirst 'laidout'. The symptom is that an out of bounds access is attempted on the `depths` array in the function `getWeightedPercent` (breadthfirst.js:303). This happens because the neighbors array initialised on line 294 takes **all** neighbours of that node, whereas it should make sure it only takes neighbours that were in the collection being laid out. 

The out of bounds access occurs because the previous steps in the layout function populate the `breadthfirst` object inside `_private.scratch` for each node in the specified collection. The call to get the neighbours picks up all neighbours which may have been in a breadthfirst layout before, it will then use the data from the previous layout in this function which will not be valid.

I believe the fix here is to ensure that the neighbours used in this function are a part of the collection being laid out.

PR #1635, opened 2016-12-03
Fix bug in breadthfirst layout for a collection of elements

markgoffin:
This is a bugfix for #1634. The majority of the details are mentioned in the issue.


markgoffin:
Also, sorry I haven't added any tests. I'm not a javascript dev by trade and am not familiar with mocha.

maxkfranz:
Good catch; thanks!

PR #1672, opened 2017-01-06
arrow-scale property with getArrowWidth()

josephst:
Addresses issue #884 and PR #1606.The arrow-scale property for edge styles will multiply the edge by the given factor to make edge tips (arrow heads, usually) larger or smaller. When a user does not specify the scaling factor, the default value of 1 is used. The signature for `getArrowWidth` has been changed to accommodate a scaling factor >= 0. All calls to `getArrowWidth()` now include the scaling factor.

Documentation is provided for the new style property.

This is a commit that cleans up several smaller commits. 

[JSFiddle to illustrate the new scaling factors](https://jsfiddle.net/josephst18/rkd4bLs4/1/)

- Large: 
Large scaling
- Small: 
Small scaling


src/extensions/renderer/base/coord-ele-math.js
```diff
@@ -2182,15 +2184,15 @@ BRp.findEndpoints = function( edge ){
   }
 };
 
-BRp.getArrowWidth = BRp.getArrowHeight = function( edgeWidth ){
+BRp.getArrowWidth = BRp.getArrowHeight = function( edgeWidth, scale ){
   var cache = this.arrowWidthCache = this.arrowWidthCache || {};
 
   var cachedVal = cache[ edgeWidth ];
```
maxkfranz: I think the cache key needs to take into account the scale.

josephst:
Caching now uses `edgeWidth + ', ' + scale` for the key (I played around with using various punctuation as a delimiter, or some kind of hash function, before settling with the easy thing and using the comma).

Gaps between edge lines and arrows should also be fixed; the `gap` property now takes into account `arrow-scale` for relevant shapes (all shapes except tee).

maxkfranz:
I noticed a couple things after merging.

(1) The property can allow negative numbers.  I've pushed a commit with a change that adds a positive number type.  You can see how you specify new types as a series of flags basically.  New flags (like `strictMin`) can be added in the style property parser.  If you need to add new properties in future, you can use this an example.

(2) The gap for `triangle-backcurve` seems off for `arrow-scale: 0.75` on the debug page.

(3) The gap is the same for lots of arrow types.  Could they be refactored out to a separate function, like `halfGap()` because it's half the default, and just referenced by all those shape definitions? 

Thanks!

josephst:
1. Thanks for fixing that.
2. I agree that there's something wrong. With the small gap (not multiplying by 2) the edge will overshoot the arrow on small scaling values. If I multiply by two (bringing it in line with the `halfGap()` calculation method) it will never overshoot but will leave an empty space at the bottom of the arrow (between concave base and flat end of edge). I'll play around with it and see if there's a way I can balance the two.
3. Absolutely.

maxkfranz:
OK, re. (2) it probably needs some float value between the two.  Probably easiest way is to just experiment with different values, doing a binary search between 1 and 2

Issue #1266, opened 2016-02-24
Spacing factor override option in layouts


maxkfranz:
Allows a postprocessing factor to be applied to spacing, because auto avoid overlap can space things out far if one element is large

This would update the definition of `eles.layoutPositions()` so that the option would work for all discrete layouts that call it.

`options.spacingFactor` : Applies a multiplicative factor to expand or compress the overall area that the nodes take up.  For example, a value of 2 would make the nodes twice as spaced apart.

josephst:
I've added `spacingFactor` to the grid layout and am working on adding to others. Note: breadthfirst layout already includes a `spacingFactor` option which appears to work identically to the suggested `spacingFactor` option.

maxkfranz:
@josephst Probably the `spacingFactor` in the bfs layout could be removed then.  We should probably use the name `spacingFactor` overall for backwards compatibility -- or we should check if  `spacingFactor` exists and use it for `scalingFactor`.

I think the positions have to be normalised before scaling is applied.  The scaling needs to happen as a vector relative to the bounding box centre, because the bounding box might not be (0, 0) to (w, h).

josephst:
Newest additions to #1673 should check the bounding box for the elements that the layout is being run on. I've taken out the `spacingFactor` option from bfs now that spacing is handled in core layout. I verified that the bfs demo produces an identical result. 

The spacing factor doesn't work with the cose and cose-bilkent layouts; perhaps because it's iterative and the node positions that are being scaled aren't the final node positions?

PR #1673, opened 2017-01-07
Add scalingFactor for adjusting layouts post-calculation #1266

josephst:
I've tentatively named the option scalingFactor rather than spacingFactor as the `breadthfirst` layout option already includes a spacingFactor option. This PR adds a `scalingFactor` option that multiplies the layout-calculated position of the node by the (absolute value of the) `scalingFactor` to produce a final position to be rendered. 

Ex: a layout which spaces nodes out by 50px will have 100px gaps between nodes when `scalingFactor` is 2.

### DIFF
--- a/src/collection/layout.js
+++ b/src/collection/layout.js
@@ -22,6 +22,10 @@ var elesfn = ({
         var newPos = fn.call( node, i, node );
         var pos = node.position();
 
+        if( !is.number( pos.x ) || !is.number( pos.y ) ){
+          node.silentPosition( { x: 0, y: 0 } );
+        }
+
         if ( options.spacingFactor && options.spacingFactor !== 1){
           var spacing = Math.abs(options.spacingFactor);
           var nbb = nodes.boundingBox();
@@ -34,10 +38,6 @@ var elesfn = ({
             y: (newPos.y - center.y) * spacing
           };
           newPos = util.extend( {}, newPos, { x: center.x + spacingVector.x, y: center.y + spacingVector.y } );
-        }
-
-        if( !is.number( pos.x ) || !is.number( pos.y ) ){
-          node.silentPosition( { x: 0, y: 0 } );
         }
 
         var ani = node.animation( {
--- a/src/collection/layout.js
+++ b/src/collection/layout.js
@@ -22,9 +22,18 @@ var elesfn = ({
         var newPos = fn.call( node, i, node );
         var pos = node.position();
 
-        if ( options.scalingFactor && options.scalingFactor !== 1){
-          var scale = Math.abs(options.scalingFactor);
-          newPos = util.extend( {}, newPos, { x: newPos.x * scale, y: newPos.y * scale } );
+        if ( options.spacingFactor && options.spacingFactor !== 1){
+          var spacing = Math.abs(options.spacingFactor);
+          var nbb = nodes.boundingBox();
+          var center = {
+            x: nbb.x1 + nbb.w / 2,
+            y: nbb.y1 + nbb.h / 2
+          };
+          var spacingVector = { // scale the spacing from center of bounding box (not necessarily 0,0)
+            x: (newPos.x - center.x) * spacing,
+            y: (newPos.y - center.y) * spacing
+          };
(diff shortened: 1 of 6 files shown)


src/collection/layout.js
```diff
@@ -26,6 +26,20 @@ var elesfn = ({
...
+          var nbb = nodes.boundingBox();
+          var center = {
+            x: nbb.x1 + nbb.w / 2,
+            y: nbb.y1 + nbb.h / 2
+          };
+          var spacingVector = { // scale the spacing from center of bounding box (not necessarily 0,0)
+            x: (newPos.x - center.x) * spacing,
+            y: (newPos.y - center.y) * spacing
+          };
+          newPos = util.extend( {}, newPos, { x: center.x + spacingVector.x, y: center.y + spacingVector.y } );
```
maxkfranz: Don't need extend, just need assign to new pos obj

src/collection/layout.js
```diff
@@ -74,6 +88,26 @@ var elesfn = ({
     } else {
       nodes.positions( fn );
 
+      if ( options.spacingFactor && options.spacingFactor !== 1){
```
maxkfranz: Could most of the logic in this block be refactored to a separate function to do the vector stuff.  That way, it can be reused for the top case and the bottom case, making things terser.

src/collection/layout.js
josephst: Thanks. Wasn't sure if the newPos object had other properties I would wipe by reassigning it.

josephst:
Refactored to a new method called `calculateSpacing`.

src/collection/layout.js
```diff
@@ -26,6 +42,11 @@ var elesfn = ({
           node.silentPosition( { x: 0, y: 0 } );
         }
 
+        if ( options.spacingFactor && options.spacingFactor !== 1){
+          var spacing = Math.abs(options.spacingFactor);
```
josephst: Right now `var spacing = Math.abs(options.spacingFactor)` is outside the function so that it doesn't get called *n* times in the no-animation case. It could also be moved inside the function which would make everything more concise (and would be one less argument that needs to be passed).

src/collection/layout.js
josephst: Done. Right now `var spacing = Math.abs(options.spacingFactor)` is outside the function so that it doesn't get called *n* times in the no-animation case. It could also be moved inside the function which would make everything more concise (and would be one less argument that needs to be passed).

maxkfranz:
Great.  The bb can be expensive to calculate, though.  So, it would be better to pass in the bb to the function rather than the nodes.  That allows the second case to receive the bb passed in on each call rather than recalculating on each call.

src/collection/layout.js
```diff
@@ -15,6 +15,22 @@ var elesfn = ({
 
     layout.animations = [];
 
+    var calculateSpacing = function( spacing, nodes, pos ){
```
maxkfranz: `calculateSpacing( spacing, nodesBb, pos )` would be a more efficient signature to avoid unnecessary bb calcs

src/collection/layout.js
```diff
@@ -74,6 +95,15 @@ var elesfn = ({
     } else {
       nodes.positions( fn );
```
maxkfranz: I just noticed we're calling .positions() twice.  I think this one has to go.

src/collection/layout.js
```diff
@@ -74,6 +95,15 @@ var elesfn = ({
...
 
+      if ( options.spacingFactor && options.spacingFactor !== 1){
+        var spacing = Math.abs(options.spacingFactor);
+        nodes.positions( function( i, node){
+          var pos = node.position();
```
maxkfranz: I think if the previous `.positions()` call is removed we can just have `pos` set by `fn()` with `fn()` called as described in http://js.cytoscape.org/#nodes.layoutPositions

josephst:
Should be fixed now. In the non-animated case, `nodes.position(fn)` is called if spacingFactor doesn't matter (same behavior as before). If spacingFactor specified, it'll get the new position via `fn(i, node)`, then use this position in `calculateSpacing()`. Only calls `nodes.positions()` once.

maxkfranz:
Merged; thanks!

Issue #1656, opened 2016-12-20
Implementing triangle-tee arrow shape variants


d2fong:
@maxkfranz @ugurdogrusoz  @gbader @jvwong @metincansiper @IgorRodchenkov 

### Note
**This is a translation of my notes from talking with Max about how to implement custom arrow 
shapes.  They represent a skeleton implementation and they are not fully fleshed out.**  

# Requirements
Custom arrow shapes are needed for cytoscape.js.

Here is an example arrow shape that is required:
![screen shot 2016-12-19 at 11 34 39 am](https://cloud.githubusercontent.com/assets/2328291/21365977/c74dff62-c6c5-11e6-8109-6ceaeb25d810.png)

# Implementation Notes

### Draw the arrow-tee in phases
Split up the drawing execution into phases. 

For example:
* phase 0 draws the arrow head
* phase 1 draws the arrow edge connector
* phase 2 draws extra lines that represent the variations in the arrow-tee
Each portion of the shape can be rendered differently via config values associated with each phase.

### Phase config values
* ```multiphased```: boolean representing whether this shape should be drawn in multiple phases or not
* ```matchEdgeWidth```: list of n booleans corresponding to the n drawing phases for the arrow shape.  If matchEdgeWidth[n-1] is true, then phase n-1 will be drawn with the same line width as the edge it is attached to.
* ```forceStroke```: boolean to override a shape style and render it using strokes instead of fill. (Used for special cases such as a line between two points.  In this case, you need to draw it using a stroke)

### Other remarks

 The collision function for `necessary stimulation` could be improved. 

```js
    collide: function( x, y, size, angle, translation, padding ){
      var triPts = pointsToArr( transformPoints( this.points, size + 2 * padding, angle, translation ) );
      var teePts = pointsToArr( transformPoints( this.pointsTee, size + 2 * padding, angle, translation ) );

      var inside = math.pointInsidePolygonPoints( x, y, triPts ) || math.pointInsidePolygonPoints( x, y, teePts );

      return inside;
    },
```

```math.inlineVicinity``` and ```math.squareDistanceToFiniteLine``` could be better suited for detecting collisions.

```pointsTee``` should be renamed to ```pointsLine``` or something more descriptive -- the points currently do not represent a tee.

A generic name for this type of arrow should be considered over the biology specific ```necessary stimulation``` term.  ```arrow-tee-stripe```, ```arrow-tee-cross``` are potential names.

metincansiper:
[...]

PR #1683, opened 2017-01-18
Triangle-cross arrowhead

d2fong:
## High Level Overview

- implements a new arrow head shape called 'triangle-cross'
- collision detection using inline vicinity
- scales arrow size based on edge width and size

## Miscellaneous Changes

- removes explicit arrowShape check for 'half-triangle-overshot' in the drawArrowHead function
- reuses arrowWidth and arrowShape variables that were already computed at the beginning of the drawArrowHead function
- shapeImpl.draw takes edgeWidth as a parameter for arrow scaling

See https://github.com/cytoscape/cytoscape.js/issues/1656

### DIFF
--- a/src/extensions/renderer/base/arrow-shapes.js
+++ b/src/extensions/renderer/base/arrow-shapes.js
@@ -191,6 +191,60 @@ BRp.registerArrowShapes = function(){
     }
   } );
 
+  defineArrowShape( 'triangle-cross', {
+    points: [
+      -0.15, -0.3,
+      0, 0,
+      0.15, -0.3,
+      -0.15, -0.3
+    ],
+
+    crossLinePoints: [
+      -0.24175, -0.4,
+      0.24175, -0.4,
+    ],
+
+    forceStroke: true,
+
+    matchEdgeWidth: true,
+
+    scaleCoord: function ( constant, size, edgeWidth ){
+     return constant + ( edgeWidth * 0.012 ) + ( math.log2( size - 28.95 ) * 0.001 );
+    },
+
+   scaleCrossLineXCoord: function( size, edgeWidth ){
+      return this.scaleCoord( 0.42, size, edgeWidth );
+    },
+
+    scaleCrossLineYCoord: function( size, edgeWidth ){
+      return this.scaleCoord( -0.01, size, edgeWidth );
+    },
+
+    collide: function( x, y, size, angle, translation, padding ){
+      var triPts = pointsToArr( transformPoints( this.points, size + 2 * padding, angle, translation ) );
+      var crossLinePts = pointsToArr( transformPoints( this.crossLinePoints, size + 2 * padding, angle, translation ) );
+
+      var inside = math.pointInsidePolygonPoints( x, y, triPts )
+      || math.inLineVicinity( x, y,
+        crossLinePts[0], crossLinePts[1], crossLinePts[2], crossLinePts[3], padding );
+
+      return inside;
+    },
+
+    draw: function( context, size, angle, translation, edgeWidth ){
+      var scaledCrossLine = [
+        this.crossLinePoints[0] + this.scaleCrossLineXCoord( size, edgeWidth ),
+        this.crossLinePoints[1] - this.scaleCrossLineYCoord( size, edgeWidth ),
+        this.crossLinePoints[2] - this.scaleCrossLineXCoord( size, edgeWidth ),
+        this.crossLinePoints[3] - this.scaleCrossLineYCoord( size, edgeWidth )
+      ];
+      var triPts = transformPoints( this.points, size, angle, translation );
+      var crossLinePts = transformPoints( scaledCrossLine, size, angle, translation );
+
+      renderer.arrowShapeImpl( this.name )( context, triPts, crossLinePts );
+    }
+  } );
+
   defineArrowShape( 'vee', {
     points: [
       -0.15, -0.3,
--- a/src/extensions/renderer/canvas/drawing-edges.js
+++ b/src/extensions/renderer/canvas/drawing-edges.js
@@ -183,20 +183,20 @@ CRp.drawArrowhead = function( context, edge, prefix, x, y, angle ){
 
   var self = this;
   var arrowShape = edge.pstyle( prefix + '-arrow-shape' ).value;
-
-  if( arrowShape === 'none' ){
-    return;
-  }
-
-  var gco = context.globalCompositeOperation;
+  if( arrowShape === 'none' ) { return; }
 
   var arrowClearFill = edge.pstyle( prefix + '-arrow-fill' ).value === 'hollow' ? 'both' : 'filled';
   var arrowFill = edge.pstyle( prefix + '-arrow-fill' ).value;
+  var edgeWidth = edge.pstyle( 'width' ).pfValue;
   var opacity = edge.pstyle( 'opacity' ).value;
 
-  if( arrowShape === 'half-triangle-overshot' ){
-    arrowFill = 'hollow';
-    arrowClearFill = 'hollow';
+  var gco = context.globalCompositeOperation;
+
+  var shapeImpl = self.arrowShapes[ arrowShape ];
+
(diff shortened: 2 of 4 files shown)


src/extensions/renderer/base/arrow-shapes.js
```diff
@@ -191,6 +191,56 @@ BRp.registerArrowShapes = function(){
...
+    forceStroke: true,
+
+    matchEdgeWidth: true,
+
+    scaleCrossLineYCoord: function( size, edgeWidth ){
+      return -0.01 + ( edgeWidth * 0.012 ) + ( math.log2( size - 28.95 ) * 0.001 );
+    },
+    scaleCrossLineXCoord: function( size, edgeWidth ){
+      return 0.42 + ( edgeWidth * 0.012 ) +  ( math.log2( size - 28.95 ) * 0.001 );
+    },
```
d2fong: These two functions could be changed to just one, with the costant as a parameter, but I thought it would be more clear that both the x and y coordinates of the cross can be scaled independently.

src/extensions/renderer/base/arrow-shapes.js
```diff
@@ -191,6 +191,56 @@ BRp.registerArrowShapes = function(){
...
+      var crossLinePts = pointsToArr( transformPoints( this.crossLinePoints, size + 2 * padding, angle, translation ) );
+
+      var inside = math.pointInsidePolygonPoints( x, y, triPts )
+      || math.inLineVicinity( x, y,
+        crossLinePts[0], crossLinePts[1], crossLinePts[2], crossLinePts[3], padding );
+
+      return inside;
+    },
+
+    draw: function( context, size, angle, translation, edgeWidth ){
```
d2fong: I found that I could make the scaling functions more accurate if I used the edgeWidth parameter to calculate the scaling.

src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -181,45 +181,42 @@ CRp.drawArrowheads = function( context, edge, drawOverlayInstead ){
 CRp.drawArrowhead = function( context, edge, prefix, x, y, angle ){
   if( isNaN( x ) || x == null || isNaN( y ) || y == null || isNaN( angle ) || angle == null ){ return; }
 
-  var self = this;
```
d2fong: I don't think self is needed in this case?

src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -181,45 +181,42 @@ CRp.drawArrowheads = function( context, edge, drawOverlayInstead ){
...
-
-  var gco = context.globalCompositeOperation;
+  if( arrowShape === 'none' ) { return; }
 
   var arrowClearFill = edge.pstyle( prefix + '-arrow-fill' ).value === 'hollow' ? 'both' : 'filled';
   var arrowFill = edge.pstyle( prefix + '-arrow-fill' ).value;
+  var arrowWidth = edge.pstyle( 'width' ).pfValue;
   var opacity = edge.pstyle( 'opacity' ).value;
 
-  if( arrowShape === 'half-triangle-overshot' ){
```
d2fong: Removes old code like you said.

src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -181,45 +181,42 @@ CRp.drawArrowheads = function( context, edge, drawOverlayInstead ){
...
-    self.fillStyle( context, 255, 255, 255, 1 );
-    self.strokeStyle( context, 255, 255, 255, 1 );
+    this.fillStyle( context, 255, 255, 255, 1 );
+    this.strokeStyle( context, 255, 255, 255, 1 );
 
-    self.drawArrowShape( edge, prefix, context,
-      arrowClearFill, edge.pstyle( 'width' ).pfValue, edge.pstyle( prefix + '-arrow-shape' ).value,
-      x, y, angle
+    this.drawArrowShape( edge, prefix, context,
+      arrowClearFill, arrowWidth, arrowShape, x, y, angle
```
d2fong: Uses arrowWidth, arrowShape variables instead of getting the value again.

src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -253,7 +250,7 @@ CRp.drawArrowShape = function( edge, arrowType, context, fill, edgeWidth, shape,
   if( context.beginPath ){ context.beginPath(); }
 
   if( !pathCacheHit ){
-    shapeImpl.draw( context, size, angle, translation );
+    shapeImpl.draw( context, size, angle, translation, edgeWidth );
```
d2fong: This is where I send the edgeWidth to the 'triangle-cross' draw function.  Again, thoughts on this?

src/extensions/renderer/base/arrow-shapes.js
maxkfranz: Maybe `scaleCrossLineYCoord` and X can just reference the same impl of a function defined in a block above?

src/extensions/renderer/canvas/drawing-edges.js
maxkfranz: OK.  It's probably just used in most functions in case a ref to `this` is needed in an inner function.  It may not be needed here, but a side benefit is smaller minified files.  `this` is always 4 chars minified because it's a keyword; `self` becomes something 1 char like `x`.

src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -181,45 +181,42 @@ CRp.drawArrowheads = function( context, edge, drawOverlayInstead ){
...
-  if( arrowShape === 'none' ){
-    return;
-  }
-
-  var gco = context.globalCompositeOperation;
+  if( arrowShape === 'none' ) { return; }
 
   var arrowClearFill = edge.pstyle( prefix + '-arrow-fill' ).value === 'hollow' ? 'both' : 'filled';
   var arrowFill = edge.pstyle( prefix + '-arrow-fill' ).value;
+  var arrowWidth = edge.pstyle( 'width' ).pfValue;
```
maxkfranz: Should not be called `arrowWidth`.  That's actually a different value, semantically.  Better to call it `edgeWidth` or it will be confusing.

src/extensions/renderer/base/arrow-shapes.js
```diff
@@ -191,6 +191,60 @@ BRp.registerArrowShapes = function(){
...
+      0.24175, -0.4,
+    ],
+
+    forceStroke: true,
+
+    matchEdgeWidth: true,
+
+    scaleCoord: function ( constant, size, edgeWidth ){
+     return constant + ( edgeWidth * 0.012 ) + ( math.log2( size - 28.95 ) * 0.001 );
+    },
```
d2fong: base function to scale a coordinate

src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -183,20 +183,20 @@ CRp.drawArrowhead = function( context, edge, prefix, x, y, angle ){
...
-  if( arrowShape === 'none' ){
-    return;
-  }
-
-  var gco = context.globalCompositeOperation;
+  if( arrowShape === 'none' ) { return; }
 
   var arrowClearFill = edge.pstyle( prefix + '-arrow-fill' ).value === 'hollow' ? 'both' : 'filled';
   var arrowFill = edge.pstyle( prefix + '-arrow-fill' ).value;
+  var edgeWidth = edge.pstyle( 'width' ).pfValue;
```
d2fong: renamed variable to edgewidth

d2fong:
@maxkfranz  I have made the changes that you have suggested. 

List of Changes:
- Reverted the self -> this change
- Renamed the incorrect arrowWidth name to edgeWidth 
- Renamed variables with prefix 'tee' to the prefix 'crossLine' in the triangle-cross implementation

This new arrow head scales both with edge width and arrow scale.  When both values are large, it makes it look a bit wonky, but I believe it is working.  EdgeWidth scales the edge width, arrow scale scales the arrow head.

Issue #1392, opened 2016-05-31
6,000 elements demo


maxkfranz:
Take code from https://github.com/cytoscape/cyjs2.6-cyjs2.7-desktop-comparison

mj3cheun:
Merged in pull request https://github.com/cytoscape/cytoscape.js/pull/1704. Closing.

PR #1704, opened 2017-02-03
Added 6000 element demo #1392

mj3cheun:


mj3cheun:
Ok changes made

maxkfranz:
A couple of things:

(1) It would be good to remove the reference to 2.7 on the demo, as it will carry forward to future versions

(2) There needs to be a `fetch()` polyfill for older browsers.  You can just include a ref to a copy from a CDN: https://cdnjs.com/libraries/fetch

mj3cheun:
I have been doing to more testing and I found that fetch doesn't work on Chrome and Chrome derivative browsers when opening the index.html file manually as url scheme "file" is not supported. It works fine when opening the file through a static file server as "http" is being used. It also works fine in firefox. Do you see this as a problem?

maxkfranz:
> I have been doing to more testing and I found that fetch doesn't work on Chrome and Chrome derivative browsers when opening the index.html file manually as url scheme "file" is not supported. It works fine when opening the file through a static file server as "http" is being used. It also works fine in firefox. Do you see this as a problem?

No, it's always running behind an HTTP server.  You should use `http-server` from npm if you want to use the Cyjs debug page or view the docs page etc.

mj3cheun:
Ok made changes and also included promise polyfill just in case.

PR #1706, opened 2017-02-06
Added cola-compound demo #1043

mj3cheun:


documentation/demos/cola-compound/cytoscape-cola.js
```diff
@@ -0,0 +1,549 @@
+/*!
```
maxkfranz: This file should not be included, nor the Cola lib itself

documentation/demos/cola-compound/index.html
```diff
@@ -0,0 +1,1223 @@
...
+						css: {
+							'background-color': '#aaa',
+							'line-color': 'black',
+							'target-arrow-color': 'black',
+							'source-arrow-color': 'black'
+						}
+					}
+				],
+
+				elements: {
```
maxkfranz: Please put the graph in a separate json file to keep the demo easier to read.

documentation/demos/cola-compound/index.html
```diff
@@ -0,0 +1,1223 @@
...
+				container: document.getElementById('cy'),
+
+				layout: {
+					name: 'cola'
+				},
+
+				style: [{
+						selector: 'node',
+						css: {
+							'content': 'data(label)',
```
maxkfranz: Please use 2 spaces rather than tabs for indentation.

maxkfranz:
Great.  Could we use a simpler dataset like the one here? http://marvl.infotech.monash.edu/webcola/examples/gridifiedSmallGroups.html

mj3cheun:
I couldn't find any compound graph smaller in in Dylan's graph list or in Pathway Commons so I just started deleting nodes from the existing graph. The graph probably doesn't make sense biologically anymore but its smaller. Also added ES6 polyfills just in case.

maxkfranz:
Could we use the same data from http://marvl.infotech.monash.edu/webcola/examples/gridifiedSmallGroups.html ? It would be good for it to be comparable to the Cola example

mj3cheun:
As far as I can tell, they use a different data format. I guess I could figure out how to reconstruct that dataset though.

maxkfranz:
OK, great.  Worst case, it's not many nodes

Issue #1158, opened 2015-11-14
Padding as percent values


maxkfranz:
Specify padding values relative to the width/height of the node.  The horizontal paddings are relative to the (inner) width and the vertical paddings are relative to the (inner) height.

PR #1712, opened 2017-02-14
Added percentage padding functionality #1158

mj3cheun:


documentation/md/style.md
```diff
@@ -168,7 +168,7 @@ Padding:
 
 A padding defines an addition to a node's dimension.  For example, `padding` adds to a node's outer (i.e. total) width and height.  This can be used to add spacing around the label of `width: label; height: label;` nodes, or it can be used to add spacing between a compound node parent and its children.
 
-* **`padding`** : The amount of padding around all sides of the node.
+* **`padding`** : The amount of padding around all sides of the node. Either percentage or pixel value can be specified.
```
maxkfranz: It would be good to have example values here

src/collection/dimensions.js
```diff
@@ -266,6 +266,20 @@ fn = elesfn = ({
         };
       }
 
+      function computePaddingValues( width, paddingObject ) {
+        // Assuming percentage is number from 0 to 1
+        // Follows CSS2 specifications for padding
+        if(paddingObject.units === '%') {
+          return (width > 0 ? paddingObject.pfValue * width : 0);
+        }
```
maxkfranz: Preferable to use same line `} else {` style

src/collection/dimensions.js
```diff
@@ -266,6 +266,20 @@ fn = elesfn = ({
         };
       }
 
+      function computePaddingValues( width, paddingObject ) {
```
maxkfranz: Because this only depends only on the width, we may want another style property (enum) to control the values used.  For example, we may want to use

- height
- average of height and width
- min of height and width
- max of height and width

@d2fong What cases would be needed for SBGN usecases?  Do you think it would be best to just support all above cases with another style property?

I.e. `padding-relative-to: width | height | min | max | average`

src/collection/dimensions.js
```diff
@@ -925,6 +941,10 @@ defineDimFns( {
   name: 'height'
 } );
 
+defineDimFns( {
```
maxkfranz: The padding function should be defined manually.  `defineDimFns()` is only for brevity in defining width and height.  It may seem to work, but there are cases where using this for padding with break things.

src/collection/dimensions.js
d2fong: Yes I think it would be nice to have these options supported.  I can think of cases where I can use all of these options.

mj3cheun: Ok. I would prefer the idea of having `padding-relative-to: width | height | min | max | average` though. I tried to follow the spec here (https://www.w3.org/TR/CSS2/box.html#padding-properties) so that padding behaves like people are used to, which means everything is calculated relative to width. I think if it behaves any other way, it wouldn't really be padding. We can maybe make padding an alias for `padding-relative-to: width` for backwards compatibility.

maxkfranz: `width` could be the default if it most dev usecases would use it.  `padding-relative-to` would be a different property than `padding` and would just control what `padding` as a percent would mean.

It's not a big deal if we're different than CSS.  We already are in several places, because CSS proper doesn't always fit well with graphs.

mj3cheun: Ok sure I will work something out

mj3cheun:
Ok have made all requested changes and added the padding-relative-to feature. What do you think?

documentation/md/style.md
```diff
@@ -168,7 +168,8 @@ Padding:
...
 
-* **`padding`** : The amount of padding around all sides of the node.
+* **`padding`** : The amount of padding around all sides of the node. Either percentage or pixel value can be specified. For example, both `50%` and `50px` are acceptable values. By default, percentage padding is calculated as a percentage of node width.
+* **`padding-relative-to`** : Determines how padding is calculated if and only if the percentage unit is used. If `width` is specified, the padding will be calculated as a percentage of the node width. If `height` is specified, the padding will be calculated as a percentage of the node height. If `whAverage` is specified, the padding will be calculated as a percentage of the average of the node height and width. If `whMin` is specified, the padding will be calculated as a percentage of the smallest number between width and height. If `whMax` is specified, the padding will be calculated as a percentage of the largest number between width and height.
```
d2fong: I think bullet form would be easier to read as opposed to a sequence of if statements.
e.g

* ```whAverage```:  padding will be the average of the node height and width
* ```whMin```:  ....

It will be easier for people to understand what these properties actually do.

src/collection/dimensions.js
```diff
@@ -266,6 +266,33 @@ fn = elesfn = ({
         };
       }
 
+      function computePaddingValues( width, height, paddingObject, relativeTo ) {
+        // Assuming percentage is number from 0 to 1
```
d2fong: If you assume the number is from 0 to 1, where do you convert a users value from '50%' to 0.5?

src/collection/dimensions.js
```diff
@@ -266,6 +266,33 @@ fn = elesfn = ({
...
+        if(paddingObject.units === '%') {
+          return {
```
maxkfranz: Using a `switch` statement would increase performance here and it would make sense to use `default:`

src/collection/dimensions.js
```diff
@@ -925,6 +954,16 @@ defineDimFns( {
   name: 'height'
 } );
 
+elesfn.padding = function(){
+  var ele = this[0];
+  var _p = ele._private;
+  if( _p.autoPadding !== undefined ) {
```
maxkfranz: Need to return zero if the collection is length zero

src/style/properties.js
```diff
@@ -33,6 +33,7 @@ var styfn = {};
     bidirectionalSize: { number: true }, // allows negative
     bidirectionalSizes: { number: true, multiple: true }, // allows negative
     sizeMaybePercent: { number: true, min: 0, allowPercent: true },
+    paddingRelativeTo: { enums: [ 'width', 'height', 'whAverage', 'whMin', 'whMax' ] },
```
maxkfranz: Style property values should always be `dash-case` rather than `camelCase`.

I think `width, height, average, min, max` is fine

src/collection/dimensions.js
mj3cheun: When a percentage value is parsed, the % sign is removed and the resulting number is divided by 100. This parsed value is retrieved by `ele.pstyle( 'padding' )`. Technically if you put in 1000%, it will still work though, you will just get a padding larger than the width or whatever, so maybe I should remove that comment.

d2fong: Ok.  Maybe the comment is fine,  I just didn't understand how the pfValue worked in this case.

mj3cheun:
Ok done

src/collection/dimensions.js
maxkfranz: `.pstyle()` / `.parsedStyle()` gives the parsed style property

`.pstyle().pfValue` is the value in preferred units that can be used for calculations easily

`.pstyle().value` is the value as specified in the stylesheet (e.g. `20` for `20%`)

src/collection/dimensions.js
```diff
@@ -925,6 +951,21 @@ defineDimFns( {
   name: 'height'
 } );
 
+elesfn.padding = function(){
+  if(this.length > 0) {
+    var ele = this[0];
+    var _p = ele._private;
+    if( _p.autoPadding !== undefined ) {
```
maxkfranz: This check isn't safe.  A node can change from parent to non-parent and the value would still be defined

maxkfranz: @d2fong 

I assume we're supporting percent padding only for compound parents.  That means we should check that the element is both a parent and it's padding is in percent units to use the autopadding.

We also need to note then that the percent padding is only for compounds in the docs.  I don't think it would be very useful for non-compound nodes, anyway.

mj3cheun: Sorry I misunderstood your earlier change request. I have made changes so that it uses autoPadding if it is parent. Autopadding then only returns the pvalue if it is in pixel units, so a check might not be necessary.

maxkfranz: The autopadding could already be defined if changed from a parent node to a non-parent node, so the percent check is still necessary

src/collection/dimensions.js
```diff
@@ -925,6 +951,21 @@ defineDimFns( {
   name: 'height'
 } );
 
+elesfn.padding = function(){
+  var ele = this[0];
+  if(ele.isParent()) {
+    var _p = ele._private;
+    if( _p.autoPadding !== undefined ) {
```
maxkfranz: Parent check

Issue #1541, opened 2016-09-21
Replace `cy.onRender()` and `cy.offRender()` with a `render` event


maxkfranz:
Because supporting `.onRender()` for backwards compatibility is trivial, the old code will still work.  But you should still update, anyway.

maxkfranz:
Already implemented.

PR #1713, opened 2017-02-17
add documentation for the render event

josephst:
`onRender()` and `offRender()` are being replaced by using the `render` event #1541 

In Cytoscape.js v3, `onRender(fn)` should be `cy.on('render', fn)`.


maxkfranz:
Great, would you please make the PR on the 3.x branch instead?

josephst:
Done. #1720

Issue #1348, opened 2016-04-27
Migrate demos to repo


maxkfranz:
We don't really need Jsbin support anymore, since we host the demos live on the docs site now.  This keeps things more centralised and removes the need for using gists.

- [x] Disable gulp target to download new demos
- [x] Change folder names of old demos to meaningful and short names (lowercase, dash-separated); also needs thumbnail filenames updated
- [x] Make symlinks for old folder names (gist IDs)
- [x] May need to update `docmaker.js` and `docmaker.json`

maxkfranz:
@mj3cheun This would be a good issue to look at, since it's related to the new demos you made.  Basically, it's converting the existing demos to ones stored primarily in the repo rather than pulled from gists.

mj3cheun:
For the record, I am using the following hacked together bash script to do the renaming. This script automates all tasks stated above.

```
newName=cytoscape-spread
oldName=${PWD##*/}
oldFolder=$PWD
newFolder=$oldFolder/../$newName
mkdir $newFolder
mv $oldFolder/* $newFolder
cd ../$newName
newFolder=$PWD
cd ..
rm -rf $oldFolder
ln -s $newFolder $oldFolder
cd ../img/demos
mv $PWD/$oldName.png $PWD/$newName.png
cd ../..
perl -pi -e "s/"$oldName"/"$newName"/g" docmaker.json
cd $newFolder
```

PR #1722, opened 2017-02-22
Migrate demos to repo #1348

mj3cheun:


documentation/demos/57e7cc43342193d9d21a
```diff
@@ -0,0 +1 @@
+/Users/manfredcheung/cytoscape.js/documentation/demos/performance-tuning
```
maxkfranz: Links like this need to be relative rather than absolute dirs

maxkfranz: Looks like I reviewed an old version; merging...

PR #1732, opened 2017-02-24
Coseoverlap

PeterTSmith:
Finished the update, I wanna check it over one more time but this one seems goood


maxkfranz:
@PeterTSmith Great!  The PR needs to go on the unstable branch rather than master though

Issue #1747, opened 2017-03-09
Allow for background image properties to be specified as multiples (i.e. arrays or space separated strings)


maxkfranz:
To allow for multiple background images per node, the simplest solution is to make each background image style property of type multiple.  Each index in the arrays corresponds to an image.

- [x] Change each image property type to have the multiple flag.
- [x] The node drawing function needs to iterate over the images.
- [x] This needs to be tested.  There may be edge cases or race conditions with loading the images, for example.

maxkfranz:
@mj3cheun This would be a great issue to do while I'm reviewing your extensions changes.  Dylan needs it for his SBGN lib

PR #1751, opened 2017-03-15
Allowed multiple URLs in background-image property #1747

mj3cheun:
Did basic testing and did not notice anything odd. If the top image has transparency, one can see the image underneath. They are basically all stacked on each other. Also designed changes to be resistant to race-condition bugs, however, cannot be sure none exist as they are hard to test.


maxkfranz:
I'm getting errors on the debug page and it seems that the `background-image` property isn't getting the proper result for its parsed object representation.

I made commit ad1337dc23bab7013882ac2a7c68b785d6a2a97e to give you an example of how the property definition should work.  If you need multiple support on a type but the type could potentially be used singly elsewhere, then a plural type with a similar spec should be defined.  See `url` and `urls` as an example.  (It should also simplify your `url[][]` code to `url[]`.)

You'll have to `git rebase ad1337dc23bab7013882ac2a7c68b785d6a2a97e` or merge to put the change in your fork.

Most of the background image properties also need to have multiple variants or be made multiple themselves:

```js
    // node background images
    { name: 'background-image', type: t.url }, // urls
    { name: 'background-image-crossorigin', type: t.bgCrossOrigin }, // chg to mult
    { name: 'background-image-opacity', type: t.zeroOneNumber }, // zeroOneNumbers
    { name: 'background-position-x', type: t.bgPos }, // chg to mult
    { name: 'background-position-y', type: t.bgPos },// chg to mult
    { name: 'background-repeat', type: t.bgRepeat }, // chg to mult
    { name: 'background-fit', type: t.bgFit }, // chg to mult
    { name: 'background-clip', type: t.bgClip }, // this makes sense as once per node
    { name: 'background-width', type: t.bgWH }, // chg to mult
    { name: 'background-height', type: t.bgWH }, // chg to mult
```

Basically, each image needs to be defined independently for the most part.

Does that make sense?

mj3cheun:
Ok I am done making my fixes. Note that the tests fail because getting one numeric value for a property with potentially multiple values does not make any sense. Those tests may have to be modified.

src/collection/style.js
```diff
@@ -71,6 +71,25 @@ var elesfn = ({
     }
   },
 
+  // get the associated values at the specified index when parsing a property with a multiple tag
+  parsedStyleMult: function( property, index ){
+    var ele = this[0];
```
maxkfranz: - This is too expensive to run for reading properties.
- It's potentially mutating the style for the element.
- It removes other needed info by only giving you the `value`.  You often need the `pfValue` and/or the `units` as well.

You can't have a general function like this for style properties.  

You should define a general `util.arrayValueAt( array, index, fallbackIndex )` function in `util` and use `util.arrayValueAt( ele.pstyle('foo').value, i, 0 )`.

src/extensions/renderer/canvas/drawing-nodes.js
```diff
@@ -155,12 +152,21 @@ CRp.drawNode = function( context, node, shiftToOriginWithBb, drawLabel ){
...
-  if( urlDefined ){
-    if( image.complete ){
-      this.drawInscribedImage( context, image, node );
+  var prevBging = _p.backgrounding;
+  var totalCompleted = 0;
+
+  for( var i = 0; i < numImages; i++ ){
+    if( ( urlDefined[i] ) && image[i].complete ){
+      totalCompleted++;
+      this.drawInscribedImage( context, image, node, i );
```
maxkfranz: You should be passing in `image[i]`.  The function doesn't need an array of images.

src/collection/style.js
mj3cheun: Ok, so at the moment, if the index is not found, it defaults to the default settings in getDefaultProperty. So it is ok to default to index 0 instead?

maxkfranz: If you want to use the default you need to ` || defaultPropertyValue` for each property as it's read (`pfValue`, `value`, etc.).  You can't use the approach you have above, especially because you can't ever modify `style`.  Only the styler does that, and nothing changes the properties once they're created.

Issue #1651, opened 2016-12-19
Reference `eles.move()` in the docs where `source`, `target`, and `parent` fields are mentioned


maxkfranz:
This should help clarify that nodes and edges can effectively be moved, as it's not so clear without thoroughly searching through the docs.

Issue #1734, opened 2017-02-27
Note font loading in init docs


HugoSalt:
In the init section of the docs, mention that fonts must be loaded prior to cy load to guarantee custom font usage.

OP follows:

---

I am using CytoscapeJs 2.7.15, I have loaded in the CSS of my html page a custom font as such :

```
  @font-face {
            font-family: 'Source Code Pro';
            font-weight: 400;
            font-style: normal;
            font-stretch: normal;
            src: url('fonts/SourceCodePro-Regular.otf.woff2') format('woff2');
        }
```
I then initialize the graph with this font for my nodes label :

  ```
style: { 
                        ...
                        'label': 'data(id)',            
                        'font-family': 'Source Code Pro',
                        'font-weight': '400',
                        'font-size': '0.5em',
                        ...
                    }
```

And the label gets cropped at a certain zoom level, past this zoom level the label is well displayed.
Here's a [video](http://filebin.ca/3DqvD3vKBHfL/CytoscapeCrop.webm) of the problem, and here are screenshots :

Not cropped when close enough :
![notcropped](https://cloud.githubusercontent.com/assets/25799656/23369441/ad28a72c-fd11-11e6-9324-070ab6c8f84e.png)
Cropped past a certain zoom level :
![cropped](https://cloud.githubusercontent.com/assets/25799656/23369460/b6ce3170-fd11-11e6-8c17-66e0314ceebb.png)

I have tried using .ttf instead of .woff2 format but it doesn't solve the issue.

The issue doesn't appear when I don't specify a custom font

HugoSalt:
Actually this seems to be the same issue as #1692 , but for me it's not linked to whereas the font is set to Bold or not, it's the fact of using this custom font that triggers this issue

maxkfranz:
Just as in the other issue, you need to insure your fonts are loaded before Cytoscape is initialised.

HugoSalt:
Indeed this was the issue thanks a lot ! If there's no plan to implement in Cytoscape the checking that fonts are loaded I guess both issues can be closed. But maybe it would be worth it adding to the doc (for eg in the initialization chapter) a mention that we must wait for fonts to be loaded before initializing ?

[...]

PR #1757, opened 2017-03-24
Fix for the issue #1734 and the issue #1651

Anishka0107:
This pull request aims to fix the issue cytoscape#1734
It provides a note for font loading in the init section of the docs.


Anishka0107:
The next three commits aim to fix the issue https://github.com/cytoscape/cytoscape.js/issues/1651

They have a reference added to the eles.move() method in the documentation for both the the webpage and the .md files

maxkfranz:
Thanks.  Here's some feedback:

- Any changes for a future release (that's not a bug fix, minor version x.y.**z**) goes on unstable.
- Only md and json files should be changed to update the docs.  The html is autogenerated from those files.

PR #1760, opened 2017-03-29
implements cutrectangle node shape

d2fong:
- cutrectangle shape can be a compound or simple node
- adds a util.push method for arrays
- refactors math.polygonIntersectLine to allow polygon intersect
  checking without transforming the base points
- adds 'cutrectangle' to nodeshape enums
- adds comments for ambiguous arguments in some math functions


maxkfranz:
Great; thanks

PR #1761, opened 2017-03-29
Make sure raf is defined to fix IE9 issue.

dee-see:
IE9 has `window` defined but none of the `requestAnimationFrame` properties so we end up with an `undefined` `raf` variable and cytoscape breaks whenever `util.requestAnimationFrame` is called.

I reordered the variable assignment a bit to use the "no window" value as a default value whenever all the `requestAnimationFrame` properties don't exist.


maxkfranz:
Great; thanks

Issue #1626, opened 2016-11-28
Option to specify whether labels are included in overlap detection for layouts


maxkfranz:
Including labels can make the graph more spaced out than desired in some cases.  Having an option would give more control over the output.

`options.nodeDimensionsIncludeLabels` : If `true`, the width of a node is considered to be `node.boundingBox().w`.  If `false`, the width of a node is considered to be `node.outerWidth()`.  The analogous cases for height also hold.

Uses `node.layoutDimensions()`

Ref https://github.com/cytoscape/cytoscape.js/pull/1784

- [x] Update bundled layouts
- [x] Update first-party extensions
  - [x] Dylan
    - [x] arbor
    - [x] cose-bilkent
    - [x] spread
    - [x] springy
    - [x] cola
  - [x] Max
    - [x] klay
    - [x] euler
    - [x] dagre
- [x] Add the option in the slush-cytoscape-extension generator layout template files

d2fong:
cose-bilkent in progress - https://github.com/cytoscape/cytoscape.js-cose-bilkent/pull/53

PR #1784, opened 2017-04-17
Added bb calculation function for layouts

mj3cheun:


documentation/md/collection/layoutPositions.md
```diff
@@ -12,6 +12,7 @@ var options = {
   eles: someCollection, // collection of elements involved in the layout; set by cy.layout() or eles.layout()
   fit: true, // whether to fit the viewport to the graph
   padding: 30, // padding to leave between graph and viewport
+  nodeDimensionsIncludeLabels: false, // Boolean which changes whether label dimensions are included when calculating node dimensions
```
maxkfranz: We need to document the `node.layoutDimensions()` function.  While technically the option could also be read in `layoutPositions()`, the `layoutPositions()` function doesn't use it.

documentation/docmaker.json
```diff
@@ -1639,6 +1639,23 @@
                 }
               ],
               "md": "collection/layoutPositions"
+            },
+
+            {
+              "name": "nodes.layoutDimensions",
+              "descr": "Returns the node width and height",
```
maxkfranz: Though it's important to be concise in descriptions, it's also important to be clear enough to give users a sense of how one function differs from another.  I think users might confuse this function with ones like `width()`, `height()`, and `boundingBox()`.  Maybe it's worth mentioning that it's for layout positioning (e.g. overlap detection) in the top description?

By the way, you can run `gulp docs` to see your spec changes reflected in the docs page.

maxkfranz:
Great!  Thanks

mj3cheun:
Note: Ensure external layouts are also updated to use the new function

maxkfranz:
Updated w/ checklist : https://github.com/cytoscape/cytoscape.js/issues/1626

Issue #1781, opened 2017-04-13
Background image position percentage -- include padding in the calculation


d2fong:
New array style property (per image): 

- `background-width-relative-to: inner | include-padding`
- `background-height-relative-to: inner | include-padding`

Controls whether the width or height should include the padding in positioning, sizing (e.g. cover, contain), etc.

OP follows:

---

Background images don't include padding when ```background-position-x``` or ```background-position-y``` is set as a percent value

i.e ```background-position-x```: ```'100%'``` means 100% of the node inner width.  This should be changed to be 100% of the node outer width (include padding)

PR #1785, opened 2017-04-17
Fix for background image percentage padding bug #1781

mj3cheun:
Hey @d2fong try this and see if it fixes your problem.

Fixes issue https://github.com/cytoscape/cytoscape.js/issues/1781.


src/extensions/renderer/canvas/drawing-images.js
```diff
@@ -81,14 +81,14 @@ CRp.drawInscribedImage = function( context, img, node, index ){
 
   var x = (nodeX - nodeW / 2); // left
   if( getIndexedStyle( node, 'background-position-x', 'units', index ) === '%' ){
-    x += (nodeW - w) * getIndexedStyle( node, 'background-position-x', 'pfValue', index );
+    x += (node.outerWidth() - w) * getIndexedStyle( node, 'background-position-x', 'pfValue', index );
```
maxkfranz: It should replace `nodeW` instead I think, as the same value needs to be used for clipping, repeating, etc.

mj3cheun:
Changes made

PR #1786, opened 2017-04-17
Added documentation for multiple images #1747

mj3cheun:
Documentation for https://github.com/cytoscape/cytoscape.js/issues/1747.


maxkfranz:
Looks great, but it may be worth mentioning that it's best to use arrays for the multiple values.  At least the main `background-image` property could potentially have a space in one image (especially if it's an embedded svg data uri).

mj3cheun:
Also, add an example.

PR #1794, opened 2017-04-21
Fixed percentage padding image draw bug

mj3cheun:
@d2fong @maxkfranz Fixes problem discussed [here](https://github.com/cytoscape/cytoscape.js/commit/94e1d9cfe4b6191d42bb52fee9ea8cc358641ad1#commitcomment-21858090).


d2fong:
It works when I specify padding as a percent now

mj3cheun:
Good to hear!

src/util/style.js
```diff
@@ -0,0 +1,32 @@
...
+        return 0;
+    }
+  } else if( paddingObject.units === 'px' ) {
+    return paddingObject.pfValue;
+  } else {
+    return 0;
+  }
+};
+
+util.getIndexedStyle = function( node, property, subproperty, index ){
```
maxkfranz: Should be `ele` instead of `node`

src/util/style.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+var util = module.exports = {};
+
+util.computePaddingValues = function( width, height, paddingObject, relativeTo ){
```
maxkfranz: This function can actually stay in the dims / bb file since it's the only place it should be used

src/util/style.js
maxkfranz: It might make sense to put this in the style / stylesheet object instead of util, as it's style related rather than general object util

src/extensions/renderer/canvas/drawing-images.js
```diff
@@ -23,9 +23,11 @@ CRp.drawInscribedImage = function( context, img, node, index ){
   var nodeY = pos.y;
   var fit = getIndexedStyle( node, 'background-fit', 'value', index );
   var repeat = getIndexedStyle( node, 'background-repeat', 'value', index );
-  var paddingX2 = node.pstyle( 'padding' ).pfValue * 2;
-  var nodeW = node.width() + ( getIndexedStyle( node, 'background-width-relative-to', 'value', index ) === 'inner' ? 0 : paddingX2 );
-  var nodeH = node.height() + ( getIndexedStyle( node, 'background-height-relative-to', 'value', index ) === 'inner' ? 0 : paddingX2 );
+  var nodeW = node.width();
+  var nodeH = node.height();
+  var paddingX2 = util.computePaddingValues(nodeW, nodeH, node.pstyle( 'padding' ), node.pstyle( 'padding-relative-to' ).value ) * 2;
+  var nodeTW = nodeW + ( getIndexedStyle( node, 'background-width-relative-to', 'value', index ) === 'inner' ? 0 : paddingX2 );
```
maxkfranz: I don't think the renderer should be redoing calculations like this.  Can't we use the `node.padding()` function to read the precalculated value?

mj3cheun: I forgot the `node.padding()` function existed. However, that function doesn't exactly solve the problem. Here is the definition of the padding function above:

```
elesfn.padding = function(){
  var ele = this[0];
  var _p = ele._private;
  if( ele.isParent() ){
    ele.updateCompoundBounds();

    if( _p.autoPadding !== undefined ){
      return _p.autoPadding;
    } else {
      return ele.pstyle('padding').pfValue;
    }
  } else {
    return ele.pstyle('padding').pfValue;
  }
}
```
As you can see, it does not make special allowances for percentage padding values. However, what I can do is update the padding function to use `computePaddingValues(nodeW, nodeH, node.pstyle( 'padding' ), node.pstyle( 'padding-relative-to' ).value )` which should solve issue, and then go ahead and use `.padding()` for everything.

mj3cheun: PS: _p.autoPadding only works for parent nodes. Images can be placed on child nodes.

maxkfranz: Updating the compound bounds should update all the auto values.

I think percent padding is only supposed to work for parent nodes.  There's not much of a usecase of using percent paddings for non-parent nodes, because they're not dynamic like parents.  The point of having functions like `width()` and `padding()` is that you don't want to repeat style rules over several files --- a change just needs to update one function.

If it's not documented already, we should mention in the docs that percent padding is for compounds only.

maxkfranz: I selected the wrong line.  It's this one that can be replaced with a `padding()` call: 

```
var paddingX2 = util.computePaddingValues(nodeW, nodeH, node.pstyle( 'padding' ), node.pstyle( 'padding-relative-to' ).value ) * 2;
```

mj3cheun: Ok I see so percentage is only for use on parent nodes so autopadding is fine. Will update commit thanks!

mj3cheun:
@maxkfranz @d2fong PR rewrite done. Dylan, you want to test it to make sure it still works for you? Thanks.

src/extensions/renderer/canvas/drawing-images.js
```diff
@@ -21,11 +18,14 @@ CRp.drawInscribedImage = function( context, img, node, index ){
   var pos = node.position();
   var nodeX = pos.x;
   var nodeY = pos.y;
+  var getIndexedStyle = node.cy().style().getIndexedStyle;
```
maxkfranz: Needs `.bind()`

d2fong:
Okay I'll get back to you after I test it.

Issue #1583, opened 2016-10-20
Box select when starting over a node if modifier key down


ugurdogrusoz:
I think it's very critical that one should be able to start a box selection (Ctrl and/or Shift + select) within a compound (sometimes compound nodes are rather big like many processes within a cellular compartment). Currently this starts a drag of the compound!

maxkfranz:
Possible solutions:

(1) When a modifier key is down (e.g. shift), a node can't be dragged if box selection is enabled.  Instead, box selection starts.

(2) When a modifier key is down (e.g. shift), a *compound* node can't be dragged if box selection is enabled.  Instead, box selection starts.

I would lean towards (1), because it's more consistent.  What do you think?

ugurdogrusoz:
I agree I think (1) is better

maxkfranz:
(1) implemented on unstable

maxkfranz:
Closing since it's implemented; re-open if need be

ugurdogrusoz:
@markgoffin This is great! But there is a problem. I realize that the rule is all nodes/edges overlapping with our selection box get selected. But if we need to select a proper subset of children of a compound and say move them, we can't do this using this box selection since the compound itself is selected! If the box selection starts and ends inside a compound, can we perhaps make an exception and not select the compound? I can't image a scenario where the user is indeed interested in selecting both the compound and a proper subset of the compound's children.

maxkfranz:
There are a couple of approaches I see:

(1) Select a parent only if the box encloses it.

(2) Select a parent if there's any overlap with the box, unless the box is contained by the parent.

Case (1) avoids selection of parents who overlap each other for the case you mentioned, whereas (2) does not.  Case (1) makes selecting children by box generally easier than (2).  However, (1) is not consistent with how box selection works for non-compound nodes.  

What do you think?

ugurdogrusoz:
In fact (2) is exactly what I tried to describe. We only add an exception to the current box selection that we do not select the parent when the box is contained by the parent. Thanks.

ugurdogrusoz:
@maxkfranz @mj3cheun We upgraded our application (http://www.cs.bilkent.edu.tr/~ivis/ChiSE_daily/) to use Cytoscape.js 3.1 but box selection inside a compound still selects the compound as well?

maxkfranz:
I just tested it and it's working in the 3.1 release.  You're probably using an older release

maxkfranz:
[...]

PR #1800, opened 2017-04-25
If selection box is contained in parent do not select #1583

mj3cheun:
Solves issue brought up in #1583.

### DIFF
--- a/src/extensions/renderer/base/coord-ele-math/coords.js
+++ b/src/extensions/renderer/base/coord-ele-math/coords.js
@@ -333,7 +333,7 @@ BRp.getAllInBox = function( x1, y1, x2, y2 ){
         includeLabels: false
       } );
 
-      if( math.boundingBoxesIntersect( boxBb, nodeBb ) ){
+      if( math.boundingBoxesIntersect( boxBb, nodeBb ) && !math.boundingBoxInBoundingBox( nodeBb, boxBb ) ){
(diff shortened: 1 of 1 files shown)


src/extensions/renderer/base/coord-ele-math/coords.js
```diff
@@ -333,7 +333,7 @@ BRp.getAllInBox = function( x1, y1, x2, y2 ){
         includeLabels: false
       } );
 
-      if( math.boundingBoxesIntersect( boxBb, nodeBb ) ){
+      if( math.boundingBoxesIntersect( boxBb, nodeBb ) && !math.boundingBoxInBoundingBox( nodeBb, boxBb ) ){
```
maxkfranz: Should this second part be considered only for compounds?

mj3cheun: @maxkfranz In testing I have not noticed any weirdness with normal nodes. I have no idea how it is set up but `!math.boundingBoxInBoundingBox( nodeBb, boxBb )` is true if `boxBb` inside `nodeBb`. If `nodeBb` is essentially a point value, this condition is virtually impossible to meet.

maxkfranz: That's a good point, so it's probably fine as-is.  Thanks!

PR #1812, opened 2017-05-05
Cleanup code

josephst:
Various small changes to files with no bug fixes/ feature changes.


src/collection/algorithms/dijkstra.js
```diff
@@ -74,7 +74,7 @@ var elesfn = ({
 
       knownDist[ uid ] = smalletsDist;
 
-      if( smalletsDist === Math.Infinite ){
+      if( smalletsDist === undefined ){
```
maxkfranz: This is a good catch, but it should be checking `Infinity` rather than `Math.Infinite`.  Looks like a typo

josephst: I initially had it changed to `Infinity` but that was causing tests to fail. Going to take another look at this to see why that is.

josephst:
Okay I think I figured it out. The tests were failing because of the `break` command on the next line. Previously, the comparison `smalletsDist === Math.Infinite` was never true, so the `break` command was never hit. Changing the comparison to `=== Infinity` meant that the comparison was true for disconnected nodes; this then caused `break` to exit the while-loop early, causing calculated distances to be incorrect.

Changed `break` to `continue` so that only the current loop iteration is terminated rather than the entire loop. I think this is the correct behavior; plus, it might speed up some calculations since the current iteration of the loop can be ended here instead of after finding that `u.neighborhood().intersect( nodes )` is empty, as (I think) happened previously when the comparison was always false.

Issue #1814, opened 2017-05-08
Image 404 with background-image and background-image-fit breaks graph


tumri:
I am currently working with a large data set(few thousand nodes and edges).

I am initializing cytoscape as follows:

```
var cy = cytoscape({
                    container: document.getElementById('cy'), // container to render in
                    elements: getPos(),
                    style: cytoscape.stylesheet()
                                        .selector('node')
                                            .css({
                                                'height': 200,
                                                'width': 200,
                                                'background-image': 'data(node_image)',
                                                'background-fit': 'cover',
                                                'border-color': '#000',
                                                'border-width': 3,
                                                'border-opacity': 0.5
                                            }),
                    minZoom: 0.01,
                    maxZoom: 10.0,
                });
```

When any of the thousands of node images returns a 404 error the entire graph layout breaks and I get the following error:

`Uncaught DOMException: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The HTMLImageElement provided is in the 'broken' state.`

I was able to fix the error by making the following change:

```
CRp.safeDrawImage = function( context, img, ix, iy, iw, ih, x, y, w, h ){
  var r = this;

  // detect problematic cases for old browsers with bad images (cheaper than try-catch)
  if( iw <= 0 || ih <= 0 || w <= 0 || h <= 0){
    return;
  }

  context.drawImage( img, ix, iy, iw, ih, x, y, w, h );
};
```

to

```
CRp.safeDrawImage = function( context, img, ix, iy, iw, ih, x, y, w, h ){
  var r = this;

  // detect problematic cases for old browsers with bad images (cheaper than try-catch)
  if( iw <= 0 || ih <= 0 || w <= 0 || h <= 0 || !img.width || !img.height){
    return;
  }

  context.drawImage( img, ix, iy, iw, ih, x, y, w, h );
};
```

maxkfranz:
Some images won't have dimensions defined (like some svgs) but they can be used if an explicit background width and height are set.  I'm not sure whether the change you posted could be used as it stands if it causes issues with existing usecases.

Maybe it would be better to handle the 404 at the load event itself.

tumri:
@maxkfranz - I modified my PR #1815 to check if image exists and if image height and width are both === 0 in drawing_nodes.js where you have the check for image[i].complete.


`this.drawInscribedImage( context, image[i], node, i );`

to

```
if (!image[i] || (image[i].height === 0 && image[i].width === 0)) {
[...]

PR #1815, opened 2017-05-08
Image 404 Breaks Graph (Issue #1814)

tumri:
Fixes render error when image 404s.


src/extensions/renderer/base/images.js
```diff
@@ -10,7 +10,7 @@ BRp.getCachedImage = function( url, crossOrigin, onLoad ){
   if( cache ){
     if( !cache.image.complete ){
       cache.image.addEventListener('load', onLoad);
-      cache.image.addEventListener('error', function(){ cache.image.width = cache.image.height = 0; });
+      cache.image.addEventListener('error', function(){ cache.image.renderable = cache.image.renderable = false; });
```
maxkfranz: The duplicate assignment isn't needed with one property now

src/extensions/renderer/base/images.js
```diff
@@ -19,7 +19,7 @@ BRp.getCachedImage = function( url, crossOrigin, onLoad ){
 
     var image = cache.image = new Image(); // eslint-disable-line no-undef
     image.addEventListener('load', onLoad);
-    image.addEventListener('error', function(){ image.width = image.height = 0; });
+    image.addEventListener('error', function(){ image.renderable = image.renderable = false; });
```
maxkfranz: Same with this prop

src/extensions/renderer/canvas/drawing-nodes.js
```diff
@@ -148,7 +148,9 @@ CRp.drawNode = function( context, node, shiftToOriginWithBb, drawLabel ){
   for( var i = 0; i < numImages; i++ ){
     if( ( urlDefined[i] ) && image[i].complete ){
       totalCompleted++;
-      this.drawInscribedImage( context, image[i], node, i );
+      if ( !(image[i].renderable == false) ) {
```
maxkfranz: Maybe better to name the prop `error` so it's not a double negative condition

tumri:
Changed as requested.

tumri:
Changed as requested.

maxkfranz:
@tumri OK, I squashed and cherry-picked across master and unstable.  But the extra listener in the existing cache case isn't needed so I removed it.

PR #1822, opened 2017-05-15
Double arrow and circle shape

xMarkusSpringerx:
I added two new shapes for edges.
- circle (just the border)
- double arrow


maxkfranz:
Thanks for the pull request!

Some feedback:

- [ ] Because these would be new features, they would have to go in the `unstable` branch.
- [ ] You can already make any arrow shape hollow, so probably there doesn't need to be an extra circle shape: http://js.cytoscape.org/#style/edge-arrow  A shape should not change the other visual styles like colour.
- [ ] The `double-arrow` canvas implementation is identical to the `triangle-tee` one.  Rather than copy-pasting the implementation, it should be pulled out and given a generic name so both use the same implementation.

xMarkusSpringerx:
Ok I'll have a look at that.
@maxkfranz do you know a know where I can additional info to a node which is only displayed when I click on the node? Like a toggle box which can be clicked to get detailed info to a node.

maxkfranz:
The [qtip extension](https://github.com/cytoscape/cytoscape.js-qtip) covers that usecase.

xMarkusSpringerx:
The qtip extension only covers the case when I want to show a tooltip. I want to rebuild something like that
[gorgeouscode](https://gorgeouscode-alpha.herokuapp.com/github/itmiguelfernandes/apiflat_book). When I doubleclick on an entity it show the additional info. @maxkfranz do you have any ideas how to do this?

maxkfranz:
You can use one or more svg backgrounds if you want custom node rendering (either whole node or just decorations on it): http://js.cytoscape.org/#style/background-image

maxkfranz:
@xMarkusSpringerx Do you need any pointers to help you with your PR?

xMarkusSpringerx:
@maxkfranz it's ok. My problem is way more complex as I thought. I switched to plain d3.js.
Should I just pull out the canvas implementation of the double arrow and delete circle shape so that you can close the pr with at least one additional feature?

maxkfranz:
@xMarkusSpringerx Yes, I think it would be good to have the double arrow.  Thanks

PR #1852, opened 2017-06-05
concavehexagon, tag, bottomroundrectangle shape implementations

d2fong:
Adds three new shapes
- bottomroundrectangle
- tag
- concave hexagon

### DIFF
--- a/src/extensions/renderer/base/node-shapes.js
+++ b/src/extensions/renderer/base/node-shapes.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var math = require('../../../math');
-var util = require('../../../util');
 
 var BRp = {};
 
@@ -57,13 +56,7 @@ BRp.generateEllipse = function(){
     },
 
     checkPoint: function( x, y, padding, width, height, centerX, centerY ){
-      x -= centerX;
-      y -= centerY;
-
-      x /= (width / 2 + padding);
-      y /= (height / 2 + padding);
-
-      return x * x + y * y <= 1;
+      math.checkInEllipse( x, y, padding, width, height, centerX, centerY );
     }
   } );
 };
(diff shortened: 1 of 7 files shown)


maxkfranz:
@d2fong Do you have pictures of each node shape so I can have a sense of what the code intends to do?

d2fong:
screen shot 2017-06-06 at 1 36 09 pm

Here is a pic of the three shapes

d2fong:
Left:  bottomroundrectangle Middle: concavehexagon Right: tag

d2fong:
This is a continuation of the work to port some of the polygon shapes from the sbgn lib to cy.js.

concavehexagon and tag use the generatePolygon function.  bottomroundrectangle uses a similar approach for checkpoint and line intersection like the roundrectangle shape.

src/extensions/renderer/base/node-shapes.js
```diff
@@ -348,6 +348,104 @@ BRp.generateBarrel = function(){
...
+        x, y,
+        nodeX,
+        nodeY,
+        width, height,
+        padding )
+      ;
+    },
+
+    // Looks like the width passed into this function is actually the total width / 2
+    checkPoint: function(
```
maxkfranz: I think the full width is supposed to be passed here.  Could we debug this Wed?

src/extensions/renderer/base/node-shapes.js
```diff
@@ -348,6 +348,104 @@ BRp.generateBarrel = function(){
...
+        centerX - outerWidth, centerY,
+        centerX + outerWidth, centerY,
+        centerX + outerWidth, centerY - outerHeight
+      ];
+      if( math.pointInsidePolygonPoints( x, y, points)){
+        return true;
+      }
+
+
+      var checkInEllipse = function( x, y, centerX, centerY, width, height, padding ){
```
maxkfranz: Is is possible to use the same function as the roundrect or could this function be pulled out to `math` and possibly have the roundrect function use `checkInEllipse`?

d2fong:
Okay, we can review in person on Wednesday.

PR #1857, opened 2017-06-07
Use webpack

d2fong:
adds webpack tooling to enable external dependency compatibility.


maxkfranz:
TODO 

- Inspect test failures on old versions of node
- Remove unused browserify dev dependencies from package.json

PR #1866, opened 2017-06-21
document the includeOverlays option

josephst:
I ran into an issue while working on the Popper plugin; I eventually tracked it down to unknowingly leaving `includeOverlay` in it's default (true) value. I think the option should be documented, since it's a (relatively) user-facing option. There's also a `useCache` option, but since I can't think of any (non-development) use of it, I'm not documenting it.


maxkfranz:
It's probably a good idea to document this

maxkfranz:
Whoops, hit the button accidentally.

It's probably a good idea to document this, but for the extension it might make sense to use outer width and outer height instead.  That would work more   There could be an option for the user to use a bb instead and pass in bb options, but it would be on them to set the options properly.

maxkfranz:
Merged into unstable/3.2

PR #1899, opened 2017-07-18
Refactored element classes to use Set

alexcli:
Related: #1843 

Performance stays relatively constant with the exception of eles.classes() and eles.flashClass() which have become slower.


package.json
```diff
@@ -135,8 +135,10 @@
     "first": "images"
   },
   "dependencies": {
+    "bluebird": "^3.5.0",
     "gulp-download": "0.0.1",
     "heap": "^0.2.6",
+    "i": "^0.3.5",
```
maxkfranz: Is this a typo?

src/collection/class.js
```diff
@@ -5,13 +5,13 @@ let elesfn = ({
     classes = ( classes || '' ).match( /\S+/g ) || [];
     let self = this;
     let changed = [];
-    let classesMap = {};
+    let classesMap = new Set();
```
maxkfranz: You need to `require()` the `set.js` file under `/src` so the polyfill is used in case the code is running in an old environment where `Set` is not supported.

src/collection/class.js
```diff
@@ -72,7 +67,7 @@ let elesfn = ({
 
   hasClass: function( className ){
     let ele = this[0];
-    return ( ele != null && ele._private.classes[ className ] ) ? true : false;
+    return ( ele != null && ele._private.classes.has(className) ) ? true : false;
```
maxkfranz: Your change looks fine on this line, but `? true : false` could be removed to make it shorter.

src/collection/class.js
```diff
@@ -34,14 +33,10 @@ let elesfn = ({
...
-          let eleCls = classes[i];
-          let eleHasClass = eleClasses[ eleCls ];
-          let specdClass = classesMap[ eleCls ]; // i.e. this class is passed to the function
-
-          if( eleHasClass && !specdClass ){
+          let specdClass = classesMap.has(eleCls);
+          if( !specdClass ){
             changedEle = true;
             break;
           }
```
maxkfranz: It doesn't make sense to `util.copy()` a `Set`, since it's not a plain object (`{}`).  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

One way to copy a set is `new Set( oldSet.values() )`, but you might want to experiment with different ways of copying if that's slow.

PR #1949, opened 2017-08-23
Multi mount

d2fong:
# Note
- This branch still needs testing, I am opening this PR to get your feedback about implementation i.e should it be implemented like this or not.

## Functional Changes
- adds mount() to the core
- adds unmount() to the core
- mount can be called multiple times

## Low Level Details
- ~~~changes the Core constructor to wrap a coreInit function. coreInit handles all of the things that were previously handled in the Core constructor.~~~
- ~~~coreInit is a common function used by both mount and the Core constructor~~~
- mount mutates renderer options

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -243,7 +243,7 @@ util.extend( corefn, {
   },
 
   headless: function(){
-    return this._private.options.renderer.name === 'null';
+    return this._private.renderer.isHeadless();
   },
 
   styleEnabled: function(){
@@ -280,7 +280,9 @@ util.extend( corefn, {
       container = container[0];
     }
   
-    cy.unmount();
+    cy.stopAnimationLoop();
+    
+    cy.destroyRenderer();
 
     _p.container = container;
     _p.styleEnabled = true;
@@ -296,8 +298,10 @@ util.extend( corefn, {
     let cy = this;
 
     cy.stopAnimationLoop();
-
+    
     cy.destroyRenderer();
+
+    cy.initRenderer( { name: 'null' } );
   },
 
   options: function(){
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -5,10 +5,15 @@ let is = require( '../is' );
 let Promise = require( '../promise' );
 let define = require( '../define' );
 
-let coreInit = function( cy, opts ){
-  let options = util.extend( {}, opts );
-  
-  let container = options.container;
+let Core = function( opts ){
+  let cy = this;
+
+  opts = util.extend( {}, opts );
+
+  let container = opts.container;
+
+  // allow for passing a wrapped jquery object
+  // e.g. cytoscape({ container: $('#cy') })
   if( container && !is.htmlElement( container ) && is.htmlElement( container[0] ) ){
     container = container[0];
   }
@@ -23,11 +28,12 @@ let coreInit = function( cy, opts ){
   }
 
   let readies = reg.readies = reg.readies || [];
-  
+
   if( container ){ container._cyreg = reg; } // make sure container assoc'd reg points to this cy
   reg.cy = cy;
-  
+
   let head = window !== undefined && container !== undefined && !opts.headless;
+  let options = opts;
   options.layout = util.extend( { name: head ? 'grid' : 'null' }, options.layout );
   options.renderer = util.extend( { name: head ? 'canvas' : 'null' }, options.renderer );
 
@@ -41,18 +47,17 @@ let coreInit = function( cy, opts ){
     }
   };
 
-  let _p = cy._private = {
+  let _p = this._private = {
     container: container, // html dom ele container
     ready: false, // whether ready has been triggered
     options: options, // cached options
-    elements: new Collection( cy ), // elements in the graph
+    elements: new Collection( this ), // elements in the graph
     listeners: [], // list of listeners
-    aniEles: new Collection( cy ), // elements being animated
+    aniEles: new Collection( this ), // elements being animated
     scratch: {}, // scratch object for core
     layout: null,
     renderer: null,
     destroyed: false, // whether destroy was called
-    mounted: head,  // whether mount was called
     notificationsEnabled: true, // whether notifications are sent to the renderer
     minZoom: 1e-50,
     maxZoom: 1e50,
@@ -77,7 +82,7 @@ let coreInit = function( cy, opts ){
     hasCompoundNodes: false
   };
 
-  cy.createEmitter();
+  this.createEmitter();
 
   // set selection type
   let selType = options.selectionType;
@@ -190,13 +195,6 @@ let coreInit = function( cy, opts ){
   } );
 };
 
-let Core = function( opts ){
-
-  let cy = this;
-
-  coreInit( cy, opts );
-};
-
 let corefn = Core.prototype; // short alias
 
 util.extend( corefn, {
@@ -280,13 +278,38 @@ util.extend( corefn, {
   mount: function( container ){
     let cy = this;
     let _p = cy._private;
-    if( container ){
-      cy.destroyRenderer();
-      _p.options.container = container;
-      _p.options.renderer.name = 'canvas';
-      _p.options.headless = false; 
-      coreInit(cy, _p.options);
+    let options = _p.options;
+
+    if( container == null ){ return; }
+
+    if( !is.htmlElement( container ) && is.htmlElement( container[0] ) ){
+      container = container[0];
     }
+  
+    cy.stopAnimationLoop();
+    cy.destroyRenderer();
+
+    options.container = container;
+    _p.container = container;
+
+    options.styleEnabled = true;
+    _p.styleEnabled = true;
+
+    options.renderer.name = 'canvas';
+    options.headless = false; 
+
+    cy.initRenderer( util.extend( {
+      hideEdgesOnViewport: options.hideEdgesOnViewport,
+      textureOnViewport: options.textureOnViewport,
+      wheelSensitivity: is.number( options.wheelSensitivity ) && options.wheelSensitivity > 0 ? options.wheelSensitivity : 1,
+      motionBlur: options.motionBlur === undefined ? false : options.motionBlur, // off by default
+      motionBlurOpacity: options.motionBlurOpacity === undefined ? 0.05 : options.motionBlurOpacity,
+      pixelRatio: is.number( options.pixelRatio ) && options.pixelRatio > 0 ? options.pixelRatio : undefined,
+      desktopTapThreshold: options.desktopTapThreshold === undefined ? 4 : options.desktopTapThreshold,
+      touchTapThreshold: options.touchTapThreshold === undefined ? 8 : options.touchTapThreshold
+    }, options.renderer ) );
+    cy.startAnimationLoop();
+    cy.style( options.style );    
   },
 
   options: function(){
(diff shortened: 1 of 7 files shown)


src/core/index.js
```diff
@@ -5,15 +5,10 @@ let is = require( '../is' );
...
-let Core = function( opts ){
-  let cy = this;
-
-  opts = util.extend( {}, opts );
-
-  let container = opts.container;
-
-  // allow for passing a wrapped jquery object
-  // e.g. cytoscape({ container: $('#cy') })
+let coreInit = function( cy, opts ){
```
maxkfranz: I think `coreInit` is doing too much.  Probably `mount()` only needs a subset of what's going on here.  For example, re-mounting would destroy all the existing elements...

maxkfranz:
Looks like generally the right approach, but I think `coreInit` needs to be revisited.

src/core/index.js
```diff
@@ -79,42 +95,23 @@ let Core = function( opts ){
...
-  // init zoom bounds
-  if( is.number( options.minZoom ) && is.number( options.maxZoom ) && options.minZoom < options.maxZoom ){
-    _p.minZoom = options.minZoom;
-    _p.maxZoom = options.maxZoom;
-  } else if( is.number( options.minZoom ) && options.maxZoom === undefined ){
-    _p.minZoom = options.minZoom;
-  } else if( is.number( options.maxZoom ) && options.minZoom === undefined ){
-    _p.maxZoom = options.maxZoom;
-  }
+  _p.elements = new Collection( cy );
```
d2fong: Not sure if I should be creating a new Collection every time coreInit (i.e mount) is called, or if this should only be called once during the Core constructor.

d2fong:
WIP.  Changed coreInit to not destroy existing elements, etc. Load external data only on Core constructor call, not on calls to coreInit.

maxkfranz:
I think it may make sense to leave the core constructor the way it was.  Remounting shouldn't be doing most of init work anyway -- the state should remain between mounts.  Only the renderer instance should change.  

So it's probably best to duplicate whatever you need in `mount()` and maybe consider refactoring later.

Also we probably don't need a separate `mounted` flag.  We already have `headless()`.

d2fong:
Okay, ill just copy over what is needed by mount()

src/core/index.js
```diff
@@ -183,11 +200,11 @@ let Core = function( opts ){
       }
 
       // bind all the ready handlers registered before creating this instance
-      for( let i = 0; i < readies.length; i++ ){
-        let fn = readies[ i ];
+      for( let i = 0; i < options.reg.readies.length; i++ ){
```
d2fong: @maxkfranz 
When I call ```mount()```, do I need to:
* call the ready functions associated with the container?
* call ```cy.startAnimationLoop()```

maxkfranz: Ready doesn't have anything *directly* to do with rendering: It's about when the data is loaded and the layout has finished, because either could be async.   *Indirectly*, ready happens around the same time as -- but a bit before -- the first frame.

You will have to handle the animation loop, in case you went from a headless instance (with `styleEnabled: false`) to a rendered one.  You'll need to stop the loop at the start of remounting and then start a new one towards the end of remounting:  Rendered instances need to hook into the renderer's render loop to sync animation calculations just before rendering each frame.

src/core/index.js
```diff
@@ -275,6 +275,43 @@ util.extend( corefn, {
...
+    if( container == null ){ return; }
+
+    if( !is.htmlElement( container ) && is.htmlElement( container[0] ) ){
+      container = container[0];
+    }
+  
+    cy.stopAnimationLoop();
+    cy.destroyRenderer();
+
+    options.container = container;
```
d2fong: Not sure if ```mount()``` needs to mutate the option values to keep the options and private fields in sync.

maxkfranz: Everything in options should only be used by the constructor.  The only exception would be `cy.json()`, because its output is the options.

So you probably only need to modify `_p.options.renderer`

d2fong: Alright I will remove the unnecessary operations.

src/core/index.js
```diff
@@ -275,6 +275,43 @@ util.extend( corefn, {
...
+    _p.container = container;
+
+    options.styleEnabled = true;
+    _p.styleEnabled = true;
+
+    options.renderer.name = 'canvas';
```
maxkfranz: We can't really assume this generally.  The dev can specify which renderer to use at init, so we also probably need to have a `rendererOptions` argument for `options.renderer`.

If unspecified, it should use the canvas renderer like you have here.

src/core/index.js
```diff
@@ -275,6 +275,43 @@ util.extend( corefn, {
...
+    cy.initRenderer( util.extend( {
+      hideEdgesOnViewport: options.hideEdgesOnViewport,
+      textureOnViewport: options.textureOnViewport,
+      wheelSensitivity: is.number( options.wheelSensitivity ) && options.wheelSensitivity > 0 ? options.wheelSensitivity : 1,
+      motionBlur: options.motionBlur === undefined ? false : options.motionBlur, // off by default
+      motionBlurOpacity: options.motionBlurOpacity === undefined ? 0.05 : options.motionBlurOpacity,
+      pixelRatio: is.number( options.pixelRatio ) && options.pixelRatio > 0 ? options.pixelRatio : undefined,
+      desktopTapThreshold: options.desktopTapThreshold === undefined ? 4 : options.desktopTapThreshold,
+      touchTapThreshold: options.touchTapThreshold === undefined ? 8 : options.touchTapThreshold
+    }, options.renderer ) );
```
maxkfranz: As for the later refactoring part, we probably just need to address this line (301-310).

The defaults handling could be moved into `cy.initRenderer()` so you'd just pass in `options.renderer` here and in the constructor.

maxkfranz:
We may also want to have `cy.unmount()` that does the starting bits of `cy.mount()`.  You may need to unmount some time before you can remount, for example.

src/core/index.js
```diff
@@ -275,6 +266,40 @@ util.extend( corefn, {
...
   },
 
+  mount: function( container, rendererOptions ){
+    if( container == null ){ return; }
+
+    let cy = this;
+    let _p = cy._private;
+    let options = _p.options;
+    
+    let rOpts = util.extend( {}, options.renderer, { name: 'canvas' }, rendererOptions );
```
maxkfranz: Should we override instead of extend?  If I specified renderer options it would only be to specify everything new from scratch.

src/core/renderer.js
```diff
@@ -42,9 +43,19 @@ let corefn = ({
...
+      motionBlurOpacity: 0.05,
+      desktopTapThreshold: 4,
+      touchTapThreshold: 8,
+      wheelSensitivity: 1
+    };
+
+    let rOpts = util.extend( {}, defaults, options, { 
+      cy: cy,
+      wheelSensitivity: is.number( options.wheelSensitivity ) && options.wheelSensitivity > 0 ? options.wheelSensitivity : defaults.wheelSensitivity,
+      pixelRatio: is.number( options.pixelRatio ) && options.pixelRatio > 0 ? options.pixelRatio : undefined
```
maxkfranz: `undefined` would be set with `util.extend()` --- which behaves like `Object.assign()`

Probably better to fall back on `defaults` instead of `undefined` here.

src/core/index.js
d2fong: I am not sure.  If you specify everything new from scratch then your options should override everything right?  If someone wants to only specify a few new options this approach would also support that I think.

src/core/renderer.js
```diff
@@ -42,9 +43,20 @@ let corefn = ({
       return;
     }
 
-    let rOpts = util.extend( {}, options, {
-      cy: cy
-    } );
+    let defaults = {
+      motionBlur: false,
+      motionBlurOpacity: 0.05,
+      pixelRatio: undefined,
```
d2fong: pixelRatio had a undefined default value from before these changes.  Not sure if there is a more appropriate default value.

src/core/index.js
maxkfranz: Say I did

```js
let cy = new Cytoscape({ container: div1, renderer: { name: 'canvas', foo: 'bar', baz: 'bat' } }); // 1

cy.mount( div2, { name: 'webgl', baz: 'bop' } ); // 2
```

At 2, we have `{ name: 'webgl', baz: 'bop', foo: 'bar' }` but I didn't ask for `{ foo: 'bar' }`.

maxkfranz: The only usecase for setting the renderer options differently than init is if you want to change the rendering technology / renderer class used.  If you do that, you probably don't want old values around -- they could potentially conflict in the worst case.

src/core/renderer.js
maxkfranz: You're right. The correct value is undefined. My mistake. It's because the 'auto'/nil value can only be calculated later by the renderer.

The change makes line 58 clearer, though it's not really necessary.

src/core/index.js
d2fong: So get rid of ```options.renderer``` in the extend call?

maxkfranz: Makes sense.

src/core/index.js
```diff
@@ -275,6 +266,40 @@ util.extend( corefn, {
...
   },
 
+  mount: function( container, rendererOptions ){
+    if( container == null ){ return; }
+
+    let cy = this;
+    let _p = cy._private;
+    let options = _p.options;
+    
+    let rOpts = rendererOptions ? rendererOptions : { name: 'canvas' };
```
d2fong: Does it make sense to ignore previous renderer options for both cases (when rendererOptions is supplied and when  it is not? )

maxkfranz: I think so.  Any internal call to `mount()` that wanted the previous options taken into account would pass the previous options in, I think.

src/core/index.js
```diff
@@ -275,6 +266,40 @@ util.extend( corefn, {
...
+
+    cy.style( options.style );    
+  },
+
+  unmount: function(){
+    let cy = this;
+
+    cy.stopAnimationLoop();
+
+    cy.destroyRenderer();
```
d2fong: Do you think unmount should create a null renderer?  That is what currently happens if you set ```headless: true``` in the core initialization.  Right now unmount makes cy._private.renderer null.

If we create a null renderer on unmount, it means we would be calling initRenderer twice in mount() but it would make unmount() consistent with setting headless to true.  

We could also change the core ```headless()``` function to read ```cy._private.renderer.name``` instead of reading ```cy._private.options.renderer.name``` so that ```headless()``` returns true once ```unmount()``` is called.

maxkfranz: It's usually better to have the `NullRenderer`.  For dev-facing API, we should use a `NullRenderer`.  But that means that `mount()` shouldn't re-use `unmount()`.  We don't want a `NullRenderer` unnecessarily when we're just about to create a `CanvasRenderer`.

Re. `headless()`, I don't know if that would work generally.  Our internal renderers may follow that pattern, but I don't think that's part of any API contract.  The renderer API is already a bit loosely defined, and I'd rather keep the API space as small as possible for it until it's formally defined for external renderer extensions.  

I'd rather check something like `this._private.renderer.isHeadless()` by adding that function to the `BaseRenderer` (false) and the `NullRenderer` (true) -- so that future renderer extensions don't have to worry about defining special functions or properties.

Issue #1644, opened 2016-12-07
Allow classes to be specified as arrays


maxkfranz:
If the dev maintains an array of class names, this saves him from having to `.join()` every time he uses the API.

- At init: `{ data: { id: 'foo' }, classes: ['bar', 'baz'] }`
- In `eles.classes()` : `eles.classes(['bar', 'baz'])`
- In `eles.toggleClass()` / `eles.removeClass()` / `eles.addClass()` : `eles.addClass(['foo', 'bar'])`

Tasks

- [x] Add changes to allow for arrays as input
- [x] Update the docs to indicate that passing arrays is supported

PR #1984, opened 2017-10-01
Allow classes to be specified as arrays #1644

rjmill:
This implements #1644. I think this is just about ready, but I wanted to open it up for maintainer input before going any farther.

Do you have any opinions on what to do with arrays of strings with spaces in them - e.g.
```
[ 'foo', 'bar baz' ]
```

This PR takes care of classes at initialization, `eles.classes()`, `eles.addClass()`, `eles.removeClass()`, and `eles.toggleClass()`. Is there anything else that needs to support arrays of classes?

### DIFF
--- a/src/collection/class.js
+++ b/src/collection/class.js
@@ -1,8 +1,15 @@
 let Set = require('../set');
+let is = require('../is');
 
 let elesfn = ({
   classes: function( classes ){
-    classes = ( classes || '' ).match( /\S+/g ) || [];
+    if( is.array( classes ) ){
+      // use a copy of the array
+      classes = classes.slice(0);
+    } else {
+      // extract classes from string
+      classes = ( classes || '' ).match( /\S+/g ) || [];
+    }
     let self = this;
     let changed = [];
     let classesSet = new Set( classes );
@@ -57,8 +64,14 @@ let elesfn = ({
     return ( ele != null && ele._private.classes.has(className) );
   },
 
-  toggleClass: function( classesStr, toggle ){
-    let classes = classesStr.match( /\S+/g ) || [];
+  toggleClass: function( classes, toggle ){
+    if( is.array( classes ) ){
+      // use a copy of the array
+      classes = classes.slice(0);
+    } else {
+      // extract classes from string
+      classes = classes.match( /\S+/g ) || [];
+    }
     let self = this;
     let toggleUndefd = toggle === undefined;
     let changed = []; // eles who had classes changed
--- a/src/collection/class.js
+++ b/src/collection/class.js
@@ -3,10 +3,7 @@ let is = require('../is');
 
 let elesfn = ({
   classes: function( classes ){
-    if( is.array( classes ) ){
-      // use a copy of the array
-      classes = classes.slice(0);
-    } else {
+    if( !is.array( classes ) ){
       // extract classes from string
       classes = ( classes || '' ).match( /\S+/g ) || [];
     }
@@ -65,10 +62,7 @@ let elesfn = ({
   },
 
   toggleClass: function( classes, toggle ){
-    if( is.array( classes ) ){
-      // use a copy of the array
-      classes = classes.slice(0);
-    } else {
+    if( !is.array( classes ) ){
       // extract classes from string
       classes = classes.match( /\S+/g ) || [];
     }
--- a/src/collection/element.js
+++ b/src/collection/element.js
@@ -88,8 +88,7 @@ let Element = function( cy, params, restore ){
       _p.classes.add(cls);
     }
   } else if( is.array( params.classes ) ){
-    // use a copy of the array
-    let classes = params.classes.slice(0);
+    let classes = params.classes;
     for( let i = 0, l = classes.length; i < l; i++ ){
       let cls = classes[ i ];
       if( !cls || cls === '' ){ continue; }
--- a/src/collection/element.js
+++ b/src/collection/element.js
@@ -87,6 +87,15 @@ let Element = function( cy, params, restore ){
 
       _p.classes.add(cls);
     }
+  } else if( is.array( params.classes ) ){
+    // use a copy of the array
+    let classes = params.classes.slice(0);
+    for( let i = 0, l = classes.length; i < l; i++ ){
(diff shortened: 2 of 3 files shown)


maxkfranz:
@rjmill Thanks for the PR.

> Do you have any opinions on what to do with arrays of strings with spaces in them

It's probably fine to just keep it simple (and less expensive) by just using whatever the user specifies in the array.  If we start to split strings within an array, the dev may expect that you could do this too: 

```js
eles.toggleClass(['foo', ['bar', 'baz']]);
```

> Is there anything else that needs to support arrays of classes?

Those are the most important ones.  I think `eles.flashClass()` would make sense too.

src/collection/class.js
```diff
@@ -1,8 +1,15 @@
 let Set = require('../set');
+let is = require('../is');
 
 let elesfn = ({
   classes: function( classes ){
-    classes = ( classes || '' ).match( /\S+/g ) || [];
+    if( is.array( classes ) ){
+      // use a copy of the array
+      classes = classes.slice(0);
```
maxkfranz: I don't think the copying is necessary.  I don't think we mutate this list in functions like `toggleClass()`, and whatever json is specified for an element at init is considered owned by that element.  So it would be cheaper to avoid copying, if possible.

rjmill: That makes sense. I didn't check to see if it gets mutated or not, but if I remember correctly, the array gets put into a Set pretty quickly anyway. I'll remove the copy.

rjmill:
>just using whatever the user specifies in the array.

Sounds good!

>I think eles.flashClass() would make sense too.

I'll take a look at it.

src/collection/class.js
```diff
@@ -3,10 +3,7 @@ let is = require('../is');
 
 let elesfn = ({
   classes: function( classes ){
-    if( is.array( classes ) ){
-      // use a copy of the array
-      classes = classes.slice(0);
-    } else {
+    if( !is.array( classes ) ){
```
rjmill: It's probably more sensible to check if it's a string instead of if it's not an array, but this way will preserve the current error behavior if someone puts something that's not a string in here (ie throwing errors about doing string stuff to something that's not a string.)

rjmill:
I looked into `eles.flashClass()`, and it uses `eles.addClass()` and `eles.removeClass()` under the hood (which in turn use `eles.toggleClass()` under the hood.) So `eles.flashClass()` should "just work."

src/collection/element.js
```diff
@@ -87,6 +87,14 @@ let Element = function( cy, params, restore ){
 
       _p.classes.add(cls);
     }
+  } else if( is.array( params.classes ) ){
```
rjmill: Doing an "if/else if" where both blocks only differ by one line feels messy. I have a few refactors in mind, but I held off because I wanted to preserve as much of the original code as possible for my initial PR. Also, this is arguably simple/small enough that refactoring might not make a huge difference anyway.

src/collection/class.js
maxkfranz: Just some background info: We try to avoid copying except where absolutely necessary for performance reasons.  We've run into cases where copying causes issues before.  

Copying can affect large graphs.  For example, something like this can be common:

```js
cy.elements().forEach( ele => { // this callback would have many array copies made
  ele.toggleClass([ 'foo', 'bar' ]);

  if( eleIsBaz ){
    ele.addClass(['baz']);
  }
  
 // maybe even more class calls
} ); 
```

It all adds up, so we try to keep the lib as lean as possible to give the dev who uses the lib more flexibility.

src/collection/element.js
maxkfranz: Since the code is the same apart from assigning classes, you could remove the if/else and just use a ternary operator (`?`) inline in the declaration of `classes`.

rjmill: If I understand correctly, doing that would make it something like
```
if ( is.array( params.classes ) || is.string( params.classes ) ){
  let classes = is.string( params.classes ) ? params.classes.split( /\s+/ ) : params.classes;
  // ...
```
which still feels kind of wonky to me. One of the refactors I had in mind was pulling the for loop out of the conditional and having the conditional just set the value for the `classes` variable. If params.classes is not a string or array, it'll just default to being an empty array (and therefore would be skipped by the for loop without causing errors.)

Maybe something like
```
let classes = params.classes || [];
if ( is.string( classes ) ){
  classes = classes.split( /\s+/ );
}
for( let i = 0, l = classes.length; i < l; i++ ){
  let cls = classes[ i ];
  if( !cls || cls === '' ){ continue; }

  _p.classes.add(cls);
}
```
This is probably the cleanest option I have in mind, but it now will throw an error when you put an invalid type in for `classes` instead of silently passing through, which might break some people's code. To avoid that, I'm thinking something like
```
let classes = [];
if ( is.array( params.classes ) ){
  classes = params.classes;
} else if ( is.string( params.classes ) ){
  classes = params.classes.split( /\s+/ );
}
for( let i = 0, l = classes.length; i < l; i++ ){
  let cls = classes[ i ];
  if( !cls || cls === '' ){ continue; }

  _p.classes.add(cls);
}
```
It's not as clean as the other one, but it won't run the risk of breaking people's stuff. On the other hand, one could argue that anything that depends on passing invalid data in for `classes` is stuff that should break anyway.

At the end of the day, it's a matter maintainer preference. I'm fine with any of these options (including the option of just leaving it as is and calling it a day.)

maxkfranz: I was thinking of the third one basically:

```js
let classes = is.array(params.classes) ? params.classes : is.string(params.classes) ? params.classes.split( /\s+/ ) : null;
if( classes != null ){ for( let i = 0, l = classes.length; i < l; i++ ){
  let cls = classes[ i ];
  if( !cls || cls === '' ){ continue; }

  _p.classes.add(cls);
} }
```

rjmill:
Refactored classes initialization in elements.js to the third option we discussed. Is there anything else you'd like me to take care of for this PR? Should I go ahead and squash all these commits into one?

maxkfranz:
It looks good.  I'll just squash on merge.  Thanks

Issue #1948, opened 2017-08-23
Allow specifying the line-cap of edge lines


maxkfranz:
**Issue type**

Feature request

**Description of new feature**

`line-cap: butt | round | square` : Set the [line cap](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap) of the edge

- [x] Add the `line-cap` style property
- [x] Change the [canvas line-cap](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap) when drawing edges, based on the `line-cap`value.  Make sure to reset to `butt` after drawing the edge so that things like node shapes aren't affected.
- [x] Add documentation to `style.md`.
- [x] Make sure to note in the documentation that the line-cap extends past the endpoint.  It's not suitable for use with transparent nodes, cases where you want the line-cap shape visible, etc.

**Motivation for new feature**

For edges without arrows that point towards non-circle nodes, the squared-off ends of edges can look a bit weird.  A curved end would look like it aligns better on non-circle nodes.  It would be a simple change, because it's supported directly in canvas: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap

PR #1991, opened 2017-10-12
Add property edge-cap

Aerilius:
This pull request adds the feature to specify the edge cap (shape of the edge endings) as proposed in issue #1948. This is important to avoid "cracks" between edges when nodes are smaller than the edge width or hidden. 
Since it is based on the canvas property `line-cap`, it should not have significant impact on performance.

### DIFF
--- a/src/extensions/renderer/canvas/drawing-edges.js
+++ b/src/extensions/renderer/canvas/drawing-edges.js
@@ -27,10 +27,11 @@ CRp.drawEdge = function( context, edge, shiftToOriginWithBb, drawLabel ){
   let opacity = edge.pstyle('opacity').value;
   let lineStyle = edge.pstyle('line-style').value;
   let edgeWidth = edge.pstyle('width').pfValue;
+  let edgeCap = edge.pstyle('edge-cap').value;
 
   let drawLine = ( strokeOpacity = opacity ) => {
     context.lineWidth = edgeWidth;
-    context.lineCap = 'butt';
+    context.lineCap = edgeCap;
 
     r.strokeStyle( context, lineColor[0], lineColor[1], lineColor[2], strokeOpacity );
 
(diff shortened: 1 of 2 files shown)


maxkfranz:
Thanks for the PR.

Do you think there is value to having all of { butt, round, square }?  Since the cap extends beyond the endpoints, a non-butt cap is not visible.  Maybe we should just have { butt, square } or { butt, extend }.

Either way, we may need to be really clear about the cap positioning in the docs.  I suspect many people would expect the cap to be visible, i.e. to not extend beyond the endpoint.

Aerilius:
A line-cap is visible if the line is as wide or wider than the node.
The default `butt` ends in the center point of the node, so that two lines that are not exactly at 180° will always have a "crack", some ε° gap on one side. To cover the gap one can extend the line over the center point (ideally rounded, my usecase). 
Of course if the node is wider than the line then this is not visible. I have no usecase yet for `square`, but I thought I just reflect all possible values from the canvas API.

src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -27,10 +27,11 @@ CRp.drawEdge = function( context, edge, shiftToOriginWithBb, drawLabel ){
   let opacity = edge.pstyle('opacity').value;
   let lineStyle = edge.pstyle('line-style').value;
   let edgeWidth = edge.pstyle('width').pfValue;
+  let edgeCap = edge.pstyle('edge-cap').value;
 
   let drawLine = ( strokeOpacity = opacity ) => {
     context.lineWidth = edgeWidth;
-    context.lineCap = 'butt';
+    context.lineCap = edgeCap;
```
maxkfranz: The `lineCap` should be reset to `'butt'` at the end of the function.  

Background: We don't use `.save()` or `.restore()`, because they're expensive.  So you have to clean up the canvas state for things like lineCap, transform, etc. manually.

maxkfranz:
That makes sense.  I didn't consider the case where the edge is wider than the node.

PR #1992, opened 2017-10-14
Fix depth adjustment in breadthfirst layout

pangel:
Indexes were not updated after splicing; this could cause wrongful move to higher depths after 1st iteration.

### DIFF
--- a/src/extensions/layout/breadthfirst.js
+++ b/src/extensions/layout/breadthfirst.js
@@ -263,6 +263,10 @@ BreadthFirstLayout.prototype.run = function(){
       }
       depths[ newDepth ].push( ele );
 
+      for ( let j = info.index; j < depths[info.depth].length; j++ ){
+        depths[info.depth][j]._private.scratch.breadthfirst.index--;
+      }
+
       info.depth = newDepth;
       info.index = depths[ newDepth ].length - 1;
     }


maxkfranz:
Do you have an example where this causes an issue?  It seems to be with `index`, not `depth.  And the `index` is reset every call to `assignDepthsToEles()`, which happens right at the end of the block in question.

pangel:
`assignDepthsToEles()` is called outside of the loop which iterates over `elesToMove`, so the call to `splice` on line 257 may use a stale index if 2 elements from the same depth have to be moved and the one with the smallest index is moved before the one with the larger index.

This should trigger the issue:

```
[
  {
    "data": {
      "id": "1"
    }
  },
  {
    "data": {
      "id": "2"
    }
  },
  {
    "data": {
      "id": "3"
    }
  },
  {
    "data": {
      "id": "4"
    }
  },
  {
    "data": {
      "id": "5"
    }
  },
  {
    "data": {
      "id": "6"
    }
  },
  {
    "data": {
      "id": "7"
    }
  },
  {
    "data": {
      "id": "5-7",
      "source": "5",
      "target": "7"
    }
  },
  {
    "data": {
      "id": "5-6",
      "source": "5",
      "target": "6"
    }
  },
  {
    "data": {
      "id": "5-3",
      "source": "5",
      "target": "3"
    }
  },
  {
    "data": {
      "id": "4-6",
      "source": "4",
      "target": "6"
    }
  },
  {
    "data": {
      "id": "3-7",
      "source": "3",
      "target": "7"
    }
  },
  {
    "data": {
      "id": "1-4",
      "source": "1",
      "target": "4"
    }
  }
]
```

maxkfranz:
OK, I did something slightly different with a sparse array that gets cleaned up.

PR #2006, opened 2017-10-26
Handle nested undefined properties in styles

michaelpj:
Currently, if you use a data property in a style with a nested pattern, like `label: data(foo.bar)`, and `foo.bar` is not set, then we crash trying to look up `bar` on the undefined `data.foo`.

The code in fact tries to check whether the values it is indexing into are defined, but it only does so for the very first one (i.e. `data` itself). Probably the check should be the same as the one on line 366, which checks at each iteration of the loop, so I changed it to match.

### DIFF
--- a/src/style/apply.js
+++ b/src/style/apply.js
@@ -431,10 +431,10 @@ styfn.applyParsedProperty = function( ele, parsedProp ){
     let fields = prop.field.split( '.' );
     let fieldVal = _p.data;
 
-    if( fieldVal ){ for( let i = 0; i < fields.length; i++ ){
+    for( let i = 0; i < fields.length && fieldVal; i++ ){
       let field = fields[ i ];
       fieldVal = fieldVal[ field ];
-    } }
+    }
 
     flatProp = this.parse( prop.name, fieldVal, prop.bypass, flatPropMapping );
 


maxkfranz:
This is a necessary, but not sufficient, change to allow for non-blocking invalid mapper data.  This change would allow the issue to fall through to another error, later in the code.

See the following for a more general solution, which incorporates your PR: Fall back on the default style value if a style mapper fails #2010

michaelpj:
Happy to wait for a more thorough version - I've worked around it in my own code. I simply noticed the inconsistency here when I was debugging why I got an exception.

maxkfranz:
I think that these changes should go in the unstable branch (i.e. in a new feature release), as they affect only invalid mappers (i.e. broken code/data).  Lib patches generally fix bugs in the lib itself, not the app using the lib.  I'd also like this to have some time to be tested for a while before it's pushed out to npm.

maxkfranz:
Merged into unstable for #2010

maxkfranz:
@michaelpj Thanks for the PR!

michaelpj:
FWIW, here's why I want this to work: I'm working on a graph viewer that will accept user graph data, and I'd like to use some pre-specified properties for styles, labels, etc. However, it should also work if the user doesn't provide those. So gracefully degradation is very much a desirable feature in this case.

maxkfranz:
I would validate the user input, if I were you.  You can never trust the user.

michaelpj:
Sure - but in this case I want the properties to be optional. An alternative would be to have some kind of explicit marking for optional accesses/fallbacks, but I guess you don't want to complicate the styling language too much.

maxkfranz:
A function may be better suited in that case:

```js
'background-color': el => isValid( getColor(el) ) ? getColor(el) : 'red'
```

You may want to memoize.

Issue #2005, opened 2017-10-26
Unbundled bezier edges sometimes don't render


epatters:
**Issue type**

Bug report

**Environment info**

- Cytoscape.js version : 3.2.1 - 3.2.4
- Browser/Node.js & version : Google Chrome 61.0.3163.100, Node.js 8.7.0

**Current (buggy) behaviour**

Since version 3.2.1, my unbundled Bezier edges are rendering unreliably. Sometimes they appear, sometimes they don't. In version 3.1.5, I don't have this problem.

The issue is possibly related to https://github.com/cytoscape/cytoscape.js/issues/1451

**Minimum steps to reproduce**

This example (not minimal--sorry!) illustrates the problem:

http://jsbin.com/hurisituta

If you drag the nodes around you can see the Bezier edges appear and disappear.

jack5678:
Sad

epatters:
Thank you both for the quick fix! I will give this a whirl when you cut the next release.

maxkfranz:
I was planning on releasing 3.2.5 today anyway, so I'll just close this for now and publish.

PR #2007, opened 2017-10-26
Fix checkInEllipse signature

joeybrk372:
The `checkInEllipse` function was refactored into the math util but the signature was changed. Updated signature to be consistent with the other functions and then fixed a call in `ellipse.checkPoint`.

Fixes issue with disappearing bezier edges (https://github.com/cytoscape/cytoscape.js/issues/2005), and a more minimal example: http://jsbin.com/rohebeqivi/2/edit?js,output

### DIFF
--- a/src/extensions/renderer/base/node-shapes.js
+++ b/src/extensions/renderer/base/node-shapes.js
@@ -54,7 +54,7 @@ BRp.generateEllipse = function(){
     },
 
     checkPoint: function( x, y, padding, width, height, centerX, centerY ){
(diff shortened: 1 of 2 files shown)


joeybrk372:
Whoops, looks like this has already been addressed in: https://github.com/cytoscape/cytoscape.js/commit/0954ada89789a296d442e6f2343c4d91a76fef38. Closing

maxkfranz:
A couple of us were working on the fix but we didn't get a chance to push it for a couple of hours.  Sorry for the duplicated effort.  We appreciate the PR anyway!

PR #2012, opened 2017-11-02
Rewrite commonjs with es-modules and replace webpack with rollup

TrySound:
Before
```
  cytoscape.cjs.js   760.3K
  cytoscape.js       779.6K
  cytoscape.min.js   312.1K
```

After
```
  cytoscape.cjs.js    704.7K
  cytoscape.js        724.0K
  cytoscape.min.js    287.7K
```


src/emitter.js
```diff
@@ -28,7 +29,7 @@ const defaults = {
   parent: function( /*context*/ ){
     return null;
   },
-  context: this
+  context: window
```
TrySound: Here I decided to use window module, because this in es modules is not allowed.

maxkfranz: `context: null` would be clearer as we also support a headless mode (Node.js).

maxkfranz:
Thanks for the PR.  Here are some initial points of discussion:

- [ ] All non-bugfix changes, like this PR,  should go in the unstable branch.  I think the changes in unstable for 3.3 are just additions (not major refactoring), so your changes should probably just be able to be replayed onto unstable.  Since unstable is probably just a superset of master, there may still be some (new) `require()`s that need to be addressed.
- [ ] Being able to run Mocha tests without having to do a build is a large benefit to productivity, and that wasn't possible with ES modules before.  Perhaps now [`@std/esm`](https://github.com/standard-things/esm) could address that usecase.
- [ ] The ESLint config needs to be updated to support ES modules.
- [ ] `cytoscape.Stylesheet` is undefined.  The debug page doesn't work, for example.
- [ ] We have `"main": "dist/cytoscape.cjs.js"` in `package.json`.  We probably should also have [`"module": "src/index.js"`](http://2ality.com/2017/07/npm-packages-via-babel.html) for ES-module-aware systems.
- [ ] We probably have to always build with format `"umd"`.  Unfortunately, there are still people who use AMD.
- [ ] We should probably use the `.mjs` extension so that things work natively in upcoming versions of Node.
- [ ] Should probably use something like `commonjs( Object.keys( pkg.dependencies || {} ).map( pkg => `node_modules/${pkg}/**` ) )` so new dependencies can be added without modifying the Rollup config.
- [ ] Is there a reason we can't use `.babelrc`?

rollup.config.js
```diff
@@ -0,0 +1,62 @@
...
+  input: './src/index.js',
+  output: {
+    file: path.resolve(__dirname, 'build', FILENAME),
+    name: 'cytoscape',
+    format: NODE_ENV === 'production' ? 'cjs' : 'umd',
+    sourcemap: SOURCEMAPS ? 'inline' : false
+  },
+  external: NODE_ENV === 'production' ? Object.keys( pkg.dependencies || {} ) : [],
+  plugins: [
+    nodeResolve(),
```
maxkfranz: This might not be necessary if we use `.mjs`.

rollup.config.js
```diff
@@ -0,0 +1,62 @@
...
+    commonjs({
+      include: [
+        'node_modules/lodash.debounce/**',
+        'node_modules/heap/**'
+      ]
```
maxkfranz: This could be automated like `external` so new deps can be added without changes to the config.

rollup.config.js
```diff
@@ -0,0 +1,62 @@
...
+var SOURCEMAPS = boolEnv('SOURCEMAPS', false);
+var pkg = require('./package.json');
+var path = require('path');
+
+module.exports = {
+  input: './src/index.js',
+  output: {
+    file: path.resolve(__dirname, 'build', FILENAME),
+    name: 'cytoscape',
+    format: NODE_ENV === 'production' ? 'cjs' : 'umd',
```
maxkfranz: Probably always need umd

src/index.js
```diff
@@ -32,9 +33,8 @@ cytoscape.use = function( ext ){
 };
 
 // replaced by build system
-cytoscape.version = require('./version');
+cytoscape.version = version;
 
 // expose public apis (mostly for extensions)
-cytoscape.stylesheet = cytoscape.Stylesheet = Stylesheet;
```
maxkfranz: It looks like this line was missed

src/test.js
```diff
@@ -10,5 +10,6 @@ if( process.env.TRAVIS || process.env.TEST_BUILD ){
 // Otherwise use the unbundled, unbabelified, raw source
 // NB : Must use a version of Node that natively supports all JS features used in the lib
 } else {
-  module.exports = require('./index.js');
+  require('babel-register');
```
maxkfranz: Is `babel-register` strictly necessary or could we get by with `@std/esm`?

maxkfranz:
Here's what I get when the missing stylesheet is used:

```
$ ls -l build
total 3496
-rw-r--r--  1 max  staff  731039  2 Nov 18:51 cytoscape.cjs.js
-rw-r--r--  1 max  staff  750783  2 Nov 18:51 cytoscape.js
-rw-r--r--  1 max  staff  299438  2 Nov 18:51 cytoscape.min.js
```

src/index.js
maxkfranz: When I revise this line, I get the following in the debug page while `npm run watch`ing:

```
Uncaught TypeError: Cannot read property 'finiteLinesIntersect' of undefined
```

src/emitter.js
TrySound: So context is not used since it's always null?

maxkfranz: The context is used in each emit with `Function.apply()`.  If it's `null`, it would be `window` in the browser and `global` in Node.

rollup.config.js
TrySound: This is necessary to resolve `index.js` as node specific rule and `node_modules` dependencies for dev mode.

TrySound:
mjs is still discussed. And considered by many not necessary.
Watch has a problem yet, so we can wait a bit. https://github.com/standard-things/esm/issues/138 As a workaround we can test only bundled version since only entry point is used in tests.

rollup.config.js
TrySound: There is already umd here with bundled dependencies. This version is clean node module. Also umd will require global aliasing for missing deps.

maxkfranz: The current UMD made by Webpack allows for using external dependencies with each of globals, cjs, amd.  Is this not possible with Rollup?  

Would Rollup be able to automatically support cjs and globals with external dependencies?  If we go this route (dropping amd support), we'd have to release a new major version (v4).

maxkfranz:
Closing this one as https://github.com/cytoscape/cytoscape.js/pull/2084 is newer

PR #2044, opened 2018-01-21
Add support for `line-color: linear-gradient(auto, ...colors)`

elisherer:
Add support for linear gradients on edges as color.

The syntax is:
`line-color: linear-gradient(auto, color1, color2, ...)`

- `auto` - currently is the only direction supported, between source and target points.
- At least 2 colors are required.
- Color can be: Color Name, RGBA, HSLA, Hex3 or Hex6

Example:
cytoscape-gradient-pr

### DIFF
--- a/src/extensions/renderer/canvas/drawing-edges.js
+++ b/src/extensions/renderer/canvas/drawing-edges.js
@@ -32,8 +32,18 @@ CRp.drawEdge = function( context, edge, shiftToOriginWithBb, drawLabel ){
     context.lineWidth = edgeWidth;
     context.lineCap = 'butt';
 
-    r.strokeStyle( context, lineColor[0], lineColor[1], lineColor[2], strokeOpacity );
-
+    if (Array.isArray(lineColor[0])) { //check if gradient
+      let grd = context.createLinearGradient(rs.allpts[0], rs.allpts[1], rs.allpts[rs.allpts.length-2], rs.allpts[rs.allpts.length-1]);
+      let length = lineColor.length;
+      for (let i = 0; i < length ; i++) {
+        grd.addColorStop(i / (length - 1), 'rgba(' + lineColor[i][0] + ',' + lineColor[i][1] + ',' + lineColor[i][2] + ',' + strokeOpacity + ')');
+      }
+      context.strokeStyle = grd;
+    }
+    else {
+      r.strokeStyle(context, lineColor[0], lineColor[1], lineColor[2], strokeOpacity);
+    }
+    
     r.drawEdgePath(
       edge,
       context,
--- a/src/style/parse.js
+++ b/src/style/parse.js
@@ -374,6 +374,19 @@ styfn.parseImpl = function( name, value, propIsBypass, propIsFlat ){
       bypass: propIsBypass
     };
 
+  } else if( type.colorOrGradient ) {
+    let tuple = util.colorOrGradient2tuple( value );
+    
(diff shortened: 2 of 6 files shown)


src/extensions/renderer/canvas/drawing-edges.js
```diff
@@ -32,8 +32,18 @@ CRp.drawEdge = function( context, edge, shiftToOriginWithBb, drawLabel ){
     context.lineWidth = edgeWidth;
     context.lineCap = 'butt';
 
-    r.strokeStyle( context, lineColor[0], lineColor[1], lineColor[2], strokeOpacity );
-
+    if (Array.isArray(lineColor[0])) { //check if gradient
+      let grd = context.createLinearGradient(rs.allpts[0], rs.allpts[1], rs.allpts[rs.allpts.length-2], rs.allpts[rs.allpts.length-1]);
```
maxkfranz: (5) If you're not iterating, you shouldn't use `rs.allpts`.  Use `rs.source` and `rs.target`.

src/style/parse.js
```diff
@@ -374,6 +374,19 @@ styfn.parseImpl = function( name, value, propIsBypass, propIsFlat ){
...
+    if ( !tuple ){ return null; }
+
+    return {
+      name: name,
+      value: tuple,
+      pfValue: tuple,
+      strValue: '' + value,
+      bypass: propIsBypass
+    };
+
```
maxkfranz: Also (4):

(4) The method used for checking whether you have a colour or a gradient will be bad for maintenance.  It would be better to use a flag in the style property, and it seems that the style property parsing could be handled a bit cleaner.  For example, there should be no `colorOrGradient2tuple`.  Colours and gradients should be parsed separately so the style property can better represent which one is detected.

maxkfranz:
@elisherer This looks like a great start.  Here are some things to consider:

(0) New features don't go in the master branch.  Only bug fixes do.  This belongs in the unstable branch.

(1) Colour handling shouldn't really be happening in the edge drawing function.  There's a more general function that sets stroke colour, and that should be amended instead.

(2) Supporting gradients means that we should support them wherever a colour property can be specified.  That means support for gradients in `strokeStyle()` and `fillStyle()`.

(3) By (1) and (2), you'd almost already have support for gradients on nodes.  Although you yourself probably only need gradients on edges for your usecase, I wouldn't consider a gradient feature complete unless it's universal.

(4) The method used for checking whether you have a colour or a gradient will be bad for maintenance.  It would be better to use a flag in the style property, and it seems that the style property parsing could be handled a bit cleaner.  For example, there should be no `colorOrGradient2tuple`.  Colours and gradients should be parsed separately so the style property can better represent which one is detected.

Further, you can't assume the browser has `Array.isArray()`.  That's a relatively new feature, and you break support for old browsers if you do that.  Look at the `is.js` file if you want to do same type checking.  And what if radial gradients were supported in future?  The array check would be ambiguous.

(5) If you're not iterating, you shouldn't use `rs.allpts`.  Use `rs.source` and `rs.target`.

(6) Using gradients means that other features, such as animations and style mappers, no longer work on those elements.

(7) This is going to be a very expensive feature that will not work well for larger graphs.

This is my initial review pass.  I think this is an interesting feature, but (6) and (7) are a bit concerning.

maxkfranz:
@elisherer Could you explain your usecase a bit so I can understand the motivation for gradients a bit better?

elisherer:
Replaced with #2058

PR #2058, opened 2018-01-30
Add support for `line/background-color-style: linear/radial-gradient`

elisherer:
Add support for linear gradients on edges as color.

The syntax is:

~~`line-color: linear-gradient(auto, color1, color2, ...)`~~

- ~~`auto` - currently is the only direction supported, between source and target points.~~
- ~~At least 2 colors are required.~~
- ~~Color can be: Color Name, RGBA, HSLA, Hex3 or Hex6~~

**New suggestion**:

New style property type `colorStyle` (enum)
Values: `none` (default), `linear-gradient`, `radial-gradient`

All new properties will be:

~~`line-color-style: linear-gradient;` (type `colorStyle`)~~
~~`line-gradient-stop-1-color: red;` (type `color`)~~
~~`line-gradient-stop-2-color: green;` (type `color`)~~
~~(we could add `stop-#-offset` in the future)~~
...

`line-fill: linear-gradient;` (**new** type: `fill` (**solid**, linear-gradient or radial-gradient))
`line-gradient-stops-colors: red blue ...;` (type: `colors`)
`line-gradient-stops-positions: 0 ... 100%;` (type: `percentages`) *new; multiple of percent

`background-fill: linear-gradient;` (type `fill`)
`background-gradient-direction: to-right;` (**new enum type** `gradientDirection`) values can be: `to-bottom` (default), `to-top`, `to-left`, `to-right`, `to-bottom-right`, `to-bottom-left`, `to-top-right`, `to-top-left`.
`background-gradient-stops-colors: red blue ...;` (type: `colors`)
`background-gradient-stops-positions: 0 ... 100%;` (type: `percentages`)

Example:
cytoscape-gradient-pr

Original comments by @maxkfranz (made on #2044)

@elisherer This looks like a great start. Here are some things to consider:

(0) New features don't go in the master branch. Only bug fixes do. This belongs in the unstable branch.

(1) Colour handling shouldn't really be happening in the edge drawing function. There's a more general function that sets stroke colour, and that should be amended instead.

(2) Supporting gradients means that we should support them wherever a colour property can be specified. That means support for gradients in strokeStyle() and fillStyle().

(3) By (1) and (2), you'd almost already have support for gradients on nodes. Although you yourself probably only need gradients on edges for your usecase, I wouldn't consider a gradient feature complete unless it's universal.

(4) The method used for checking whether you have a colour or a gradient will be bad for maintenance. It would be better to use a flag in the style property, and it seems that the style property parsing could be handled a bit cleaner. For example, there should be no colorOrGradient2tuple. Colours and gradients should be parsed separately so the style property can better represent which one is detected.

Further, you can't assume the browser has Array.isArray(). That's a relatively new feature, and you break support for old browsers if you do that. Look at the is.js file if you want to do same type checking. And what if radial gradients were supported in future? The array check would be ambiguous.

(5) If you're not iterating, you shouldn't use rs.allpts. Use rs.source and rs.target.

(6) Using gradients means that other features, such as animations and style mappers, no longer work on those elements.

(7) This is going to be a very expensive feature that will not work well for larger graphs.

This is my initial review pass. I think this is an interesting feature, but (6) and (7) are a bit concerning.


elisherer:
First of all, thanks for the comprehensive review,
Second, as shown in the screenshot, I need a fast way to show the direction of money flow (red to green), and the arrows are not enough.

**New suggestion**:

New style property type `colorStyle` (enum)
Values: `none` (default), `linear-gradient`, `radial-gradient`

All new properties will be:

`line-color-style: linear-gradient;` (type `colorStyle`)
`line-gradient-stop-1-color: red;` (type `color`)
`line-gradient-stop-2-color: green;` (type `color`)
(we could add `stop-#-offset` in the future)

`background-color-style: linear-gradient;` (type `colorStyle`)
`background-gradient-direction: to right;` (**new enum type** `gradientDirection`) values can be: `to bottom` (default), `to top`, `to left`, `to right`, `to bottom right`, `to bottom left`, `to top right`, `to top left`.
`background-gradient-stop-1-color: red;` (type `color`)
`background-gradient-stop-2-color: green;` (type `color`)

`background-color: radial-gradient;` (type `colorStyle`)
`background-gradient-stop-1-color: red;` (type `color`)
`background-gradient-stop-2-color: green;` (type `color`)

elisherer:
I changed virtual-style demo (hope it's OK)

image

maxkfranz:
Sorry for the delayed response, but I haven't had much time to devote to Cytoscape.js and I wanted to research and address this properly:

(8) `*-color-style` : I think the name would be clearer as `*-fill*`, e.g. `background-fill: linear-gradient | radial-gradient | color`.  Using `none` is confusing to me, as I think it means "draw nothing".

(9) `*-gradient-direction` : Fine, but please use dash-separated enum values.  If you use `to bottom` for an enum, then `to  bottom` (double space) doesn't work as it would in real CSS.  Dash-case is more consistent with other properties.  It also avoids making the property parsing more complicated.  Whitespace can be used to separate multiple property values, for instance.

(10) `*-gradient-stop-n-color` : I don't think it makes sense to separate the values this way.  There is a `multiple` flag for property types that allows a property to have multiple values in it.  That could be used for a single `*-gradient-stops` (e.g. `background-gradient-stops: cyan magenta yellow`).  

The pie properties existed when there wasn't multiple support and when there was no `function()` property value support in the stylesheets.  But then again, each pie slice has two values to it (colour and size), so a single `multiple: true` property probably isn't a good fit for pies.  This is why you have properties 1-16 for pies.

For just a gradient colour list, I think `multiple` is OK.  That way we support a dynamic number of stops.  Animations should still work, and users can specify `function()` mappers if need be.  I've added a `colors` type that uses `multiple:true` in afaa979637920a7af7c6514c7ce5f91fe29febf5.

It's also important to note that the Cytoscape desktop software supports an arbitrary number of stops, so that should be a requirement here for compatibility.  Further, Cytoscape desktop allows for specifying colour stop positions as percents.  That means we also need another property for stop positions (e.g. `background-gradient-stop-positions: 0% 50% 100%`), where the indices match up with those of the stops.

(11) Currently, `multiple: true` properties only support simple space separation when specified as a string, though JS arrays can be specified directly.  This is something we'd have to consider in the docs.  For instance, `rgb(1, 2, 3) rgb(4, 5, 6)` would not work but `rgb(1,2,3) rgb(4,5,6)` would -- where a string is used in place of an array.

(12) Thinking about it more, I think it would be best to separate the `fillStyle()` and `strokeStyle()` functions into separate cases, rather than trying to handle them all in one.  I'd also prefer to avoid creating intermediary objects for arguments for rendering code that we want lean.

I'm thinking this:

- `eleFillStyle( cxt, ele )` : Set fill style automatically based on element style.
- `colorFillStyle( cxt, r, g, b, o )` : Rename the old `fillStyle()` to make it clear it's just for colours.
- `gradientFillStyle( cxt, stops )` : `stops` is an array of `[r,g,b]` tuples -- `stops` could be received from `ele.pstyle('background-gradient-stops').value`.

So all current `fillStyle()` on unstable would be converted to `colorFillStyle()`.  The next step would be that the draw node code would call `eleFillStyle()`, which would call `colorFillStyle()` or `gradientFillStyle()` as appropriate.  This way, the code is simple for the common element case, but we still have the ability to set manual colours if the renderer needs to do so (e.g. overlays, panning indicator).

And then same applied to `strokeStyle()`.

What do you think?

elisherer:
Went with all of your comments,

You can see example usage at the debug page with:
```
edge {
line-fill: linear-gradient;
line-gradient-stops-colors: red blue green;
}
node {
background-fill: linear-gradient;
background-gradient-direction: to-left-bottom;
background-gradient-stops-colors: rgb(10,50,100) #77ffaa;
background-gradient-stops-positions: 30 70;
}
```

maxkfranz:
@elisherer This looks great!  Sorry for the delayed response.  I'll try to set aside some time for final testing and merging of PRs tomorrow

maxkfranz:
The code is merged with some minor cleanup.

TODO: Fill a ticket for the milestone: TODO-insert-link

elisherer:
@evgarik, Are you using the unstable release?

evgarik:
yes

maxkfranz:
There is no unstable release.  You have to build yourself if you want to use the unstable branch (at your own risk -- it's labelled "unstable" for a reason).

evgarik:
in file properties.js not enough lines
```
let edgeLine = [
    ....
    { name: 'line-gradient-stop-colors', type: t.colors },
    { name: 'line-gradient-stop-positions', type: t.percentages },
    ....
]
```

but when building these lines are not added to the build :(
if manually added to the cytoscope.js then works!

PR #2069, opened 2018-02-25
Fix lint

TrySound:
Most of errors are missing semicolons and unused vars

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,6 +2,7 @@
     "env": {
         "commonjs": true,
         "shared-node-browser": true,
+        "browser": true,
         "node": true,
         "mocha": true
     },
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "watch-babel": "gulp watch-babel",
     "test": "gulp test",
     "test-build": "gulp test-build",
-    "test-travis": "gulp build-cjs && gulp test-build",
+    "test-travis": "gulp build-cjs && gulp test-build && gulp lint",
     "sniper": "sniper .",
     "docs": "gulp docs",
     "docs-min": "gulp docs-min",
--- a/src/collection/algorithms/kerger-stein.js
+++ b/src/collection/algorithms/kerger-stein.js
@@ -4,10 +4,8 @@ var elesfn = ({
 
   // Computes the minimum cut of an undirected graph
   // Returns the correct answer with high probability
-  kargerStein: function( options ){
+  kargerStein: function(){
     var eles = this;
-
-    options = options || {};
 
     // Function which colapses 2 (meta) nodes into one
     // Updates the remaining edge lists
--- a/src/extensions/layout/breadthfirst.js
+++ b/src/extensions/layout/breadthfirst.js
@@ -16,7 +16,7 @@ let defaults = {
   animate: false, // whether to transition the node positions
   animationDuration: 500, // duration of animation in ms if enabled
   animationEasing: undefined, // easing of animation if enabled,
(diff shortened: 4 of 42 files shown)


.eslintrc.json
```diff
@@ -2,6 +2,7 @@
     "env": {
         "commonjs": true,
         "shared-node-browser": true,
+        "browser": true,
```
maxkfranz: I don't want this on, as we also support Node.  Each use of a browser global should have to be manually toggled per file with eslint comments.

package.json
```diff
@@ -78,7 +78,7 @@
     "watch-babel": "gulp watch-babel",
     "test": "gulp test",
     "test-build": "gulp test-build",
-    "test-travis": "gulp build-cjs && gulp test-build",
+    "test-travis": "gulp build-cjs && gulp test-build && gulp lint",
```
maxkfranz: Fair enough

src/collection/algorithms/kerger-stein.js
```diff
@@ -4,11 +4,9 @@ var elesfn = ({
 
   // Computes the minimum cut of an undirected graph
   // Returns the correct answer with high probability
-  kargerStein: function( options ){
+  kargerStein: function(){
     var eles = this;
 
-    options = options || {};
-
```
maxkfranz: I don't care about unused arguments.  They're often used here to make function signatures clear.  Because we're not using something like TypeScript, this is probably the clearest we can make it.  So I'd rather turn off [the args option in the eslint config](https://eslint.org/docs/rules/no-unused-vars#args) rather than changing the code for each of these.

The option for eslint errors when a `var`/`let`/`const` variable in a block isn't used is good and should remain.

src/extensions/layout/breadthfirst.js
```diff
@@ -16,7 +16,7 @@ let defaults = {
   animate: false, // whether to transition the node positions
   animationDuration: 500, // duration of animation in ms if enabled
   animationEasing: undefined, // easing of animation if enabled,
-  animateFilter: function ( node, i ){ return true; }, // a function that determines whether the node should be animated.  All nodes animated by default on animate enabled.  Non-animated nodes are positioned immediately when the layout starts
+  animateFilter: function (){ return true; }, // a function that determines whether the node should be animated.  All nodes animated by default on animate enabled.  Non-animated nodes are positioned immediately when the layout starts
```
maxkfranz: This gets used to generate docs, so removing the signature is problematic.

src/extensions/layout/breadthfirst.js
```diff
@@ -67,7 +67,7 @@ BreadthFirstLayout.prototype.run = function(){
 
         eles.bfs( {
           roots: unhandledNodes[0],
-          visit: function( node, edge, pNode, i, depth ){
+          visit: function( node ){
```
maxkfranz: This is an example of where it would be OK to remove the args, but I'd rather just remove the eslint rule instead of evaluating which should be removed and which ones not.  I'd rather address those when those particular parts of the code are being worked on and refactored.

src/style/string-sheet.js
```diff
@@ -32,6 +32,7 @@ styfn.appendFromString = function( string ){
     }
   }
 
+  // eslint-disable-next-line
```
maxkfranz: I generally don't like ignoring an entire line for all rules.  I think it's better to selectively disable the rule of interest in case there's another actual problem in the line in future.

maxkfranz:
All changes that aren't bug fixes should go in the unstable branch.  I've addressed the linting with less changes there.  Let me know what you think.

PR #2072, opened 2018-03-06
Add support for chevron shape

wrnrlr:


maxkfranz:
I've fixed some issues with the shape in #2082

PR #2084, opened 2018-03-26
Basic convert to esm with only default exports

TrySound:
Ref #2012 

I decided to make at first basic conversion with only default exports. Also I added esm loader.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,5 @@
 {
-  "presets": ["env"]
+  "presets": [
+    ["env", { "modules": false }]
+  ]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,6 +6,7 @@
         "mocha": true
     },
     "parserOptions": {
+      "sourceType": "module",
       "ecmaVersion": 6
     },
     "extends": "eslint:recommended",
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: node_js
 node_js:
-  - "4"
-  - "5"
   - "6"
   - "7"
-  - "stable"
+  - "node"
 sudo: false
 script: npm run test:travis
--- a/package-lock.json
+++ b/package-lock.json
@@ -2336,13 +2336,15 @@
         "is-valid-instance": "0.1.0",
         "kind-of": "3.2.2",
         "lazy-cache": "2.0.2",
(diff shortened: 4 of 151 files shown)


TrySound:
before

cytoscape.js 815.1K
cytoscape.min.js 325.2K
cytoscape.umd.js 795.9K

after

cytoscape.js 888.0K
cytoscape.min.js 327.0K
cytoscape.umd.js 868.8K

seems like currently size increased with 2kb, but it should be fixed with rollup's scope hoisting

TrySound:
Benchmarks has a lot of lint errors so I'd like to convert them later.

TrySound:
@maxkfranz Thoughts?

maxkfranz:
It looks OK from what I saw before.  I just haven't had time to test and merge...

I'll try to set aside some time tomorrow to try it out (and possibly merge).

maxkfranz:
Some questions re. next steps:

(1) Should we be specifying 'module' in package.json? http://2ality.com/2017/07/npm-packages-via-babel.html

(2) Does Rollup support everything we're currently doing with Webpack in the webpack.config.js and the package.json?

maxkfranz:
Re (2), we need:

(a) Watch support : --watch via CLI: https://rollupjs.org/guide/en#command-line-flags

(b) Inline source maps toggled on/off via config file: https://rollupjs.org/guide/en#output-sourcemap-m-sourcemap

(c) A way to minify with Uglify and toggle it on/off with an env var in the config file: Not sure if this is possible. We could just pipe the output to uglify, but then Windows users can no longer build.  For myself, that's not a problem but it might discourage Windows users from contributing to the lib.

(d) Babel support: https://rollupjs.org/guide/en#babel

(e) Making a UMD build that supports external refs to dependencies.  Not sure if this is supported.

(f) Bundling dependencies from npm that aren't ES modules:  It looks like a plugin does this https://rollupjs.org/guide/en#npm-packages

(g) Something analogous to EnvironmentPlugin from Webpack : Not sure if this exists

TrySound:
1. Yes. `module` field will make sense when we will land esm bundle.
2. Yes. Rollup support everything we need.
a. rollup-watch
b. yes
c. There is `cross-env` package which works for all shells.
d. rollup-plugin-babel
e. yes
f. rollup-plugin-node-resolve and rollup-plugin-commonjs
g. rollup-plugin-replace

PR #2098, opened 2018-04-30
Bundle with rollup

TrySound:
Ref https://github.com/cytoscape/cytoscape.js/pull/2084

In this diff I replace webpack with rollup for bundling source code.

I decided to use multiconfig instead of config with many options to
improve build time by eliminating node startup two more times. Also this
will simplify configuration in the future updates.

Changes

"umd" bundle for browser environment loads `lodash.debounce` from `global._.debounce`.

`snapshot` string is added by rollup to eliminate runtime condition.

`watch:fast` script does not generate source maps because they are not
fast enough.

Here's size comparision. I have some ideas how to reduce even
more.

cytoscape.js 893.0K
cytoscape.min.js 330.0K
cytoscape.umd.js 873.9K

cytoscape.js 834.9K
cytoscape.min.js 319.0K
cytoscape.umd.js 812.2K


maxkfranz:
This looks great!

I made a small patch: 
[rollup-esm-etc.patch.txt](https://github.com/cytoscape/cytoscape.js/files/1968632/rollup-esm-etc.patch.txt)


```
git apply rollup-esm-etc.patch
```

Tweaks

- Enabled sourcemaps for watch:fast -- needed for /debug page
- Reduce repetition in rollup conf.
- Misc. linting fixes
- Added esm target and 'module' key in package.json.  Might have made some mistakes here...
- Builds for dist now include sourcemaps -- except for .min.js file
- Add build:* targets back -- not used for main build target -- just for easy 1-file building

maxkfranz:
Let me know what you think and if we need any other changes.  I'm not sure if the esm build is working perfectly, for example, I just inspected the output in vim.

PR #2107, opened 2018-05-10
add required function to falseThread that is called if a layout.run()…

sjneph:
… is called more than once in non-weaver set-up.

### DIFF
--- a/src/extensions/layout/cose.js
+++ b/src/extensions/layout/cose.js
@@ -157,7 +157,10 @@ CoseLayout.prototype.run = function(){
     },
     stop: function(){
       return this;
-    }
+    },
+    stopped: function(){
+      return true;
+    },
   };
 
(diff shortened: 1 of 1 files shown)


sjneph:
For the cose layout.

sjneph:
This is a minimal example that throws an error:
```
function loadme() {
  const layout_name = 'cose';
  const layout_opts = {
    name: layout_name,
    nodeDimensionsIncludeLabels: true,
  };

  const options = {
    container: $('#cy'),
//    elements: GlyElements,
//    style: GraphElementStyles.style,
//    layout: layout_opts,
  };

  cy = cytoscape(options);
  const layout = cy.layout(layout_opts);
  layout.run();
  layout.run();
}
```

`TypeError: thread.stopped is not a function`

PR #2113, opened 2018-05-14
Remove useless minified bundle size snapshot

TrySound:
Since both `umd` and `min` bundles are identical there is not reason to
make a snapshot of them both.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -4,11 +4,6 @@
     "minified": 317897,
     "gzipped": 98067
   },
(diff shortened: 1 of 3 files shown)


maxkfranz:
The build is OK, so I merged the change.  But is there any reason in particular for the package-lock changes?

TrySound:
No reason. I installed today npm6. Probably it's breaking change. However I got this and with npm5 too. With yarn my lockfile never changes without the reason.

PR #2118, opened 2018-05-19
Fail on CI if size snapshot is not updated

TrySound:
Sometimes it's easy to forget to run build and commit snapshot. This
leads to unobjective differences in snapshots in further commits.

`matchSnapshot` option allows to fail if existing commit and generating
one are not matched. This is useful to fail on CI.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,21 +1,21 @@
 {
   "build/cytoscape.umd.js": {
-    "bundled": 854989,
-    "minified": 317897,
-    "gzipped": 98067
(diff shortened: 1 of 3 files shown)


maxkfranz:
Fair enough, but I suspect we may get false failures if the dev isn't using the same version of npm.

TrySound:
I added threshold in size snapshot plugin  option for such cases.

maxkfranz:
Good to know.  I'll merge this in then and we can set a threshold later if need be

PR #2123, opened 2018-05-30
Bellman-Ford returning negative weight cycles

TexKiller:
Additional option for the bellmanFord method:

- **findNegativeWeightCycles**: If this option is set to true, the *bellmanFord* method won't stop when it finds a negative weight cycle. Instead, it will enumerate all the unique negative weight cycles it can find and return them in the array **negativeWeightCycles**.


maxkfranz:
@TexKiller Do you see a way for this to fit in with the changes made to the algorithms regarding refactoring and performance?  It would be nice to have this feature, especially if it could be integrated in a simple way.

TexKiller:
Sorry about the delay, I'll get on that. Thanks!

Issue #1279, opened 2016-03-02
How to rotate layout on 90 degree?


gig177:
How to rotate layout of network entirely? I expect something like that:
```javascript
var cy = cytoscape({
    ...
    layout: {
        name: 'breadthfirst',
        directed: true,
        flow: 'LR', // hypothetical option
    }
});
```
I need to change network flow from `top->bottom` to `left->right`.
![a](https://cloud.githubusercontent.com/assets/8968462/13459957/715277a6-e088-11e5-9d3d-c337a78f7efe.png)

appunnicer13:
You can use CSS on the shadow elements , so rotate the whole cy by 90 deg and then rotate each element by -90deg . I think that will solve the problem , no need of options

gig177:
Thanks for your reply! This snippet doesn't solve this problem:
```css
canvas {
    transform:rotate(-90deg);
}
```
It isn't work perfectly cause axis [is break now (demo)](http://pagerank-gig177.rhcloud.com/?cytoscape-issue=1279).

maxkfranz:
Use [dagre](https://github.com/cytoscape/cytoscape.js-dagre) with the `rankDir` option

gig177:
Dagre calculates edge's length in unexpected way. For dynamic CRUD system it's unsuitable.
Therefore `breadthfirst` layout is ideally for me.

maxkfranz:
Then you could rotate the node positions after breadthfirst finishes.  It's just [a little bit of trig](https://en.wikipedia.org/wiki/Rotation_%28mathematics%29#Two_dimensions).  If you want, you could put that into breadthfirst directly as an option and push a PR.

chirchiAbdelhamid:
You can make a rotate function like that
function rotate() {
			setTimeout(function (){
				cy.nodes().positions(function( i, node ){
					  return {
						    x: node.position('y')*6,
						    y: -1*node.position('x')*0.2
						  };
						});
				 
				cy.fit()
	
				},100); 
}

PR #2147, opened 2018-06-29
Allow to rotate `BreadthFirst` graph by any angle

josejulio:
I'm not sure if point to `master` or `unstable` as this is a backwards compatible change, let me know.

rotate

Related to #1279

### DIFF
--- a/src/extensions/layout/breadthfirst.js
+++ b/src/extensions/layout/breadthfirst.js
@@ -4,6 +4,7 @@ let is = require( '../../is' );
 
 let defaults = {
   fit: true, // whether to fit the viewport to the graph
+  rotate: 0, // whether to rotate the final graph by any specific angle (degrees)
   directed: false, // whether the tree is directed downwards (or edges can point in any direction if false)
   padding: 30, // padding on fit
   circle: false, // put depths in concentric circles if true, put depths top down if false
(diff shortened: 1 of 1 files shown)


josejulio:
I'm wondering if an extension over `core` or `collection` would be better approach.
`cy.rotate(-90)` or `cy.nodes().rotate(-90)`
I don't think rotating a few nodes would be useful for anyone, so maybe for the `core`?

maxkfranz:
It might be better as an option to `nodes.layoutPositions()`: http://js.cytoscape.org/#nodes.layoutPositions

That function is called by each discrete layout to set the positions.  It already includes an arbitrary `transform()` function to transform the positions.  Maybe there could be a `rotate: angle` option that just sets the `transform()` function appropriately.

maxkfranz:
The advantage in that approach is that you could easily apply `rotate: angle` to any discrete layout.

maxkfranz:
> I'm not sure if point to master or unstable as this is a backwards compatible change, let me know.

New things should always go in unstable.  Only bug fixes go to master.

maxkfranz:
It also may make sense to have an IdentityLayout that just sets the existing positions.  That could be used, for example, if you want to apply only transforms.

josejulio:
I'm going to close this as there is already a transform() function.

Maybe adding a cy.makeTransformWithRotation(angle) could be created, and the layouts accepting this function could use.

maxkfranz:
I'd rather have the feature as an option to `layoutPositions()` as that would add the feature to all discrete layouts -- even third party extension layouts -- automatically.

josejulio:
Ok @maxkfranz I'll try to update the PR with the requested changes

cappslock:
Hi, does this feature exist somewhere currently?

PR #2150, opened 2018-07-06
Feature/ignore hidden nodes during layout process

bgross37:
Hi all,
_this is my first time contributing to anything, so tell me if I am doing anything wrong.
I tried to follow your guide http://blog.js.cytoscape.org/2017/06/13/contributing/ as close as possible._

This change adds the option to only layout the visible elements in a graph.
For my use, I have to make nodes invisible in certain situation, and re-layout the graph, therefore I have holes in the circle, which look wierd.

I simply added an option to the code, and if it is set to true, it will filter out the hidden nodes before the rest of the layout process calculates the positions.

To maintain compatibility, I set this option to false by default.


maxkfranz:
Thanks for the PR, but you should be able to run layouts on arbitrary subsets of the graph using the existing API, e.g.

```js
cy.elements(':visible').layout({ /* ... */ }).run();
```

bgross37:
Aw man, how come I did not think of this?!
Sorry to bother you, I should have realized this on my own.

PR #2188, opened 2018-09-12
Master

Silverhumor:
l want to use this components


maxkfranz:
Unstable is merged into master only for official releases

Issue #2231, opened 2018-12-06
When using `cy.json`, sometimes some nodes are not being deleted


josejulio:
**Issue type**

Bug report

**Environment info**

- Cytoscape.js version 3.2.22 (i suppose it might happen on 3.3.0 too):
- Browser/Node.js & version : Chrome Version 70.0.3538.67 (Official Build) (64-bit)

**Current (buggy) behaviour**

When using `cy.json`, sometimes some nodes are not being deleted

**Desired behaviour**

When using `cy.json`, all nodes not in the provided json should be deleted.

josejulio:
I think the bug was introduced here:

https://github.com/cytoscape/cytoscape.js/commit/47a354eba7d5143ceb3778446921bdd3ae3924d1#diff-4cbda7cf5642b53f0b4ada2eb9e65229R339


The for goes from 0 to eles.length, but the content of the loop is removing elements, thus updating eles.length.

I'm working on the bug, but we should either go from eles.length to 0, or filter all the nodes that are going to be deleted and iterate over that array.

maxkfranz:
Removing from the graph doesn't affect the contents of a collection.  Do you have a reproducible test case?

josejulio:
Is reproducible in the context of the application, but couldn't reproduce in a jsbin.

Let me try again.

josejulio:
Also, not sure if this helps, but the collection used is the `mutableElements` defined in `core/search.js` as:
```js
mutableElements: function(){
    return this._private.elements;
  }
```

josejulio:
got this reproducer: https://jsbin.com/qulepikabe

Click the cat and it will remove about the half, when it should remove all but one.

josejulio:
Smaller one, which might be better for unit testing. I'm going to add to my PR:
https://jsbin.com/muyebuhana

josejulio:
Is `3.3.1` scheduled for any day of this week? I'm wondering to know how to deal with an issue, thank you!

maxkfranz:
I've been away and unable to release.  I'll publish tomorrow morning.

josejulio:
Thank you

PR #2232, opened 2018-12-06
Fixes a bug introduced by removing elements in a for 0 -> eles.length

josejulio:
As this is a bugfix, i'm setting it to master. 

Please let me know if should change that.

Ref #2231

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -370,20 +370,17 @@ util.extend( corefn, {
         }
 
         // elements not specified in json should be removed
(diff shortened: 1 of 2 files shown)


josejulio:
Added test case, without the fix it fails:

```
  1 failing

  1) Core graph manipulation
       cy.json()
         cy.json() can remove all elements:

      AssertionError: expected 1 to equal 0
      + expected - actual

      -1
      +0

```

PR #2239, opened 2018-12-12
Optimizations for mapped style properties

ntfink:
This PR includes two proposed performance optimizations to styling, specifically for mapped properties.

Optimization 1:
* Mapped properties for contexts an element doesn't have still get flagged as ```diffProps``` and are processed as such only to get "deleted" (those properties won't get set in ```getContextStyle```) from the style later . The optimization catches this case earlier in the styling process.
* Specific changes: In ```getPropertiesDiff```, only consider a context if 1) the context has changed, or 2) if the element has the context and the context has mapped properties (case 2 contains the new behavior).

Optimization 2:
* Mapped properties get recomputed and reapplied every time regardless of whether they changed or not. The optimization caches the previous, un-parsed value (could be changed to the full ```argHash``` if necessary) for a property and compares it to the return value of the mapped property on the next iteration to determine if reapplying can be skipped.
* Specific changes: Cache the previous, un-parsed value in ```parse``` and skip properties in ```applyContextStyle``` if the newly computed return value for a mapped property matches the cached value (also temporarily cache the return value in case they don't match to avoid computing it a second time during parsing).

### DIFF
--- a/src/style/apply.js
+++ b/src/style/apply.js
@@ -578,7 +578,7 @@ styfn.applyParsedProperty = function( ele, parsedProp ){
 
   case types.fn: {
     let fn = prop.value;
-    let fnRetVal = fn( ele );
+    let fnRetVal = prop.fnValue || fn( ele ); // check for cached value before calling function
 
     if( fnRetVal == null ){
       util.warn('Custom function mappers may not return null (i.e. `' + prop.name + '` for ele `' + ele.id() + '` is null)');
--- a/src/style/apply.js
+++ b/src/style/apply.js
@@ -74,7 +74,7 @@ styfn.getPropertiesDiff = function( oldCxtKey, newCxtKey ){
     let cxtHasDiffed = oldHasCxt !== newHasCxt;
     let cxtHasMappedProps = cxt.mappedProperties.length > 0;
 
-    if( cxtHasDiffed || cxtHasMappedProps ){
+    if( cxtHasDiffed || ( newHasCxt && cxtHasMappedProps )){
       let props;
 
       if( cxtHasDiffed && cxtHasMappedProps ){
@@ -205,6 +205,13 @@ styfn.applyContextStyle = function( cxtMeta, cxtStyle, ele ){
 
     // save cycles when the context prop doesn't need to be applied
     if( eleProp === cxtProp ){ continue; }
+
+    // save cycles when a mapped context prop doesn't need to be applied
+    if( is.fn( cxtProp.value )) {
+      cxtProp.fnValue = cxtProp.value( ele ); // temporarily cache the value in case of a miss
+
+      if( cxtProp.fnValue === eleProp.previousValue ){ continue; }
+    }
 
     let retDiffProp = retDiffProps[ diffPropName ] = {
       prev: eleProp
--- a/src/style/parse.js
+++ b/src/style/parse.js
@@ -35,6 +35,11 @@ styfn.parse = function( name, value, propIsBypass, propIsFlat ){
     }
   }
 
+  // cache the original, unparsed value for comparison on the next iteration
+  if( ret ){
+    ret.previousValue = util.copy( value );
+  }
+
   return ret;
 };
 


src/style/apply.js
```diff
@@ -206,6 +206,13 @@ styfn.applyContextStyle = function( cxtMeta, cxtStyle, ele ){
     // save cycles when the context prop doesn't need to be applied
     if( eleProp === cxtProp ){ continue; }
 
+    // save cycles when a mapped context prop doesn't need to be applied
+    if( is.fn( cxtProp.value )) {
```
maxkfranz: Generally if `cxtProp.mapped != null`

The function case is probably cheaper as [`cxtProp.mapped === types.fn`](https://github.com/cytoscape/cytoscape.js/blob/unstable/src/style/parse.js#L99)

src/style/parse.js
```diff
@@ -35,6 +35,11 @@ styfn.parse = function( name, value, propIsBypass, propIsFlat ){
     }
   }
 
+  // cache the original, unparsed value for comparison on the next iteration
+  if( ret ){
+    ret.previousValue = util.copy( value );
```
maxkfranz: Is the copying strictly necessary?  It adds expense.

It's certainly not necessary in the non-mapper cases.

src/style/apply.js
```diff
@@ -74,7 +74,7 @@ styfn.getPropertiesDiff = function( oldCxtKey, newCxtKey ){
     let cxtHasDiffed = oldHasCxt !== newHasCxt;
     let cxtHasMappedProps = cxt.mappedProperties.length > 0;
 
-    if( cxtHasDiffed || cxtHasMappedProps ){
+    if( cxtHasDiffed || ( newHasCxt && cxtHasMappedProps )){
```
maxkfranz: Makes sense

src/style/apply.js
```diff
@@ -206,6 +206,13 @@ styfn.applyContextStyle = function( cxtMeta, cxtStyle, ele ){
     // save cycles when the context prop doesn't need to be applied
     if( eleProp === cxtProp ){ continue; }
 
+    // save cycles when a mapped context prop doesn't need to be applied
+    if( is.fn( cxtProp.value )) {
+      cxtProp.fnValue = cxtProp.value( ele ); // temporarily cache the value in case of a miss
+
+      if( cxtProp.fnValue === eleProp.previousValue ){ continue; }
+    }
+
```
maxkfranz: If we use this approach, we must delete `ele.updateMappers()` and `style.updateMappers()`.

src/style/apply.js
ntfink: Makes sense!

src/style/parse.js
ntfink: I was making an assumption that the un-parsed value could be a type (ex: array) that would require copying. But if that's not the case, I can certainly remove it!

Alternatively and/or in addition to, another option would be to move this statement to [```applyParsedProperty```:](https://github.com/cytoscape/cytoscape.js/blob/6feca07b2c6f422a41993bcdc16fe3fc4dae3995/src/style/apply.js#L588)
```
case types.fn:
  ...
  flatProp = this.parse( prop.name, fnRetVal, prop.bypass, flatPropMapping );
  flatProp.previousValue = fnRetVal; // or util.copy(fnRetVal)
```

Then regardless of whether copying is required or not, only previous values for mapper cases get cached. Which would make sense as that's the only case where it gets used.

src/style/apply.js
ntfink: I'm afraid I'm not too familiar with these functions, what makes them incompatible with this approach?

ntfink:
And then I was on vacation and failed to respond in a reasonable time - sorry for the delay!

Thanks much for the in depth feedback! I added a couple in-line comments, but also wanted to address your more general question about (2) vs. memoize. I think they both contribute to improved performance but solve different problems. 

Memoize would definitely speed up the actual computation of the property value via its caching functionality. The performance gains from (2) are not from caching the previous value (which is just a means to compare it to the new value), but from from bypassing the ```applyParsedProperty``` function chain for mapped properties that haven't changed since the last style update which you can't achieve with memoize at the consumer level. For large graphs with complicated (i.e., lots of mapped properties), cutting out that one function chain for each element adds up.

That being said, I do recognize the added overhead of copying (if it ends up being necessary). But (2) would still provide performance gains, even for memoized style properties, since they're impacting different parts of the process.

Does that make any more sense than before?

src/style/parse.js
maxkfranz: If the goal is to avoid the expense of `applyParsedProperty()` for function mappers, could we use an early exit condition within `applyParsedProperty()`?  Then all the logic for this purpose would be contained in the same function.

Additionally, the top-level `parse()` function is just about caching the properties to avoid parsing expense.  It's the `parseImpl()` function that contains all of the real parsing logic.

It's also worth considering that line 39 would copy for all property types, whereas the code in your comment is better scoped:  Copying happens only for function property values in your comment.

maxkfranz:
That helps to clarify things.  As described in the code comment, if possible, I'd like to see all the logic within `applyParsedProperty()` with an early exit condition.  That would keep all the logic in one place, but maybe it's not practical.  I'll take a closer look on Monday.

maxkfranz:
@ntfink The changes only include the proposals that we discussed.  The overall structure is the same as your initial PR.

Thanks again for the PR, and you should see the change in the 3.4.0 release at the beginning of February.

PR #2255, opened 2019-01-02
cy.json should remove the contents of compounds nodes.

josejulio:
eles.move(...) creates a copy of the elements to be "moved" and
 deletes the originals, thus we need to check if we need to remove the
 new elements.

Targeting master as this is a bugfix

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -313,6 +313,7 @@ util.extend( corefn, {
     let cy = this;
     let _p = cy._private;
     let eles = cy.mutableElements();
+    let getFreshRef = ele => cy.getElementById(ele.id());
 
     if( is.plainObject( obj ) ){ // set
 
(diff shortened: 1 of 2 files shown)


src/core/index.js
```diff
@@ -369,18 +369,25 @@ util.extend( corefn, {
           }
         }
 
+        let filterElesToRemove = function (eles) {
+          return eles.stdFilter( function( ele ){
+              return !idInJson[ ele.id() ];
+          }).toArray();
```
maxkfranz: Better to use the collection than an array

src/core/index.js
```diff
@@ -369,18 +369,25 @@ util.extend( corefn, {
...
-
-            ele.remove(); // remove parent
-          } else {
+        for( let i = 0; i < elesToRemove.length; i++ ){
+            let ele = elesToRemove[i];
+
+            if( ele.isParent() ) {
+              // Elements are not "moved" they are copied, extend elesToRemoveList
+              let children = ele.children().move({ parent: null }); // so that children are not removed w/ parent
+              util.push(elesToRemove, filterElesToRemove(children));
```
maxkfranz: Better to use merge() on a collection than push

src/core/index.js
josejulio: i used array and merge, because I can't be sure of the order in a collection, as I'm in a for-loop I rely on the order so i ensure that the moved objects are in the end. I'm using it as a "queue", but not removing any element of it, just incrementing the index.

josejulio:
@maxkfranz Updated the code, please let me know what you think. Thanks!

josejulio:
Merged `compound-json-2255` into my branch

josejulio:
> Please review and add any test cases you can think of.

It look good to me, added a test to check if removing a middle parents works well.

PR #2259, opened 2019-01-09
Add module field to package.json to enable ES modules interoperability

hator:
This pull request adds `module` field to package.json. It enables automatic consumption of ES module by bundlers like Webpack.

This is in line with the intent stated in the documentation:

> cytoscape.esm.js : A non-minified ESM (import / export) build without any bundled dependencies. This is intended to be consumed automatically by Node.js or a bundler like Webpack via import cytoscape from 'cytoscape'.
~ [http://js.cytoscape.org/#getting-started/including-cytoscape.js]()

It was also suggested as a next step in #2084.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "node": ">=0.10"
   },
   "main": "dist/cytoscape.cjs.js",
+  "module": "dist/cytoscape.esm.js",
   "scripts": {
     "lint": "eslint src benchmark",
     "build": "rollup -c",


maxkfranz:
This is a breaking change.

This isn't going to happen until either v4 is released or Webpack supports `require()` properly for packages that specify `module` in `package.json`.

See https://github.com/cytoscape/cytoscape.js/pull/2103#issuecomment-438434926

maxkfranz:
You can `import cytoscape from 'dist/cytoscape.esm.js'` if you want to use ESM.  Or you can set up an alias in your webpack config.

maxkfranz:
Thanks for the PR though

hator:
Thanks for your feedback. I didn't realize there was a problem with Webpack. I am closing this for now, feel free to reopen it whenever the problem is resolved. Have a good day :)

Issue #2256, opened 2019-01-08
In v3.3.x label distance to node acts "weird" compared to 3.2.22 for rotated labels


josejulio:
**Issue type**
Bug report

**Environment info**

- Cytoscape.js version : 3.2.22 vs 3.3.1
- Browser/Node.js & version : Version 70.0.3538.67 (Official Build) (64-bit)

**Current (buggy) behaviour**

The label is farther away than it should be:

Using v3.2.22
![bug-label-distance-in-3 2 22](https://user-images.githubusercontent.com/3845764/50849236-00915200-133c-11e9-914b-86003c4392ea.gif)

Using v3.3.1 (same applies for 3.3.0)
![bug-label-distance-in-3 3 1](https://user-images.githubusercontent.com/3845764/50849239-025b1580-133c-11e9-8573-53db63dedd09.gif)

When zooming a lot, the label distance stars showing where it should be.

**Desired behaviour**

v3.3.1 should show the label using the same distance as v3.2.22

**Minimum steps to reproduce**

- Add a label, rotated 90deg and aligned to the right

https://jsbin.com/yuwilafaxu

josejulio:
@maxkfranz Let me know if I can help with this. Will try to find where the code to handle the labels is.

josejulio:
From my tests, It seems to be rendering from the cache [1] when showing the "wrong" behavior, and when zooming, is not [2].

[1] https://github.com/cytoscape/cytoscape.js/blob/7482fdf3c61463ad574b9bea4a85a8efe6acdcbe/src/extensions/renderer/canvas/drawing-elements.js#L43
[2] https://github.com/cytoscape/cytoscape.js/blob/7482fdf3c61463ad574b9bea4a85a8efe6acdcbe/src/extensions/renderer/canvas/drawing-label-text.js#L224

josejulio:
Ok, I'm going to look the code to see how the bounding box is being build, to be able to do the right operations when drawing from the cache.

maxkfranz:
I've got a fix but it needs to be tested more today before merging.  I'll push the branch within the hour so you can try it out if you like.

josejulio:
I'm also wondering if you tested with `text-margin-x` or `text-margin-y`, it looked like they were not taken into account too. 
I'll test with branch `ele-txr-cache-rotpt`
Thanks!

josejulio:
There is still a little, but noticeable difference:

[...]

PR #2261, opened 2019-01-09
Fix cache drawing of rotated labels

josejulio:
@maxkfranz I added code on top of `ele-txr-cache-rotpt` to handle the `text-margin-x` and `text-margin-y`, please review.

Related to #2256


maxkfranz:
@josejulio See fe93c0c3c95eba166db391d16e11ea4bf27bd3c9

I think that approach is a bit simpler and cheaper.

josejulio:
I agree, is better to reuse the values :) 

Closing this.

Issue #2258, opened 2019-01-09
Suppressing warning messages


MatthijsBon:
**Issue type**
Core Library & Cose Layout.

Bug report

**Environment info**

- Cytoscape.js: v3.3.1:
- Browser: Chrome 71.0.3578.98 (Official Build) (64-bit):

**Current (buggy) behaviour**

I think this has to do with both the 'cose' layout as well as the core library. The main concern is adding nodes dynamically (using `elements.forEach(function (element) { cy.add(element) })`) or adding directly to the graph (using `{ (..), "elements": elements, (..) }`). When adding directly to the graph, the core library has no problems with adding the nodes using the 'cose' layout, but when adding dynamically, a warning is shown in the console that nodes probably overlap and therefore edges are impossible to draw. 

I can imagine edges are impossible to draw when at first they overlap, but in the end, the layout determines the position of the nodes and then they don't overlap anymore. Even when using something like `cy.batch(function () {})` to add all nodes (dynamically) at once and then run the layout, the warnings are still shown. 

This does not happen for 'breadthfirst' or 'random' layouts and (as far as I have tested) only for 'bezier' and 'straight' edge styles.

**Desired behaviour**

Some way to suppress these warnings when adding nodes dynamically, while the layout is responsible for the position of nodes.

NOTE: I did find a workaround: a (random) position when adding nodes solves the warning.

**Minimum steps to reproduce**

https://jsbin.com/manukazata/1/edit?js,console,output

maxkfranz:
What is the motivation for suppressing the warnings?  Is there a performance impact?

MatthijsBon:
Well not exactly performance impact, but showing all these warnings in the console is a pain in the ***.

maxkfranz:
That makes sense, but disabling the warnings comes at a cost:  When a dev creates an edge and its not visible because of the positions of the connected nodes, it would be confusing to not have the warning.

Maybe the warnings could be collated somehow.

maxkfranz:
I guess the warning could be suppressed automatically if the connecting nodes both have the default (0, 0) position.

Collation could more generally be done within a time threshold, but I'd be wary of edgecases.

MatthijsBon:
Yes, that would be a great solution I think! I'm not that well into the codebase of Cytoscape to submit a Pull Request, but I could try with some pointers maybe?

maxkfranz:
[...]

PR #2281, opened 2019-01-31
Check for animation status before printing edge is impossible to draw

Spaxe:
This PR is a simple condition check to ignore warnings that may be out of date. For example, if nodes were too close but laid out nicer later, the warnings are not really that useful, and can be noisy.

Closes https://github.com/cytoscape/cytoscape.js/issues/2258


maxkfranz:
Closing, as we've discussed other solutions

PR #2304, opened 2019-02-26
Added `warnings` to initialisation options

### LINKED ISSUE #2258: Suppressing warning messages

Spaxe:
Closes #2258.

Invalidates and closes #2281.

### DIFF
--- a/documentation/md/core/init.md
+++ b/documentation/md/core/init.md
@@ -39,6 +39,7 @@ An instance of Cytoscape.js has a number of options that can be set on initialis
   // interaction options:
   <a href="#init-opts/minZoom">minZoom</a>: 1e-50,
   <a href="#init-opts/maxZoom">maxZoom</a>: 1e50,
+  <a href="#init-opts/warnings">warnings</a>: true,
   <a href="#init-opts/zoomingEnabled">zoomingEnabled</a>: true,
   <a href="#init-opts/userZoomingEnabled">userZoomingEnabled</a>: true,
   <a href="#init-opts/panningEnabled">panningEnabled</a>: true,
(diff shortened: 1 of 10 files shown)


Spaxe:
Please note that I only ran `npm test` and have not done further investigations to confirm that the option actually works. In theory, it should. Tests are welcome - I'm not as familiar with this code base.

maxkfranz:
This looks great, but now I wonder if we could make it a bit harder to make mistakes with future additions of `warn()` to the code.  Maybe a `cy.warn()` function would be useful.  That way, you can just call `cy.warn()` anywhere in the code without having to check the flag yourself each time.

Or maybe it would be even simpler to just make this a global option under the `cytoscape` object itself rather than per-instance.  E.g. `cytoscape.warnings(false)` would disable warnings on all `cy` instances.  I like this second option a bit better, because the `util.warn()` calls don't have to change at all.  It's also a bit more flexible, since the dev could turn the warnings off and on dynamically.

maxkfranz:
I noticed that some warnings aren't suppressed with this change (e.g. some selector warnings) --- probably because the selector object doesn't have a reference to `cy`.  It might be simpler to use the global approach.

Spaxe:
Correct. I didn't have a way to reference the `cy` object.

Not really a fan to pollute the global scope. Maybe someone else can write a PR and send in a global version. No harm done. :)

maxkfranz:
> Not really a fan to pollute the global scope

To clarify, it's global w.r.t. all `cy` instances -- it's not a global variable under `window`.

Spaxe:
OK, I see in your commit that's how you did it.

It might be easier to cherry pick your commit out of this branch and do another PR - since yours is branched after `util.warn` is called, not before.

maxkfranz:
Merged

Spaxe:
Simplest option is almost always the best. Thanks!

PR #2314, opened 2019-02-28
Fix #2313 - bug that causes an emit before emitter is created.

### LINKED ISSUE #2313: Emit is called before emitter is created on Element
Core Library - Element

Bug report

**Environment info**

- Cytoscape.js version : 
3.4.2
- Browser/Node.js & version : 
{ npm: '6.8.0',
  ares: '1.15.0',
  cldr: '34.0',
  http_parser: '2.8.0',
  icu: '63.1',
  llhttp: '1.0.1',
  modules: '67',
  napi: '3',
  nghttp2: '1.34.0',
  node: '11.6.0',
  openssl: '1.1.0j',
  tz: '2018e',
  unicode: '11.0',
  uv: '1.24.1',
  v8: '7.0.276.38-node.13',
  zlib: '1.2.11' }

**Current (buggy) behaviour**

An emit is called prior to the emitter being created on an Element.

This line causes an emit to occur.

https://github.com/cytoscape/cytoscape.js/blob/45899d447c5e0ce8562a89ea95d151fd183f06c0/src/collection/element.js#L106

**Desired behaviour**

The deprecated code should still allow for nodes to be added with the style option.

Utilizing the `cy.add({})` method I added a new node that had a style option defined.

https://jsbin.com/galavujilu/edit?js,console,output

bdparrish:
Simple fix to change the order of creation of the emitter.

### DIFF
--- a/src/collection/element.js
+++ b/src/collection/element.js
@@ -99,14 +99,14 @@ let Element = function( cy, params, restore ){
     _p.classes.add(cls);
   }
 
+  this.createEmitter();
+
   let bypass = params.style || params.css;
   if( bypass ){
     util.warn('Setting a `style` bypass at element creation is deprecated');
 
     cy.style().applyBypass( this, bypass );
   }
-
-  this.createEmitter();
 
   if( restore === undefined || restore ){
     this.restore();


maxkfranz:
@bdparrish I've created a branch, `bug/emitter-2313` for this issue.  I added some test cases.  I've also added a fix for the second test case, as your fix seems to be for the first one.

Thanks for the PR, and please review the changes when you get a chance.

bdparrish:
`bug/emitter-2313` looks good. simple but effective 😄

PR #2318, opened 2019-03-02
change style application to satisfy CSP style-src 'self'

einsweniger:
Hi!

### Problem

I'm a bit focused on security, so my webservers all run with a very strict Content-Security-Policy; currently it's set to:

> Content-Security-Policy: default-src 'none'; img-src 'self'; script-src 'self'; prefetch-src 'self'; worker-src 'self'; connect-src 'self'; style-src 'self'

With current unstable, Chrome will refuse some style applications with the message:

> Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'".

### Changes

I've changed `src/extensions/renderer/canvas/index.js` to apply the style without triggering the error.
Please check that the changes do not change behaviour, I'm not confident with JavaScript, yet.

### Need help with

There is another problem that I cannot figure out by myself:
https://github.com/cytoscape/cytoscape.js/blob/d9424465f7a90303fa122cdef9df476ed6151072/src/extensions/renderer/base/index.js#L39-L46

Line 45 will trigger the same error, but I don't know hot to fix it. Can you help?

### DIFF
--- a/src/extensions/renderer/canvas/index.js
+++ b/src/extensions/renderer/canvas/index.js
@@ -51,10 +51,11 @@ function CanvasRenderer( options ){
   };
 
   var tapHlOff = '-webkit-tap-highlight-color: rgba(0,0,0,0);';
-
+  var tapHlOffAttr = '-webkit-tap-highlight-color';
+  var tapHlOffStyle = 'rgba(0,0,0,0)';
   r.data.canvasContainer = document.createElement( 'div' ); // eslint-disable-line no-undef
   var containerStyle = r.data.canvasContainer.style;
-  r.data.canvasContainer.setAttribute( 'style', tapHlOff );
+  r.data.canvasContainer.style[tapHlOffAttr] = tapHlOffStyle;
   containerStyle.position = 'relative';
   containerStyle.zIndex = '0';
   containerStyle.overflow = 'hidden';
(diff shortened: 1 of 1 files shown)


einsweniger:
Hm. I "fixed" it by adding a hidden span with an id that satisfies the check for the style sheet. The check is sloppy enough to allow this.

https://github.com/cytoscape/cytoscape.js/blob/d9424465f7a90303fa122cdef9df476ed6151072/src/extensions/renderer/base/index.js#L31-L33

src/extensions/renderer/canvas/index.js
```diff
@@ -63,13 +64,28 @@ function CanvasRenderer( options ){
   container.appendChild( r.data.canvasContainer );
 
   if( (container.getAttribute('style') || '').indexOf(tapHlOff) < 0 ){
-    container.setAttribute( 'style', ( container.getAttribute( 'style' ) || '' ) + tapHlOff );
+    container.style[tapHlOffAttr] = tapHlOffStyle;
```
maxkfranz: The if check isn't necessary anymore if we're setting the properties in a keyed way

src/extensions/renderer/canvas/index.js
```diff
@@ -63,13 +64,28 @@ function CanvasRenderer( options ){
...
+  if(is.ms()) {
+    styleMap['-ms-touch-action'] = 'none';
+    styleMap['touch-action'] = 'none';
   }
 
   for( var i = 0; i < CRp.CANVAS_LAYERS; i++ ){
     var canvas = r.data.canvases[ i ] = document.createElement( 'canvas' );  // eslint-disable-line no-undef
     r.data.contexts[ i ] = canvas.getContext( '2d' );
-    canvas.setAttribute( 'style', '-webkit-user-select: none; -moz-user-select: -moz-none; user-select: none; -webkit-tap-highlight-color: rgba(0,0,0,0); outline-style: none;' + ( is.ms() ? ' -ms-touch-action: none; touch-action: none; ' : '' ) );
+    for(const [k,v] of Object.entries(styleMap)) {
```
maxkfranz: We can't use `Object.entries()`.  We support IE9.  You can use `Object.keys()` instead though.

I'd stay away from `for of`, in case babel doesn't convert it independently of the system supporting iterables.  It's probably better to use `Object.keys().forEach()`

src/extensions/renderer/canvas/index.js
einsweniger: check removed in b949970787f1b1ca522d23ff40f7093c1d608004

src/extensions/renderer/canvas/index.js
einsweniger: I'll check if this runs without warnings, that will take a moment.
Otherwise this is resolved in 4906fe4f5df7cf334a668def4e678138382fd1f3

einsweniger: looks good, deploys without styling errors

PR #2346, opened 2019-03-30
Opt out of loop edge with inside-to-node

assafweinberg:
It would be nice to be able to opt out of the looping behavior when a node points to its parent. This PR allows nodes to point to their parent's center with `'target-endpoint' : 'inside-to-node'`

Screenshot 2019-03-30 11 01 58

It might be cleaner to add a different style property to opt out of the loop behavior and just respect whichever target-endpoint value is set by the user. Any feedback or recommendations on that is welcome.

### DIFF
--- a/src/extensions/renderer/base/coord-ele-math/edge-control-points.js
+++ b/src/extensions/renderer/base/coord-ele-math/edge-control-points.js
@@ -808,7 +808,7 @@ BRp.findEdgeControlPoints = function( edges ){
       rs.tgtIntn = passedPairInfo.tgtIntn;
 
       if(
-        hasCompounds &&
+        edge.pstyle( 'target-endpoint' ).value !== 'inside-to-node' && hasCompounds &&
         ( src.isParent() || src.isChild() || tgt.isParent() || tgt.isChild() ) &&
         ( src.parents().anySame(tgt) || tgt.parents().anySame(src) || src.same(tgt) )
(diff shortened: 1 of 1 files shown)


maxkfranz:
@assafweinberg On the one hand, a general property is appealing.  On the other hand, it wouldn't work unless you also set `target-endpoint: inside-to-node` or `source-endpoint: inside-to-node`, depending on whether the ancestor is the target or source respectively.  Another valid combination is to set a completely manual endpoint for the ancestor.

Maybe something like `compound-loop: yes | no` could be the general property.  If you set `compound-looping: no` and an ancestor-descendant edge is detected, there should probably a warning in the console if the endpoints are not specified properly.  It should also be mentioned explicitly in the documentation for `compound-looping`.  Otherwise, it's too easy to make a mistake.

maxkfranz:
Removing from the 3.6.0 milestone, unless a more general implementation is updated in the PR before the end of the month.

assafweinberg:
Thanks for the feedback, Max. We ended up using a different pattern that didn't require this and our fork was deleted as part of the cleanup. I still think the functionality is a good idea, but closing this out for now.

maxkfranz:
OK, feel free to open an issue to spec out the feature if you like.  It could see it being a great feature to have for several usecases, if the feature is generalised a bit more.

PR #2375, opened 2019-04-26
Fixes collection link for cy.$()

josejulio:


maxkfranz:
The source markdown file needs to be changed, not the output html file.

josejulio:
Ok, will update this.

josejulio:
Updated, now i'm modifying the markdown and not the html file.

PR #2382, opened 2019-05-03
Document isDestroyed method

josejulio:


maxkfranz:
This function wasn't intended for public consumption.  It's not very consistent with the names of other functions -- they don't use `isSomething()`, only `something()`.  I think it would be OK to make this public but only if we rename the function to `destroyed()`.

josejulio:
> This function wasn't intended for public consumption. It's not very consistent with the names of other functions -- they don't use `isSomething()`, only `something()`. I think it would be OK to make this public but only if we rename the function to `destroyed()`.

Sure, I'll do that.

PR #2405, opened 2019-05-21
Zoom animation can cause zoom value to be outside of [minZoom, maxZoom]

stephencorwin:
## The problem
We are using cytoscape within a React application. We noticed that when 2 nodes were at least 15,000 units apart, a `fitSelected` followed by a `fit` would cause cytoscape to crash.
cytoscape-bug-1

## Investigation: Part 1
What is happening is that `new OffscreenCanvas(NaN, NaN)` is being invoked because `zoom` is temporarily a negative number which makes `lvl` equate `NaN`.
https://github.com/cytoscape/cytoscape.js/blob/85f0c8dfebc3ed9796651fef43be7ff871db88ce/src/extensions/renderer/canvas/layered-texture-cache.js#L118

The reason why `lvl` is a `NaN` is because `Math.log` and `Math.log2` result in `NaN` at negative values.
```js
Math.log2(1) // 0
Math.log2(0) // -Infinity
Math.log2(-1) // NaN
```

This explains why the docs specifically mention that `zoom` must be positive.
https://github.com/stephencorwin/cytoscape.js/blob/unstable/documentation/md/core/zoom.md
> The zoom level must be a positive number.

## Investigation: Part 2
The cytoscape animation `step()` leverages `ease()` which can temporarily dip `zoom` into the negative range based on the values of `startZoom` and `endZoom`.
https://github.com/cytoscape/cytoscape.js/blob/40a1a8f1e9e9e086ff8f4e43fb122194a2252b1f/src/core/animation/step.js#L113

However, we need zoom to be able to temporarily dip into the negative to give the *bounce* effect of easing.

## Resolution
Because it is unsafe to pass a negative zoom into `Math.log` and cytoscape already has a wrapper for `Math.log2`, it is reasonable to default to clamping the `n` argument to `0` and providing an optional `safe` argument in the event that `NaN` is an acceptable outcome.

**Note:** If we never see a reason where `NaN` is acceptable, we can remove the argument check and always clamp.

## Other fixes
Since I am adjusting the cytoscape wrapper in the `math.js` module, it made sense to create a new test file in `/test/modules/`. When I was attempting to do this, I realized that `npm test` was not discovering any tests within sub directories.

I added the `--recursive` argument to the npm scripts so that all sub directories in the `test` directory are now discovered. This revealed 1 test failure for `/test/modules/emitter.js`. The test was calling `removeAllListeners` which seems to have been deprecated and subsequently removed awhile back. I eliminated this test as it was no longer necessary.

### DIFF
--- a/src/core/animation/step.js
+++ b/src/core/animation/step.js
@@ -1,6 +1,7 @@
 import easings from './easings';
 import ease from './ease';
 import * as is from '../../is';
+import {bound} from '../../math';
 
 function step( self, ani, now, isCore ){
   let isEles = !isCore;
@@ -110,7 +111,7 @@ function step( self, ani, now, isCore ){
     let animatingZoom = endZoom != null && isCore;
     if( animatingZoom ){
       if( valid( startZoom, endZoom ) ){
-        _p.zoom = ease( startZoom, endZoom, percent, easing );
+        _p.zoom = bound( _p.minZoom, ease( startZoom, endZoom, percent, easing ), _p.maxZoom );
       }
 
       self.emit( 'zoom' );
--- a/test/core-graph-manipulation.js
+++ b/test/core-graph-manipulation.js
@@ -834,4 +834,19 @@ describe('Core graph manipulation', function(){
     });
   });
 
+  describe('cy.zoom()', function(){
+    describe('cy.zoom(number)', function() {
+      it('should not return a negative value', function() {
+        cy.zoom(-10);
+        expect(cy.zoom() >= 0).to.equal(true);
+      });
+    });
+
+    describe('cy.zoom({level: number})', function() {
+      it('should not return a negative value', function() {
+        cy.zoom({level: -10});
+        expect(cy.zoom() >= 0).to.equal(true);
+      });
+    });
+  });
 });


stephencorwin:
personally, I would opt for always having `log()` be safe, but I added the optional argument if for some reason `NaN` is an acceptable outcome.

maxkfranz:
Thanks for the PR!  Your isolation of the issue is good, but the solution is a bit heavy handed and does not solve the root cause of the issue.

I assume this is reproducible only if you use an easing that can assign values beyond the [S, E] range for start value S and end value E.  That would include easings like `spring()`.

> This explains why the docs specifically mention that zoom must be positive.

It generally doesn't make sense for zoom to be zero or negative.  Optically, what would a negative zoom or a zero zoom mean?  It doesn't make sense.  They're extraneous values.

> However, we need zoom to be able to temporarily dip into the negative to give the bounce effect of easing.

No, zoom should never be non-positive.  This is the real issue.  To solve the issue directly, animations on zoom should verify on each step that the newly assigned zoom value ranges on [minZoom, maxZoom].  If not, the zoom value should be clamped.

stephencorwin:
Your requested change has been made and things look to be working still 🎉 🎉 
cytoscape-bug-2

**Note:** I left in the test case for `zoom()` always returning a positive number since it seems like a good way to ensure that the default `minZoom` is always positive.

@maxkfranz, Let me know if there is anything else needed before this PR is merged. Thanks!

maxkfranz:
Aside: `npm test:build` can't generally run anything in test/modules.  The main test suite is meant to be runnable from pretty well any version of node, and that means no import support or esm.

I guess it's fine for `npm test`, because only a person would run that manually and he'd probably be using a modern version of node.

stephencorwin:
Do tests that never get exercised even to exist then?

The tests in `./test/modules/util.js` and `./test/modules/emitter.js` were not being exercised before which is why `./test/modules/emitter.js` needed to be updated -- it full out of sync with `./src/emitter.js`.

In my opinion, the `test:build` (travis) needs to catch any failures for any test that exists. Otherwise, those tests lose value.

stephencorwin:
@maxkfranz, If you would prefer table the inclusion of all tests, I can remove my second commit and open a new issue where it can be discussed further.

maxkfranz:
>  If you would prefer table the inclusion of all tests, I can remove my second commit and open a new issue where it can be discussed further.

Makes sense

maxkfranz:
There's a set of tests for animation that could contain a test case for animating zoom with an easing like spring.  It wouldn't ever generate the exception you originally had, because the instance would be headless.  However, it could poll the zoom level during the animation to make sure that it's a positive value every frame.

I'm not sure to what level, if at all, minZoom and maxZoom values are sanitised.  They'll have sensible defaults, but the consumer could specify garbage values.  That's probably another ticket though

stephencorwin:
Issue https://github.com/cytoscape/cytoscape.js/issues/2408 has been opened and the 2nd commit for this PR has been removed.

Issue #2404, opened 2019-05-21
Label hit tests assume the label is centred on the anchor point


jshaughn:
Using text-events: yes for my nodes does activate an area for label events but it seems to assume text-valign: center.  So, for my 'top' node labels the active area is below the actual label.  Is this a known issue?  I am using v3.6.1.

jshaughn:
Thanks!

PR #2409, opened 2019-05-22
Ensure we use the center of the label to compute the bounding box

### LINKED ISSUE #91: Provide options to make graph interaction faster
As tests on the iPad 3 have shown, performance can be very good if we leverage less expensive features.  For example, SVG transform makes pinch-to-zoom native speed even though the code to support this is custom-written algebra with SVG transforms.
- Ghost option for dragging a node
- Straight (i.e. bundled) edges instead of using bezier curves

josejulio:
- This applies to findNearestElements/checkLabel

~~Fixes #91~~
whoops, meant to link #2404

@maxkfranz I'm not sure if is the correct way to fix, but at least shows where the problematic code is.
I'm not sure if this computation is being done in other parts, but if you point me out I could try to make this more general.

I'm going to take a look and see if I can apply this when creating the label to ensure labelX/Y are always at the center.

### DIFF
--- a/src/extensions/renderer/base/coord-ele-math/coords.js
+++ b/src/extensions/renderer/base/coord-ele-math/coords.js
@@ -247,12 +247,36 @@ BRp.findNearestElements = function( x, y, interactiveElementsOnly, isTouch ){
     if( !eventsEnabled || !text ){ return; }
 
     var rstyle = _p.rstyle;
+
+    // vertical and horizontal align affect labelX/Y being center of the label or not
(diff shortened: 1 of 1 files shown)


josejulio:
Seems that just got fixed :)

maxkfranz:
Yes, that was the right place.  I made it a bit simpler so there was less duplicated logic, but it should be the same result.

josejulio:
Yes, I saw you simplified the code a lot.
Thanks!

PR #2429, opened 2019-06-07
Add unpkg and jsdelivr entry points

tmcw:
This PR adds [unpkg](https://unpkg.com/) & [jdelivr](https://www.jsdelivr.com/) fields, so AMD loaders like d3-require can load cytoscape.js in a web browser. The UMD builds already exist (yay), so this just configures them to be used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,8 @@
     "node": ">=0.10"
   },
   "main": "dist/cytoscape.cjs.js",
+  "unpkg": "dist/cytoscape.min.js",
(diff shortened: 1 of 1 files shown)


maxkfranz:
Fair enough.

Apparently, setting these fields should update what file unpkg and jsdelivr serve by default.  Currently, they're both serving the CJS file unless you specify a particular file path --- which makes sense for npm but not for the browser.  I can't find documentation for these fields on the sites for unpkg and jsdelivr, though they're mentioned in other lib's repos.

PR #2430, opened 2019-06-08
docs: fix typo

Jason-Cooke:


maxkfranz:
Apparently, [it can be either one](https://www.wordnik.com/words/artifact).  It's with "e" in German and French, but it looks like "i" is more popular in English.

So I suppose let's go with the more popular option...

Jason-Cooke:
@maxkfranz Thank you for merging

PR #2433, opened 2019-06-12
COSE offset component separation by left-top of component

mhermher:
There is a very slight bug in the cose layout that components that are already separate are again separated when a layout is re-run. If you perform a cose layout, then use that as the starting position to perform another layout, the components separate even more. If you keep re-doing the layout, the components keep on separating. This leads to bigger and bigger separations on re-layouts. This can be undone by offsetting the component separation by the x1 and y1 of the component.


maxkfranz:
Yes, it looks like a regression.

PR #2446, opened 2019-06-21
Link to Including Cytoscape.js from Packages

TomasHubelbauer:


maxkfranz:
Maybe this could go in the "Next steps" section of "Getting started":  http://js.cytoscape.org/#getting-started/next-steps

It would be best for someone to read the "Getting started" section in the documentation before going to a tutorial.  That gives the essential basics very quickly, and then the reader can decide if he wants to read a longer tutorial.

TomasHubelbauer:
I just realized the Getting Started section is third in the menu on the site 🙄 My goal with this PR was to make sure this information is hoisted up where people can find it quickly when learning Cytoscape, because I didn't think it was shown prominently enough, but now I see that the section I took for Getting Started (because it was the first one) was actually Introduction.

I feel like swapping the order of the two sections is too disruptive of a change for a drive by contribution like mine, but I am super confused as to why the Getting Started one doesn't come sooner in the menu. :-)

maxkfranz:
"Introduction" has all the information people are looking for when evaluating "Is this a library I want to use?"  This section can be skimmed if you already know you want to use the library.  Lots of people don't use the library yet, so it's nice to let them know what it's about.

If you become interested in using the library:  "Notation" has all the nomenclature and conventions you must learn before you can understand what's written in the following sections.

Once you understand the terminology:  "Getting started" shows you how to initialise an instance of `Cytoscape`.

TomasHubelbauer:
Okay. I tend to follow a different order when learning, but that might be just my exception. I'll leave the docs as they are then.

PR #2492, opened 2019-08-06
Check self-loop before compound-loop

josejulio:
When a self-loop was inside a compound, the findCompoundLoopPoints took precedence over the self-loop, thus `loop-direction` and `loop-sweep` values were ignored.

@maxkfranz is there any reason to have the compound check before the self-loop check?

loop-in-compound

### DIFF
--- a/src/extensions/renderer/base/coord-ele-math/edge-control-points.js
+++ b/src/extensions/renderer/base/coord-ele-math/edge-control-points.js
@@ -834,7 +834,7 @@ BRp.findEdgeControlPoints = function( edges ){
       if(
         hasCompounds &&
         ( src.isParent() || src.isChild() || tgt.isParent() || tgt.isChild() ) &&
-        ( src.parents().anySame(tgt) || tgt.parents().anySame(src) || src.same(tgt) )
+        ( src.parents().anySame(tgt) || tgt.parents().anySame(src) || (src.same(tgt) && src.isParent()) )
       ){
         this.findCompoundLoopPoints(edge, passedPairInfo, i, edgeIsUnbundled);
 


josejulio:
It looks like travis is failing because of a network problem
```Timeout waiting for network availability.```

**Comment by maxkfranz** [comment]
> It looks like travis is failing because of a network problem
> Timeout waiting for network availability.

OK, I restarted the job.

**Comment by maxkfranz** [comment]
I don't think the order should matter, but why would the compound case be true in the first place?

```js
( src.parents().anySame(tgt) || tgt.parents().anySame(src) || src.same(tgt) )
```

This part of the expression should be false for a loop.  The set of ancestors should not include the node of interest.

josejulio:
I think is because `src.same(tgt)` is true. 
From the docs [eles.same](http://js.cytoscape.org/#eles.same)
> Determine whether this collection contains exactly the same elements as another collection.

maxkfranz:
You're right.  That part was cut off in my editor so I didn't notice it.  I should have looked more carefully after I pasted the line.

I found this: https://github.com/cytoscape/cytoscape.js/commit/5b0104486c8379d1918c874e05fdab059f8c6840

maxkfranz:
Maybe `(src.same(tgt) && src.isParent())` would be more appropriate.

josejulio:
> Maybe `(src.same(tgt) && src.isParent())` would be more appropriate.

Sounds good, will update the code.

josejulio:
Updated the code with your suggested change.

josejulio:
Forgot to mention, note that this being a bug fix, i'm targeting `master`. Let me know if should change it to `unstable`

PR #2522, opened 2019-09-19
Fix label split so to avoid to render empty lines

alessandro308:
If you try to create a label with long unsplittable name in the first line, it tries to split it, push an empty line as first line and renders the whole string in the second line.

This PR adds the logic to avoid to render the first line empty in the label.


alessandro308:
Screenshot 2019-09-20 11 38 19
The problem appears with long label, in this case there is a string like "MAC address\n\nIP Address\nOther stuff" and as you can see, if the string is short enough it is render in the right way, otherwise a blank line is put as first one.

PR #2580, opened 2019-11-29
Add biconnected components algorithm

r-ba:
This PR adds an algorithm due to [Hopcroft and Tarjan](https://en.wikipedia.org/wiki/Biconnected_component#Linear_time_depth_first_search). In particular, it identifies the biconnected components of a graph, as well as any of their respective cut vertices should they exist.


documentation/md/collection/biconnected.md
```diff
@@ -0,0 +1,20 @@
+## Details
+
+Note that this function searches for biconnected components only within the subset of the graph in the calling collection.
+
+This function returns an object of the following form:
+
+```js
+{
+  cutVertices, /* Collection of nodes identified as cut vertices */
+  components /* An array, whose members are collections of edges corresponding to each biconnected component */
```
maxkfranz: If they're components, they should probably include the nodes connected to the edges.  In most cases, I expect the consumer would be interested in the nodes in each component rather than the edges.  Having a collection of both allows the consumer some simplicity and flexibility.

Maybe `kargerStein()` should also be updated to return `components` with both edges and nodes in each component.  That would make the functions more consistent.

src/collection/algorithms/biconnected.js
```diff
@@ -0,0 +1,85 @@
+let elesfn = ({
+  biconnected: function() {
```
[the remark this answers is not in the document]
maxkfranz: There could be an alias like `ht()`.

src/collection/algorithms/biconnected.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const biconnectedSearch = (root, currentNode, parent) => {
+      if (root == parent) edgeCount += 1;
+      nodes[currentNode] = {
+        id : id,
+        low : id++,
+        cutVertex : false
+      };
+      let sourceId, targetId, otherNodeId, edgeId;
+      let edges = eles.getElementById(currentNode).connectedEdges();
```
maxkfranz: In order to consider only elements in the calling collection, `connectedEdges()` must be filtered by the intersection with `eles`.

documentation/md/collection/biconnected.md
```diff
@@ -0,0 +1,20 @@
+## Details
+
+Note that this function searches for biconnected components only within the subset of the graph in the calling collection.
+
+This function returns an object of the following form:
+
+```js
+{
+  cutVertices, /* Collection of nodes identified as cut vertices */
```
maxkfranz: It should probably be `cut` to be consistent with `kargerStein()`.

src/collection/algorithms/biconnected.js
r-ba: Good catch.

documentation/md/collection/biconnected.md
r-ba: This seems reasonable to me. Supposing that `kargerStein()` was modified in this manner, if the consumer wanted `components` to be more faithful to the mathematical definition of a cut they could always filter either collection to contain only nodes.

If you'd like, I'd be happy to go ahead and make that modification either here or in a separate PR, whichever you think would be more appropriate.

src/collection/algorithms/biconnected.js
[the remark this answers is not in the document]
r-ba: My initial reservation for naming it hopcroftTarjan was the fact that other algorithms have been published by the pair (e.g. their [planarity test](https://dl.acm.org/citation.cfm?doid=321850.321852)) that we might want to add at a later date. Nevertheless I've gone ahead and renamed it to that for the sake of consistency based on the assumption that that issue could be resolved at a later date.

documentation/md/collection/biconnected.md
maxkfranz: Might as well just keep things simple and reuse this PR.  Thanks

src/collection/algorithms/biconnected.js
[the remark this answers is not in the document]
maxkfranz: In that case, neither `biconnectedComponents()` or `hopcroftTarjan()` would be great names in the longterm.  I suppose we could use something like `htbc()`

maxkfranz:
I'm planning on merging these PRs in early next week.  I'll do another review pass, and I'll add a patch if I notice any minor issues at that time.

src/collection/algorithms/biconnected.js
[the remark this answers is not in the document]
r-ba: I'm impartial to any particular name, but `htbc` seems like a fine choice to me. Hopefully you don't mind me leaving this up to your discretion.

r-ba:
Sounds great. Hopefully the update to `kargerStein()` aligns with your expectations, I must admit that editing it almost felt unsanctimonious. 

I ran `npm benchmark` against it a few times and wasn't able to observe any significant discrepancies in performance which is nice. By the way, should the `benchmark:download` script in `package.json` be updated to fetch `.../dist/cytoscape.cjs.js` instead of `.../dist/cytoscape.js`? In it's current state this is throwing `HTTPError: Response code 404 (Not Found)` (rightfully-- as far as I can tell).

maxkfranz:
I've made some minor changes re. the API and I've made the edge filtering a bit more efficient by avoiding the creation of an intermediary collection.  A build with both of your PRs merged in is attached to this comment.

Thanks for the PRs.

[cytoscape-3.13.0unstable-snapshot.zip](https://github.com/cytoscape/cytoscape.js/files/3969174/cytoscape-3.13.0unstable-snapshot.zip)

Issue #1365, opened 2016-05-10
Iterable support for collections


maxkfranz:
This will give `for of` loop support etc

Ref : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

PR #2583, opened 2019-11-30
Iterable support for collections

r-ba:
This PR implements an iterable behaviour for collections, as proposed in issue [#1365](https://github.com/cytoscape/cytoscape.js/issues/1365).

In terms of writing documentation for this some soft guidance would be greatly appreciated, as this falls a bit outside of the eles.method(...) convention.

### DIFF
--- a/src/collection/index.js
+++ b/src/collection/index.js
@@ -105,6 +105,20 @@ let Collection = function( cy, elements, options ){
   if( createdElements ){
     this.restore();
   }
+
+  // define an iterable behaviour, e.g. for...of, spread syntax
+  this[Symbol.iterator] = function() {
+    return {
+      next: function() {
+        this._values = this._valueIterator.next();
+        !this._values.done ? this._values.value = this._values.value.ele : null;
+        return this._values;
+      },
+      _valueIterator: map.values(),
+      _values: undefined
+    };
+  };
+
 };
 
 // Functions
--- a/src/collection/index.js
+++ b/src/collection/index.js
@@ -105,20 +105,6 @@ let Collection = function( cy, elements, options ){
   if( createdElements ){
     this.restore();
   }
-
-  // define an iterable behaviour, e.g. for...of, spread syntax
-  this[Symbol.iterator] = function() { // eslint-disable-line no-undef
-    return {
-      next: function() {
-        this._values = this._valueIterator.next();
-        !this._values.done ? this._values.value = this._values.value.ele : null;
-        return this._values;
-      },
-      _valueIterator: map.values(),
-      _values: undefined
-    };
-  };
-
 };
 
 // Functions
--- a/src/collection/index.js
+++ b/src/collection/index.js
@@ -107,7 +107,7 @@ let Collection = function( cy, elements, options ){
   }
 
   // define an iterable behaviour, e.g. for...of, spread syntax
-  this[Symbol.iterator] = function() {
+  this[Symbol.iterator] = function() { // eslint-disable-line no-undef
     return {
       next: function() {
         this._values = this._valueIterator.next();
--- a/src/collection/iteration.js
+++ b/src/collection/iteration.js
@@ -122,4 +122,29 @@ let elesfn = ({
 
 elesfn.each = elesfn.forEach;
 
+// implement an iterator only for those instances which support Symbol.iterator
+if (typeof Symbol != "undefined" && typeof Symbol.iterator != "undefined") { // eslint-disable-line no-undef
+  elesfn[Symbol.iterator] = function() { // eslint-disable-line no-undef
+    let entry = { value: undefined, done: false };
+    let i = 0;
+    let length = this.length;
+
+    return {
+      next: () => {
+        if ( i < length ) {
+          entry.value = this[i++];
+        } else {
+          entry.value = undefined;
+          entry.done = true;
+        }
+
+        return entry;
+      },
+      [Symbol.iterator]: function() { // eslint-disable-line no-undef
+        return this;
+      }
+    };
+  };
+}
+
 export default elesfn;
--- a/src/collection/iteration.js
+++ b/src/collection/iteration.js
@@ -122,29 +122,34 @@ let elesfn = ({
 
 elesfn.each = elesfn.forEach;
 
-// implement an iterator only for those instances which support Symbol.iterator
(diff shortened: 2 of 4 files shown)


r-ba:
Whoops, I didn't realize the linter could cause the build to fail. Had to fix that... seeing that red X made my hands sweat.

maxkfranz:
I'll look at these reviews in more detail tomorrow, but I notice these items for this PR:

(1) `Symbol.iterator` may not be supported in some browsers.  We should probably just have a guard against it.  If the browser doesn't define `Symbol`, then it's fine to not enable iterable support.

(2) Should this be defined on the prototype?

r-ba:
I agree that it is critical to address (1).

For (2) do you mean something like: 
```js
Collection.prototype[Symbol.iterator] = function() {
  ...
};
```
This was my initial approach but the issue here seems to be that `this` only gives reference to the Symbol such that `this._private` is undefined. Though it is of course possible that there's some workaround here I'm simply unaware of.

Another alternative I could think of would be something like:
```js
elesfn.enableIterable = function() {
  this[Symbol.iterator] = function() {
    ...
  };
};
```
This works. A drawback to this approach however is that it requires the user to call the function ahead of time per collection which seems suboptimal.

Perhaps there's an even better option which I'm not seeing. Looking forward to your feedback.

r-ba:
For (2) I presume you mean something along the lines of:
```js
elesfn[Symbol.iterator] = function() { ... };
```
This was indeed my initial approach. I'm not sure exactly why but it seems I had been (somewhat embarrassingly) overlooking something trivial which led me to believe there was an issue with this. 

Addressing (1) does seem like a fairly critical matter. To do that my first instinct would be to wrap the above definition in the body of an if statement which checks if the users browser is supported. How to actually check that though seems less straightforward as this is new territory for me. Using something like [this](https://stackoverflow.com/a/16938481) could work although it feels a bit inelegant and potentially error-prone. I'll have to spend some more time researching potential options.

maxkfranz:
(1) Probably something like this:

```js
const typeofUndef = typeof undefined;
const isIteratorSupported = typeof Symbol === typeofUndef || typeof Symbol.iterator === typeofUndef;
```

(2) The `this` keyword changes based on the called `function`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

You should generally just use arrow functions.  The main exception would be a function defined on a `prototype`, e.g. `SomeClass.prototype.foo = function(){ /* ... */ }`.  Any function defined within `foo`, for example, should be an arrow function.

If you use `class`es and arrow functions, you don't generally need to use the `function` keyword at all.  Cytoscape does not use the `class` keyword, because the lib predates it.  The classes in Cytoscape are also large, so it's better to have them span multiple files with `prototype`.

Rule of thumb:

`obj.foo()` : `function` keyword
`foo()` : `=>`

maxkfranz:
> In terms of writing documentation for this some soft guidance would be greatly appreciated, as this falls a bit outside of the eles.method(...) convention.

It's probably fine to just mention in the introduction markdown file for `Collection` that the collections are iterable.  That could link to this for more details: https://exploringjs.com/es6/ch_iteration.html

maxkfranz:
Ref : https://github.com/cytoscape/cytoscape.js/blob/unstable/documentation/md/collection/collection.md

src/collection/index.js
```diff
@@ -105,6 +105,20 @@ let Collection = function( cy, elements, options ){
...
+      next: function() {
+        this._values = this._valueIterator.next();
+        !this._values.done ? this._values.value = this._values.value.ele : null;
+        return this._values;
+      },
+      _valueIterator: map.values(),
+      _values: undefined
+    };
+  };
+
```
maxkfranz: It would be better to use an index rather than depending on `values()` working for another object.  Here's an example:

https://github.com/cytoscape/cytoscape.js/blob/v3.12.1/src/util/hash.js#L22-L40

r-ba:
Thanks Max, I really appreciate your patient and informative feedback.

Some quick notes about the most recent changes:
- Collection iterators are now also iterable. I've added an additional test for this.
- I used your solution to (1) but modified it so that we don't accidentally check `undefined` for an iterator property. I also removed the `const typeofUndef` declaration, instead using the string "undefined", with the thought process that it would be extraneous-- however I'm now second guessing that decision. Aside from it perhaps being also more stylistic to declare the constant, is there any potential for `typeof undefined` to not be `"undefined"`?

src/collection/iteration.js
```diff
@@ -122,29 +122,34 @@ let elesfn = ({
...
+          } else {
+            entry.value = undefined;
+            entry.done = true;
+          }
+
+          return entry;
+        },
+        [Symbol.iterator]: function() { // eslint-disable-line no-undef
+          return this;
         }
```
maxkfranz: I can't think of a case where consumer code would have direct access to the object defined on line 135, so I'm not sure about the purpose of defining `[Symbol.iterator]` on that iterator object.  It probably does no harm, but I'm curious about the motivation for including it.

r-ba: I may be misunderstanding your question, but the consumer may provide access to the object with something like `let iter = eles[Symbol.iterator]()`. So then the purpose of defining a `[Symbol.iterator]` property on this object is solely to permit the consumer to be able to treat `iter` as an iterable. A potential (albeit somewhat contrived) use case might be that you want to partially iterate over a collection but still be able to resume iteration from where it left off in a separate loop. An example of this behaviour can be found in the updated test file.

As far as motivation goes, that simply came about after reading the advisory on the [iteration protocols MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) (right below the section "the iterator protocol" in the highlighted yellow rectangle), also mentioned [here](https://exploringjs.com/es6/ch_iteration.html#sec_implementing-iterables) (under section 21.5.1). It appears that built-in es6 iterators also follow this pattern. All that being said, even in my limited experience "because I read about it somewhere" intuitively seems like a bad justification for implementing something. So I'd be entirely comfortable with it if you took it out.

maxkfranz: OK, let's leave it in as long as it's undocumented.  That way, it's easy to publicly release that functionality in future with a documentation update.  Since it's undocumented, we also have the option to remove it in future if it is problematic.

PR #2612, opened 2020-01-26
Add support for negative `taxi-turn` values

herzaso:
A negative value specifies a distance relative to the target side (from target to source) rather than the default source side (from source to target).

### DIFF
--- a/documentation/index.html
+++ b/documentation/index.html
@@ -13589,7 +13589,7 @@ Sets the subject of the selector (e.g. <code>$node &gt; node</code> to select th
 </ul>
 </li>
 <li><strong><code>taxi-turn</code></strong> : The distance along the primary axis where the first turn is applied.<ul>
-<li>This value may be an absolute distance - positive number (e.g. <code>20px</code>) to indicate distance from the source or negative number (e.g. <code>-20px</code>) to indicate distance from the target or it may be a relative distance between the source and target (e.g. <code>50%</code>).</li>
+<li>This value may be an absolute distance (e.g. <code>20px</code>) or it may be a relative distance between the source and target (e.g. <code>50%</code>).</li>
 <li>Note that bundling may not work with an explicit direction (<code>upward</code>, <code>downward</code>, <code>leftward</code>, or <code>rightward</code>) in tandem with a turn distance specified in percent units.</li>
 </ul>
 </li>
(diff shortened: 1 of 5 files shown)


documentation/index.html
```diff
@@ -13589,7 +13589,7 @@ <h2 id="style/taxi-edges">Taxi edges <a href="#style/taxi-edges"><span class="fa
 </ul>
 </li>
 <li><strong><code>taxi-turn</code></strong> : The distance along the primary axis where the first turn is applied.<ul>
-<li>This value may be an absolute distance (e.g. <code>20px</code>) or it may be a relative distance between the source and target (e.g. <code>50%</code>).</li>
+<li>This value may be an absolute distance - positive number (e.g. <code>20px</code>) to indicate distance from the source or negative number (e.g. <code>-20px</code>) to indicate distance from the target or it may be a relative distance between the source and target (e.g. <code>50%</code>).</li>
```
r-ba: Note that, in general, changes to the documentation should be made in their respective markdown files (and, when applicable, `docmaker.json`) as `index.html` is auto-generated from these. 

In this particular instance, the appropriate file to modify would be `documentation/md/style.md`

herzaso: Sorry. Totally missed it. Fixed

maxkfranz:
I'm getting odd behaviour for the following style in the demo page:

```js
      .selector('[source = "b"][target = "f"]')
        .style({
          'curve-style': 'taxi',
          'taxi-direction': 'vertical',
          'taxi-turn': '-50%' // n.b. negative
        })
```

The above example should give the same result as the one below:

```js
      .selector('[source = "b"][target = "f"]')
        .style({
          'curve-style': 'taxi',
          'taxi-direction': 'vertical',
          'taxi-turn': '50%' // n.b. positive
        })
```

A value of `-50%` should give the same visual result as `50%` --- a break in the middle.  However, the `-50%` case is uneven.

maxkfranz:
Thanks for the pull request!

herzaso:
negative percentile seems a bit redundant, but I fixed it for completeness

maxkfranz:
I agree it's redundant, but inevitably someone will try to do it.  It may make more sense to some people to use negative percent values in certain cases.  Alternatively, some people may use a mapper function whose formula sometimes gives negative values.

This pull request has been merged in and the feature will be available in 3.14.0.

Thank you for the pull request.

PR #2616, opened 2020-01-29
Skip invalid edges

tothandras:
I assume the original intention was to ignore invalid edges instead of breaking with an error.
This commit introduced the breaking change: https://github.com/cytoscape/cytoscape.js/commit/1f1b5be35f5d576172a9c7855b84303e304dc0da

### DIFF
--- a/src/collection/index.js
+++ b/src/collection/index.js
@@ -394,13 +394,13 @@ elesfn.restore = function( notifyRenderer = true, addToPool = true ){
       data.id = '' + data.id; // now it's a string
 
     } else if( is.emptyString( data.id ) || !is.string( data.id ) ){
-      util.error( 'Can not create element with invalid string ID `' + data.id + '`' );
+      util.warn( 'Can not create element with invalid string ID `' + data.id + '`' );
 
       // can't create element if it has empty string as id or non-string id
       removeFromElements();
       continue;
     } else if( cy.hasElementWithId( data.id ) ){
-      util.error( 'Can not create second element with ID `' + data.id + '`' );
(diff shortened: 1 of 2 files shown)


maxkfranz:
Consumers don't generally follow warnings, even experts like programmers who are consumers of a library.  If there were only a warning when an element is erroneously omitted, then it risks that the problem goes unnoticed and unfixed.

Having missing elements is a severe problem.  If an element should be seen by a user but it is unseen, then the user is mislead.  Users can make important decisions based on the presented data.  Compromising the integrity of the data, and possibly misleading the user, is unacceptable.

When there is an error instead of a warning, the consumer has no choice but to address the error in their app.  Anyone can notice the error -- programmers, QA analysts, and users alike.  Severe errors in an app should be quickly and easily discoverable.

Thank you for your feedback, and thank you for the pull request.  I understand that you would like the API to be convenient for your purposes, but I think you can appreciate why the API ought to remain as it is.

tothandras:
Thanks for you response @maxkfranz, how about making it a configurable behaviour?

maxkfranz:
The intention of the error is to make the consumer handle his data properly
and explicitly.  It is possible to sanitise, fix, or  ignore invalid
elements by an app's own code, thereby making the handling of invalid cases
explicit in the app itself.  Given that, I don't think I can justify the
existence of an option to disable the error outright for the sake of
expedience.

Further, I think we should generally strive to simplify and to have few
options.  The opinion in the article, "Why not add an option for that?"
<http://neugierig.org/software/blog/2018/07/options.html>, is largely
similar to my own.

Thank you for your input.

PR #2622, opened 2020-02-06
Patch `buildComponent` in `eles.htbc()`

r-ba:
In the current implementation of the `buildComponent` function within `eles.htbc()` there is a flaw such that, for certain graphs, a singular edge (or multiple, counting loops) will be omitted from a `components` collection during it's respective construction. A minimal reproducible example of this behaviour can be seen [here](https://codepen.io/r-ba/pen/jOPObZz). The intent of this PR is to amend this issue.


maxkfranz:
Merged in for 3.13.1

Thanks

PR #2630, opened 2020-02-12
Fix Firefox edge arrow performance

### LINKED ISSUE #2367: Severe performance degradation with edge arrows in firefox
**Issue type**

Bug report

**Environment info**

- Cytoscape.js version : >= 3.3.0
- Browser/Node.js & version : Firefox 66.0.3 (64-bit)

**Current (buggy) behaviour**

Over time, Firefox takes longer and longer to draw edge arrows such that even a modestly sized graph of 30 nodes can become unusable.

**Desired behaviour**

Draw quickly

https://jsbin.com/lacicozeco/
Drag nodes around, and it becomes noticeably choppier over time.  This does not happen in chrome, or with the 3.2 line of cytoscape (change the link to cytoscape@3.2.22), or with target-arrow-shape set to "none".

mohd-akram:
Fixes #2367.

### DIFF
--- a/src/extensions/renderer/canvas/drawing-edges.js
+++ b/src/extensions/renderer/canvas/drawing-edges.js
@@ -278,17 +278,15 @@ CRp.drawArrowShape = function( edge, context, fill, edgeWidth, shape, x, y, angl
     }
   }
 
-  if( context.beginPath ){ context.beginPath(); }
-
   if( !pathCacheHit ){
+    if( context.beginPath ){ context.beginPath(); }
     if( usePaths ){ // store in the path cache with values easily manipulated later
       shapeImpl.draw( context, 1, 0, { x: 0, y: 0 }, 1 );
     } else {
       shapeImpl.draw( context, size, angle, translation, edgeWidth );
     }
+    if( context.closePath ){ context.closePath(); }
   }
-
-  if( context.closePath ){ context.closePath(); }
 
   context = canvasContext;
 


maxkfranz:
The `beginPath()` and `closePath()` calls are there because, at least at one point, one or more arrow shapes required it.  Those calls are known to be expensive, and so they wouldn't be there unless absolutely necessary.  I can't say offhand whether these calls are solely necessary for the `Path2D` case, as your PR implies.

It's great that you've confirmed that this solves a Firefox-specific performance issue on your machine.  Before merging this PR, I'd like to see confirmation that the existing edge arrows are not negatively affected by this change.  The [edge arrow type demo](https://js.cytoscape.org/demos/edge-arrows/) would be a good way to test this.  (The default debug page does not contain all the arrows.)

To test, we should confirm that all edge arrows work properly in the demo for the following cases:

(1) `Path2D` is in effect.  Selecting edges in the demo should cause redraws that should be cache hits.

(2) `Path2D` is not in effect.  You can set `window.Path2D` to a nully value to trigger this.  Or, I suppose, you could comment out the path cache creation code to cause the same always-miss behaviour.

@mohd-akram, would you be able to confirm these cases?

mohd-akram:
This change wraps both the `Path2D` code and the regular code, so it's functionally identically to what was before - all it does is prevent `context.beginPath()` being followed by `context.closePath()` without any actual drawing in between (which is most frames since the path is cached). I previously tried what I said in the issue comment, removing them entirely, and arrows didn't work properly with `Path2D` set to undefined.

PR #2633, opened 2020-02-21
Correction in documentation

treecoder:


gurisko:
I am sorry, but that is not correct.

maxkfranz:
I can appreciate the desire to make things feel inclusive, but I don't think that this is the forum for debates on grammar (i.e. whether "he" is a generic, gender-neutral pronoun just as "il" is in French) or political correctness.  (And if we want to consider inclusiveness generally, we should really take into account consumers for whom English is a second language.  A singular use of "they" can be confusing to them, and they currently make up about half of the population of Cytoscape users).

I'm locking this thread down.  Any future thread that touches on political correctness will also be locked down.  I would be open to accepting a pull request in the docs to the effect of "Any use of the word 'he' in this documentation can be read as 'he or she', but 'he' may be used in the interest of brevity."

Let's focus our effort on the main aim of Cytoscape:  Make it easy and efficient for people to use interactive networks on the web.  I'm here to get things done, and I hope you are too.

PR #2640, opened 2020-02-28
Add new label background shapes: circle and ellipse

jan-san:
This PR adds two new options for `text-background-shape`: `ellipse` and `circle`.

The diameter of the circle shape can be configured via the new option `text-background-shape-circle-size` to equal the label's width/height or the min/max of these.

The PR also extends the documentation for `text-border-opacity`, noting that the border is always rectangular (which also applies to `text-background-shape = 'round-rectangle'`).


jan-san:
This PR includes an unrelated commit - will fix and resubmit.

Issue #1758, opened 2017-03-27
Remove all shadow style properties


maxkfranz:
It is not possible to properly support shadow properties.  The canvas [does not take scaling into effect](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowBlur).  The workaround to increase the blur radius with the zoom level can not work for image exporting.  The [shadow blur also doesn't correspond to a size in pixels](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowBlur), so it's not possible to use it reliably with current and future performance enhancements.

The following properties are removed:

- `shadow-blur`
- `shadow-color`
- `shadow-opacity`
- `shadow-offset-x`
- `shadow-offset-y`
- `text-shadow-blur`
- `text-shadow-color`
- `text-shadow-opacity`
- `text-shadow-offset-x`
- `text-shadow-offset-y`

Shadows can be recreated with the planned ghost style https://github.com/cytoscape/cytoscape.js/issues/1197.  The ghost style can not support blurring, however.

PR #2642, opened 2020-02-28
Add support for label background shade

jan-san:
This PR adds support for configuring shadows of label backgrounds.

Below is an example of an edge label using `circle` background shape (see #2640) and shadows:
edge with circle shaped label background and shadow

A limitation of this implementation is that the shadow gets cropped if it extends beyond the dimensions of the label box. Suggestions to overcome this would be welcome.

### DIFF
--- a/documentation/md/style.md
+++ b/documentation/md/style.md
@@ -616,6 +616,14 @@ Border:
  * **`text-border-style`** : The style of the border around the label; may be `solid`, `dotted`, `dashed`, or `double`.
  * **`text-border-color`** : The colour of the border around the label.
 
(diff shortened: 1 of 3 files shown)


jan-san:
Could you have a look at this @maxkfranz?

maxkfranz:
Thanks for the PR, but unfortunately this isn't a feature that will work properly w.r.t. scale transformations.  It might seem to work OK in your initial tests -- I made the same mistake before -- but this approach does not work generally.  See https://github.com/cytoscape/cytoscape.js/issues/1758

PR #2644, opened 2020-03-03
Add new label background shapes: circle and ellipse (fixed)

jan-san:
This is a fixed version of #2640. The PR adds two new options for text-background-shape: ellipse and circle. 

The diameter of the circle shape can be configured via the new option text-background-shape-circle-size to equal the label's width/height or the min/max of these.

The PR also extends the documentation for text-border-opacity, noting that the border is always rectangular (which also applies to text-background-shape = 'round-rectangle').

### DIFF
--- a/documentation/md/style.md
+++ b/documentation/md/style.md
@@ -607,12 +607,13 @@ Background:
 
  * **`text-background-color`** : A colour to apply on the text background.
  * **`text-background-opacity`** : The opacity of the label background; the background is disabled for `0` (default value).
- * **`text-background-shape`** : The shape to use for the label background, can be `rectangle` or `round-rectangle`.
+ * **`text-background-shape`** : The shape to use for the label background, can be `rectangle`, `round-rectangle`, `ellipse` or `circle`.
+ * **`text-background-shape-circle-size`** : Method for determining circle size (only applicable for `text-background-shape = 'circle'`). Can be `height`, `width`, `min` or `max` (default value).
  * **`text-background-padding`** : A padding on the background of the label (e.g `5px`); zero padding is used by default.
 
 Border:
 
- * **`text-border-opacity`** : The width of the border around the label; the border is disabled for `0` (default value).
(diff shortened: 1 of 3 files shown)


maxkfranz:
Would you please post the motivating use cases, along with some figures or screenshots?

jan-san:
Our use case for the circle shape is to use it as a background for a (font) icon we display as edge label as shown in the screenshot below: 
image

The  combination of circle background + icon is commonly used.

Having only rectangle and rounded rectangle background shapes limits the choices for styling graphs. Given that the PR only adds about 30 lines of code to the project and should have next to 0 performance impact, I would say that this is a reasonable trade-off for the additional styling options it provides.

maxkfranz:
OK, I can understand why you may want to label an edge like that.  I wonder if the proposed change could be made simpler and more consistent.  For example, I don't understand the motivation for distinct `circle` shapes versus `ellipse` shapes.

jan-san:
The reason for having a separate `circle` shape was that icon font characters may have a non-square aspect ratio, but still look better in a circle background than in an ellipse background (e.g. a simple exclamation mark).

On the other hand, there can be situations with longer labels where the `circle` shape is impractical because the circle would become too large.

If you are concerned about the complexity of the added code, then removing the circle size modes (and always using the maximum size) might be a good option to simplify without losing much functionality.

maxkfranz:
(1) I strongly believe that there shouldn't be a separate `circle` shape.  It should be general and apply to all shapes.  Similarly, the border should work for all shapes -- not exclusively the rectangle ones.  Is there a way to make this work for all shapes (ellipse, rectangle, roundrectangle)?

(2) We need to carefully consider any property that would affect the bounding box of an edge.  That has implications to the renderer caches and also to the `dimensions.js` functions.  So if you want to make a change like this, the change needs to be reflected to all effected parts of the code.

(3) Could this feature somehow leverage padding?

maxkfranz:
As an aside:  Minimising complexity for consumers of the library is of more priority than minimising complexity of the codebase, though both are good to have.

PR #2654, opened 2020-03-18
Add curve-style: bezier to the tutorial example

icalvo:
**Issue type**

Documentation bug report

**Current (buggy) behaviour**

Since version 2.7.0, edges are haystack by default, which only support midpoint arrows. The example is supposed to show an arrow and it doesn't.

**Desired behaviour**

I have added a curve style that allows it to be shown.


maxkfranz:
Thanks for the PR.  You’ve modified the generated file rather than the source markdown files, from which the documentation is built.  Would you edit the corresponding markdown file and run `npm run docs` to verify?

icalvo:
Done, I've reverted the former commit, edited only the md and verified it with 'npm run docs'.

maxkfranz:
Merged in.  Thanks

PR #2663, opened 2020-04-13
Added circle-triangle edge arrow type

kaino:
**Issue type**
Feature request

Edge arrows as the combination of a triangle and a circle.

In the last two years I've been using with great satisfaction cytoscape.js to manipulate graphs. 
The original technology that I replaced with cytoscape, used the triangle + circle arrow style to represent the "owner" of a connection.
Since in my context such arrow style is part of a standard, I must keep it the same to avoid disruption or misunderstandings, so I would really like to have it as part of the cytoscape.js project.

Tests are passing, documentation + edges demo are updated accordingly.
Please let me know if anything is not clear, or in case of an issues with this pull request.

Thank you very much
Best Regards,
Paolo


maxkfranz:
The PR looks great.  My only suggestion would be to call the arrow type `circle-triangle`.  

Other arrows that have multiple components, like `triangle-tee`, list the components starting from the end of the line and going inwards.  For `triangle-tee`, the triangle is at the end of the line and the tee is inward:

Screen Shot 2020-04-17 at 1 23 14 PM

 

For your shape, the circle is at the end of the line and the triangle is inward:

Screen Shot 2020-04-17 at 1 19 11 PM
  
So `circle-triangle` may be a better fit.

If we support a new shape in future that has the triangle at the end and the circle going inwards, then the distinction between `triangle-circle` and `circle-triangle` will be important.

kaino:
Hi @maxkfranz ,
thanks for reviewing this.
Actually, an open point for me was just the choice between 'triangle-circle' and 'circle-triangle'.
I've made the suggested changes on the PR, and tested it locally, everything should be fine.
Please let me know for any issues about the new commits.

Thanks again
Best Regards,
Paolo

maxkfranz:
Great.  The PR is merged in and the feature will be available in the 3.15.0 release at the beginning of May.

Thanks for the PR

PR #2670, opened 2020-04-21
panningWithRightMouseButton implemented

marcidk:
**Issue type**

Feature request

Option for using the right mouse button for panning instead of the left mouse button.

In our use case, we have other graph tools and we want to keep the general behavior of the graphs the same. That means:
* LMB click should select
* LMB drag should be box selection
* RMB click should be menu
* RMB drag should be panning

This also simplifies the usage of the graph (no need for multiple keys at once), and at least from our point of view it seems to be a bit more inline of the user expectation.

The functionality has been implemented with the initialization option "panningWithRightMouseButton". It is default false (still using LMB) and no breakable changes.
Debug page and documentation has been updated too.

I dont know whether it is a feature you would want merged in, however it feels too core'ish to somehow put into an extension (it changes basic behavior, not adding new). Anyways, I would very much like to give you the option, now that it is already made :)

In any case, let me know of any questions, comments or issues.

Best regards,
Jeppe

### DIFF
--- a/debug/index.html
+++ b/debug/index.html
@@ -82,6 +82,10 @@
         User:
         <button id="enable-user-panning">on</button>
         <button id="disable-user-panning">off</button><br />
+        
+        Pan with Right Mouse Button
+        <button id="enable-panning-with-right-mouse-button">on</button>
+        <button id="disable-panning-with-right-mouse-button">off</button><br />
 
         Zooming:
         <button id="enable-zooming">on</button>
--- a/debug/view.js
+++ b/debug/view.js
@@ -34,6 +34,14 @@
 		cy.userPanningEnabled(false);
 	});
 
+	$("#enable-panning-with-right-mouse-button").addEventListener('click', function(){
+		cy.panningWithRightMouseButton(true);
+	});
+
+	$("#disable-panning-with-right-mouse-button").addEventListener('click', function(){
+		cy.panningWithRightMouseButton(false);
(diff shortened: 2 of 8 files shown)


maxkfranz:
I appreciate the motivation for this pull request.  However, this feature would indeed be better as an extension.  

The library supports the most ubiquitous gestures.  It does not support customising what low-level events are a part of particular gestures.  Someone may want to use the scroll wheel to pan.  Someone else may want to use right click for a menu.  Yet another person may want to tap and hold on the background to fit the graph to the screen.  All of those gestures are valid for certain kinds of apps, but they should not be directly implemented by the core library.

If a gesture would be often-used by most end-users, then it may be a good fit for the core library.  If not, the gesture is should generally be an extension.

I hate to reject such a well-written pull request, but I don’t think that inviting low-level gesture customisation within the core library would be good for the longterm health of the project.

There are improvements to the documentation for extension writers in 3.15.0.  That should help you to get started.  Your extension could disable the default user panning and replace it with whatever gestures makes sense for your users (e.g. right-click-swipe to pan).  

See https://github.com/cytoscape/cytoscape.js/issues/2679

marcidk:
No problem at all, I understand where you are coming from.
However I believe it will be too much work to make and duplication of core code to maintain, if it was an extension, so at least for now, I do not expect to be implementing it as an extension.
Cheers

PR #2697, opened 2020-05-28
docs: fix typo in code example

SimonGolms:
**Issue type**

Typo

_IMHO: I know this PR doesn't mean much and could be a bit picky, but before anyone else is confused about it, I thought it might be better to fix it._


maxkfranz:
Great, thanks

PR #2724, opened 2020-07-10
Fix typo in example code comment (letiation -> variation)

hipe:
I think what probably happened was someone originally wrote "variation" and then there was a search and replace for /^var/ to "let" 😂

### DIFF
--- a/dist/cytoscape.cjs.js
+++ b/dist/cytoscape.cjs.js
@@ -19435,7 +19435,7 @@ var defaults$b = {
     return node.degree();
   },
   levelWidth: function levelWidth(nodes) {
-    // the letiation of concentric values in each level
+    // the variation of concentric values in each level
     return nodes.maxDegree() / 4;
   },
   animate: false,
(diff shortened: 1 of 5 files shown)


maxkfranz:
You're right.  I think the source md file needs to be updated, though.  The html just comes from a template.  The html is committed so that there's a reference to old versions of the docs by tag.

maxkfranz:
@hipe, are you able to update the source file?

hipe:
indeed! I have added a commit that updates the source file (afaik)

`src/extensions/layout/concentric.js`

As well, this branch commits the resulting change to the several generated files affected by this.

let me know if i can clean up this timeline at all (so it's one commit)! If so, not sure of the best way (a new pull request starting from scratch?) it's such a small change!

Thanks!

maxkfranz:
Looks good.

Don't worry about cleaning up the commits.  The built files should automatically get cleaned up with a fresh build prior to release.  The unstable branch is for collaboration and new ideas, so keeping that branch clean isn't a relatively high priority anyway.

Thanks for the PR

Issue #2761, opened 2020-10-14
Ellipsis text wrap option will put ellipsis when the label could fit in full


msalihaltun:
**Issue type**

Bug report

**Environment info**

- Cytoscape.js version : 3.16.2
- Browser/Node.js & version : Chrome Version 86.0.4240.75 (Official Build) (64-bit). See JSBin for all else.

**Current (buggy) behaviour**

When `text-wrap` is set to `ellipsis`, cytoscape will put ellipsis at the end of the node label even if the label could have fit at full size. The logic to put ellipsis is [here](https://github.com/cytoscape/cytoscape.js/blob/a0cb418ab125c7fb12b07f513ad280ce708b1d63/src/extensions/renderer/base/coord-ele-math/labels.js#L412) and it tries to put ellipsis and then test if the width with ellipsis and next character exceeds max width, and if it does, it will put the ellipsis. But when the next character is the last character, it won't check whether that character could have fit without adding the ellipsis. It doesn't help that the ellipsis will always be wider than a single character, so we end up with ellipsis and a wider label instead of just the last character.

**Desired behaviour**

Cytoscape should be able to check if the full label will fit when checking the character before the last one, so that the label won't end up putting ellipsis in place of a single character.

**Minimum steps to reproduce**

See the [JSBin output](https://output.jsbin.com/jineyemexi). We have two nodes with the same label, and equal values for `width` and `text-max-width`. One of them has `text-wrap` set to `wrap` and the other has it set to `ellipsis`. The node with `text-wrap: ellipsis` will put ellipsis instead of one character, when it is obvious that the label could fit inside the node as in the node with `text-wrap: wrap`.

maxkfranz:
@msalihaltun, it sounds like you already have a handle on the precise cause of the issue.  You're welcome to make a PR, if you like.  That would help to get the patch released faster.  Thanks

blazespinnaker:
I've run into something similar, I have just worked around it.  @msalihaltun if you don't want to do the builds / testing, you can paste in the code here (or in my repo) and I can do it in my build / test and you can do the PR.  I invited you to my repo.

blazespinnaker:
cool.  a little off topic but is it me, or does this code look a little problematic from a perf point of view:

for( let i = 0; i < text.length; i++ ){
      let widthWithNextCh = this.calculateLabelDimensions( ele, ellipsized + text[i] + ellipsis ).width;	     

[...]

PR #2762, opened 2020-10-15
Change ellipsis label logic

msalihaltun:
- Fixes cytoscape/cytoscape.js#2761
 - New logic will first test if the full label fits in the node, if it does, use the full label. Otherwise, the old logic will apply since we know we are going to be using ellipsis (label doesn't fit).


maxkfranz:
Looks good.  Thanks for the PR

PR #2763, opened 2020-10-18
"npm run benchmark" runs all benchmarks

joliroot:
https://github.com/cytoscape/cytoscape.js/blob/unstable/README.md#build-instructions

According to docs "npm run benchmark" should run all benchmarks, but currently only runs single benchmark.
This was changed so that it aligns with the docs.


maxkfranz:
Great.  Thanks for the PR

Issue #2770, opened 2020-11-07
Uncaught TypeError: Cannot read property 'w' of null


janhartmann:
**Issue type**

Bug report

**Environment info**

- Cytoscape.js version : 3.16.3
- Browser/Node.js & version : Chrome Version 86.0.4240.111 (Official Build) (x86_64)

**Current (buggy) behaviour**

We have a network graph that is powered by a web socket connection and it is updated/redrawn extremely often. Sometimes we notice that Cytoscape is throwing an exception when doing a lot of redraws. Currently I have not found the root cause for this, but while doing my own investigation I am doing this bug report. 

![image](https://user-images.githubusercontent.com/1107383/98434140-7c9a8380-20e6-11eb-8038-b9307fb91397.png)

It appears that the element getting the bounding box from is suddendly gone, when calling this: https://github.com/cytoscape/cytoscape.js/blob/bb9cdd93b45e12864d11beba75cb65e90ec3df9a/src/extensions/renderer/canvas/ele-texture-cache.js#L485

**Desired behaviour**

Handle the case that there is no bounding box anymore for the element (if it has been removed during fast re-draw) 

**Minimum steps to reproduce**

This error happends randomly in our application and I am unable to pin point exactly why is does this. I am suspecting a race-condition where the element is removed while running the `getElement` method.

PR #2771, opened 2020-11-07
Short circuit when `bb` is not defined

janhartmann:
See https://github.com/cytoscape/cytoscape.js/issues/2770


maxkfranz:
The `!bb` check makes sense.  Would it be a bit more robust if there were also a `!ele.removed()` check at the end of the same `if` statement?

It looks like an element is being queued up for caching, and it's being dequeued during idle time (i.e. time between redraws).  If the element has been removed, then there isn't much point in filling in its cache during that idle time.

janhartmann:
That makes sense. I added it to the branch.

maxkfranz:
I'll publish the releases today

PR #2781, opened 2020-11-25
Feature/relation arrow shapes

mscale-wmp:
**Issue type**

Feature request

New feature that adds two additional arrow shapes to cytoscape's edge points; one and many. This follows the data cardinality model.

This new feature allows users to better represent data lineage graphs. With the 'one' edge, users can show a one relation, either on the target, source, or both. The same is for the many, crows feet, arrow shape. These styles are based on the business rules of a relation between two database tables. http://www2.cs.uregina.ca/~bernatja/crowsfoot.html

The motivation for this feature was based on need. For the current project that I am professionally working on, clients want to be able to see the cardinality between two database tables, which are represented by nodes. The edge lines are the relations and the arrow shape is meant to show what kind of relation exists. Since there was not a crows feet option, nor a single line with a gap between it and the node, I decided to make the options instead.


maxkfranz:
Thanks for your PR.  It's always great to add new, useful shapes.  Here's some initial feedback:

(1) Changes to dist files shouldn't be included in the PR.  Could you reset them?  Those files get updated only when making a new release.

(2) It would be great to show off your new arrows in the edge arrow demo.  Could you add them to the demo file?  See https://js.cytoscape.org/demos/edge-arrows/

(3) The naming of the arrows could be made a bit more inclusive.  I understand that you named them according to the domain of your app, but shapes should be given general names.  Cytoscape gets used in many domains.  So using general names would make the arrows easier to understand.  Maybe something like `crows-foot`, `double-tee`, etc.?

(4) The arrow types should be documented.  Could you add them here? https://github.com/cytoscape/cytoscape.js/blob/unstable/documentation/md/style.md

mscale-wmp:
Max, thank you! I will make the changes you requested and should have an updated PR by EOD.

documentation/index.html
```diff
@@ -13686,6 +13686,8 @@ <h2 id="style/edge-arrow">Edge arrow <a href="#style/edge-arrow"><span class="fa
 <li><code>circle</code></li>
 <li><code>diamond</code></li>
 <li><code>chevron</code></li>
+<li><code>thick-tee</code></li>
+<li><code>crows-feet</code></li>
```
maxkfranz, line 13690: Docs should be edited in the md files.  The html is just generated from a template and would be overwritten.

maxkfranz:
I've tried this out a bit more, and I've noticed a few things that should be addressed before merging:

(1) The new shapes should be added to the edge arrow demo: https://js.cytoscape.org/demos/edge-arrows/

(2) The `crows-feet` doesn't work on a curved edge, because it's using the edge line as the middle part of the arrow.  You can see this in the debug page:

Screen Shot 2020-12-04 at 13 47 42

Could the middle part of the arrow be specified as the part of the arrow shape itself?

(3) The new arrows are larger than the existing arrows.  Could they be scaled down to be comparable to the others?

mscale-wmp:
1. I thought that I had added them to the demo, I will look into updating the edge arrow demo files.

2. I'll look into adding a middle line to the crows feet so that it curves with curved nodes.

3. Sure, I'll adjust the arrow sizes to being similar to the other edge points.

maxkfranz:
@mscale-wmp, I'd like to be able to include these changes for 3.18.  Do you think you'll be able to finalise the PR this week?  I can push back the release date to next Monday.

mscale-wmp:
Absolutely. It will be done before Friday.

maxkfranz:
Great!

mscale-wmp:
The crows-feet edge arrow now has a proper middle. Took me longer than I'd like to admit to render it. I also updated its size as well as the thick-tee edge arrow. The edges are a part of the demo, I tested this locally.

mscale-wmp:
Is it the small gap that is off or something else? If you can give me more details I'll work on a better render. The code will not be ready today though, I am finishing a project with my company.

maxkfranz:
I think there's a few things here:

(1) There's a gap between the end of the line and the arrowhead.  That might be fixed by changing the something like the gap value of the arrow.

(2) There's an empty spot.

(3) Based on the specification page you linked to:  Each toe of the crow's foot should equal in size, and the thickness of each toe should be uniform.

maxkfranz:
Screen Shot 2020-12-18 at 13 54 39

maxkfranz:
@mscale-wmp, let me know if you have any other questions or concerns re. your implementation.  Thanks

maxkfranz:
@mscale-wmp, is there anything that you would like help with specifically to move the PR forward?  Thanks

mscale-wmp:
@maxkfranz  I'm sorry I have been a busy project with company since the year started. I'm trying to find some time after work to move forward with the feature. It's just been a hectic time.

maxkfranz:
OK, thanks for the update

mscale-wmp:
@maxkfranz  I am so sorry that is has been so long, but I was finally able to work on cytoscape again. I rerendered the vectors for the crows-feet arrow shape. There shouldn't be any gaps/spaces and each line should be the same width/thickness. Let me know if there are any problems.

maxkfranz:
It's looking much improved.  I have two questions:

(1) Is this the intended output of `thick-tee`?  It looks like the same size, with the tee offset.  If this isn't the intended output, then it should be revised.  If it is intended, is there a clearer name we could use?

Screen Shot 2021-02-17 at 10 40 38

(2) The crow's foot arrow looks much better.  There are two details that could be improved a bit:  Firstly, the middle toe is a bit thinner than the outer toes.  Secondly, the toes could be more adaptive to the edge width, especially with thick edges:

Screen Shot 2021-02-17 at 10 42 11

mscale-wmp:
Thank you! I'm glad that you like the improvements. 

For the thick-tee, initially I had it larger and thicker but you asked that I change the size to better match the other arrows. If I could make it slightly longer and thicker, then it would be more of what my team is looking for.

Regarding the crows feet, changing the thickness of the center line shouldn't be an issue. How would you like the crows feet to be more adaptive? Length change? Thickness? Are there any arrow shapes currently used that are a good example of adapting to edge width in the code?

maxkfranz:
> Thank you! I'm glad that you like the improvements.
> 
> For the thick-tee, initially I had it larger and thicker but you asked that I change the size to better match the other arrows. If I could make it slightly longer and thicker, then it would be more of what my team is looking for.

I wasn't very clear.  The main question is whether you intend for this arrow shape to be like the double-tee shape [in the U. Regina link you posted](http://www2.cs.uregina.ca/~bernatja/crowsfoot.html).  If so, then perhaps the shape should look more like what's in your link.  If not, then we need to work out exactly how it should look and be named.  It's fine as-is, if that's the intention, but the name isn't very clear:  Something like `offset-tee` would be clearer for the current arrow.

> Regarding the crows feet, changing the thickness of the center line shouldn't be an issue. How would you like the crows feet to be more adaptive? Length change? Thickness? Are there any arrow shapes currently used that are a good example of adapting to edge width in the code?

I think the main thing would be for the thickness to match up with the edge width.  I think there's a `triangle-cross` shape that does something like that.  If that's not easy, what about modifying the overall shape of the crow's feet?  Do they absolutely need to be boxy?  It would be easier to handle if they were just three triangles.  Here's a rough sketch with triangles of different colours, just to make things clear:

Untitled

Issue #2775, opened 2020-11-15
Node's background image rendering at different zoom levels


arnaudbuy:
**Issue type**

Bug report

**Environment info**

- Cytoscape.js version : 3.17.0
- Browser/Node.js & version : Google Chrome Version 86.0.4240.183 (Official Build) (64-bit)

**Current (buggy) behaviour**

![image](https://user-images.githubusercontent.com/7791859/99195896-f8509c00-2788-11eb-9d1e-bbad56be3423.png)

**Desired behaviour**

![image](https://user-images.githubusercontent.com/7791859/99195954-672df500-2789-11eb-8021-0a3a4bfc708b.png)

In my use, it would be cool if I could choose between pixelated or smooth rendering of node background images.

**Minimum steps to reproduce**

https://jsbin.com/genefidara/edit?js,output

maxkfranz:
Could you upload your image in a comment here?  Describe the image and its associated use case in your app.

arnaudbuy:
In my app, i use node's background images to show the underlying node layer function.

![image](https://user-images.githubusercontent.com/7791859/99606883-e847ef00-2a0a-11eb-8fde-dabe4e036ee7.png)

For example in this screenshot, the node represent a network layer which process a vertical edge.

![image](https://user-images.githubusercontent.com/7791859/99608143-9a80b600-2a0d-11eb-8681-c64a8ee26e7c.png)

I build the node image from a canvas for which I can set the dimensions and in which I draw the pixels.

![image](https://user-images.githubusercontent.com/7791859/99608799-ea13b180-2a0e-11eb-8d2d-eac017c1ab72.png)


The node image serve as well as a base to contruct a visualization of the network layer state.

arnaudbuy:
To summarize:
- Currently, the node background image changes from a smooth image to pixelated depending on the zoom level since version 3.17.0, which seems to be a bug.
- In my use case, I would like to be able to set the rendering of the background image of the nodes to smooth or pixelated regardless of the zoom level which seems to be a new feature.

arnaudbuy:
Yes it's true, I could have made even lighter, a 2x1 image, to represent a black side and a white side.

Ok, but I still can't explain the difference in behavior between version 3.16 and 3.17 ?

maxkfranz:
[...]

PR #2794, opened 2020-12-05
Adding background-image-smoothing property

arnaudbuy:
This PR adds a `background-image-smoothing` property, which can be either set to `yes` (default) or `no`.

#2775


.size-snapshot.json
```diff
@@ -1,39 +1,39 @@
 {
   "build/cytoscape.umd.js": {
```
maxkfranz, line 2: Could you reset this file?  It's only for releases.  Given the confusion it creates, it might be even better to not generate this file at all any more.

documentation/md/style.md
```diff
@@ -297,6 +297,7 @@ A background image may be applied to a node's body.  The following properties su
     * The [`cytoscape-sbgn-stylesheet`](https://github.com/PathwayCommons/cytoscape-sbgn-stylesheet) package serves as a good example for the use of SVG images in a stylesheet.  That stylesheet [creates decorations](https://pathwaycommons.github.io/cytoscape-sbgn-stylesheet/) on nodes in line with the [SBGN standard](https://sbgn.github.io).
 * **`background-image-crossorigin`**: All images are loaded with a [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin) attribute which may be `anonymous` or `use-credentials`. The default is set to `anonymous`.
 * **`background-image-opacity`** : The opacity of the background image.
+* **`background-image-smoothing`** : Determines whether background image is smoothed (`yes`, default) or not (`no`). This feature isn't well supported across browsers.
```
maxkfranz, line 300: Could this be made a bit more explicit that it's a only a hint?  The browser may or may not respect the value set for this property.

src/style/properties.js
```diff
@@ -48,6 +48,7 @@ const styfn = {};
     bgFit: { enums: [ 'none', 'contain', 'cover' ], multiple: true },
     bgCrossOrigin: { enums: [ 'anonymous', 'use-credentials' ], multiple: true },
     bgClip: { enums: [ 'none', 'node' ], multiple: true },
+    bgSmooth: { enums: [ 'yes', 'no' ], multiple: true },
```
maxkfranz, line 51: Should probably be named something like `bools` so that it could be used for other props in future.  See `bool` prop that follows

maxkfranz (commented):
Looks great.  I listed some minor feedback, and after addresses those the PR should be ready to merge.

Thanks for the PR

maxkfranz (approved):
Looks great

maxkfranz:
Merged.  Thanks for the PR

arnaudbuy:
Thanks for this merge.

PR #2803, opened 2020-12-27
Use only https assets in tokyo-railways demo

peey:
Firefox blocks loading of http scripts on  https pages by default, resulting in a broken demo.

See https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content#Types_of_mixed_content

**Issue type**
Issue on the website

**Current (buggy) behaviour**
Demo doesn't load in firefox

**Desired behaviour**

Demo should load in firefox

1. Open https://js.cytoscape.org/demos/tokyo-railways/ in firefox
2. Open console and see the warning about mixed content


maxkfranz:
Good catch.  Thanks for the PR!

PR #2804, opened 2020-12-30
[WIP] fns-order folder added

rahul799:


maxkfranz:
Great!  Could you specify the `fnsOrder` so that the previously defined ordering of the functions can be preserved in future (i.e. once the docmaker.js script is updated to read that field)?

rahul799:
Working on It.

rahul799:
@maxkfranz  I think we can get rid of the `@path` annotation now as we are specifying the function names in docmaker.json.

maxkfranz:
> @maxkfranz I think we can get rid of the `@path` annotation now as we are specifying the function names in docmaker.json.

That true only if every function is specified in the json file.  The ordering might not always matter, and it's good to show a result even if the function isn't specified in the json.

So I think it's good to keep the path annotation.  If a function is not specified in the json ordering, then it should just be included towards the bottom of the list -- after the ordered ones, if any.

rahul799:
Not stale. Working on it and will be finished soon.

rahul799:
Not stale

PR #2808, opened 2021-01-14
Clarify that cy.json requires element IDs(#2805)

### LINKED ISSUE #2805: Documentation: cy.json() requires edge id attribute, while normal graph initialization does not
**Environment info**

- Cytoscape.js version : 3.17.1
- Browser/Node.js & version : Firefox 84.0

**Current (buggy) behaviour**
If the edges passed to `cy.json()` do not contain an id attribute they are not created. I didnt find this mentioned anywhere
and the Documentation(https://js.cytoscape.org/#notation/elements-json) specifically states that the id attribute may be omitted at initialization (and from what i understand, updating a graph using `cy.json` and initializing it are basically the same thing)
There is also no error whatsoever.

FYI: This issue stems from a [networkx issue of mine](https://github.com/networkx/networkx/issues/4516)

**Desired behaviour**
If the behavior described aboved is intentional i expect Cytoscape.js to throw some kind of error if i try to update the graph without edge id. It would also be nice to mention that somewhere in the [Documentation](https://js.cytoscape.org/#cy.json).

If it is unintentional i would expect Cytoscape.js to consider all edges without ids to be new edges.

Initialize a Graph and try to update it using `cy.json({elements: {nodes:[ ... ], edges: [ ... ]}})` without providing edge IDs
Take a look at this [Example](https://jsbin.com/xefukoc/edit?html,js,console,output). I consider this either an unintentional bug
or a lack of Documentation as the [cytoscape demo tutorial information](https://js.cytoscape.org/#notation/elements-json)
states that the id attribute is inferred automatically if omitted.

Wuelle:
Closes #2805. Also fixes a typo in `debug/FileSaver.js` i randomly stumbled upon.

(I added an continue statement to skip elements without IDs, as they were previously added just to be removed again)

### DIFF
--- a/debug/FileSaver.js
+++ b/debug/FileSaver.js
@@ -30,7 +30,7 @@
     if (typeof opts === 'undefined') opts = {
       autoBom: false
     };else if (typeof opts !== 'object') {
-      console.warn('Depricated: Expected third argument to be a object');
+      console.warn('Deprecated: Expected third argument to be a object');
       opts = {
         autoBom: !opts
       };
--- a/documentation/md/core/json.md
+++ b/documentation/md/core/json.md
@@ -12,6 +12,8 @@ When setting `cy.json({ elements: ... })`
 * the included elements not in the graph are added, and
 * the not included elements are removed from the graph.
 
+Note that updating the Graph elements using `cy.json()` requires all elements to have an ID attribute. Elements that do not have an ID will be ignored.
+
 When setting `cy.json({ style: ... })`
 
 * the entire stylesheet is replaced, and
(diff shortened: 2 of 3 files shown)


debug/FileSaver.js
```diff
@@ -30,7 +30,7 @@
     if (typeof opts === 'undefined') opts = {
       autoBom: false
     };else if (typeof opts !== 'object') {
-      console.warn('Depricated: Expected third argument to be a object');
+      console.warn('Deprecated: Expected third argument to be a object');
```
maxkfranz, line 33: This file is just copied from here: https://github.com/eligrey/FileSaver.js/

I suppose they have, or had, a typo in their repo.  No harm in changing that here, but the file is only used for debugging anyway.  Users would never see it.

maxkfranz (approved):

maxkfranz:
Great.  Merged in for a future 3.18 release.

Thanks for the PR!

PR #2778, opened 2020-11-21
Feature: Background Image Containment

janhartmann:
This PR adds a `background-image-containment` property, which can be either set to `inside` (default) or `over`. 

When using `over` the drawing order is changed to draw background images on top of borders.

PR #2819, opened 2021-02-09
Fix path to cytoscape.min.js in background-image-containment demo

zakjan:
Fixes deployed demo for #2778


maxkfranz:
Great.  Thanks, @zakjan!

janhartmann:
Oopps, that was my mistake.

Issue #2830, opened 2021-03-06
A* reports wrong edges if the heuristic leads it along a sub-optimal path first


phdoerfler:
**Environment info**

- Cytoscape.js version : 3.18.1
- Browser/Node.js & version : Chrome Version 88.0.4324.190 (Official Build) (64-bit)

**Current (buggy) behaviour**

running A* with a non perfect but still admissible heuristic results in edges in the found path that can not be part of the found path.
I speculate that the implementation keeps (and thus reports back to the caller) edges that were once part of the tentative best path before it eventually found a better path.

**Example**

![image](https://user-images.githubusercontent.com/163428/110204377-f105e900-7e72-11eb-9006-7146c7f3f8c2.png)

In this example the nodes have their heuristic value printed on them as labels and A* found a path from left to right. I assume the heuristic first made it try to go horizontal, again horizontal **and then downwards**. The edge allowing that last traversal step is then returned to the caller as part of the shortest path found even though that tentative path was eventually discarded in favor of an even shorter path.
In addition to that, in its stead there should have been a different edge reported, the one that is clearly missing in the highlighted path in above screenshot: The missing edge between the node marked 6 and the node marked 5 to its right.

If I disable the heuristic, all edges returned by A* are indeed part of the shortest path and connect the found nodes with each other.

**Desired behaviour**

I expect the A* implementation in Cytoscape to return a correct list of edges.

**Example Code**

[Example on JS Bin](https://jsbin.com/sasuboboya/1/edit?js,output)

phdoerfler:
Please let me know if you require a minimal working example to reproduce this. I'm using Cytoscape through Scala.js and in turn through facades generated by ScalablyTyped based on Cytoscape's typescript definitions. I try to avoid touching Javascript and my hope is that my analysis so far might make it easy enough to fix the bug without a JS example. But if necessary I'll create one.

phdoerfler:
I have added an example on JS Bin

phdoerfler:
[Travis](https://travis-ci.com/github/phdoerfler/cytoscape.js/jobs/488786971#L334-L335) shows the two failing tests I added. The edges returned by the A* search do not necessarily form a continuous path.

phdoerfler:
in:
```js
if( !isInOpenSet(wid) ){
  gScore[ wid ] = tempScore;
  fScore[ wid ] = tempScore + heuristic( w );
  addToOpenSet( w, wid );
  cameFrom[ wid ] = cMin;
[...]

PR #2832, opened 2021-03-06
Fix and tests for #2830

phdoerfler:
Added two tests to demonstrate #2830. They will cause the build to fail.


phdoerfler:
I implemented a fix and more tests for #2830.
Some of the tests I copied over to the existing A* tests, since they do not depend on the test data. I was thinking of refactoring that, because DRY but it's just [two and not three times ;)](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).
As far as I'm concerned this PR is ready to be merged.

maxkfranz:
Looks great.  Merged

Thanks for the PR, @phdoerfler.

PR #2840, opened 2021-03-15
Update markovClustering.md

aksanoble:
Documentation incorrectly refers to the method as `markovCluster` instead of `markovClustering`


maxkfranz:
Thanks for the PR

PR #2843, opened 2021-03-19
Update fCoSE demo links

hasanbalci:
Updates existing fcose demo links and adds a new one (constraint demo)


maxkfranz:
@hasanbalci, it's great to add a new demo.  In order for it to display properly, would you be able to add a thumbnail like the other demos?

hasanbalci:
@maxkfranz Added thumbnail, thanks!

Issue #2828, opened 2021-03-04
Labels are blurred, misaligned and sometimes missing


janhartmann:
**Environment info**

- Cytoscape.js version : 3.18.0
- Browser/Node.js & version : Chrome 88.0.4324.192 (Official Build) (x86_64)

**Current (buggy) behaviour**

I have been receiving reports of the graph I am rendering is sometimes not showing the right labels at their right positions and sometimes blurred or missing completely. If the user does a refresh the graph is rendered fine again. I am unable to reproduce this issue on my own setup. Screenshot of the rendering:

![image](https://user-images.githubusercontent.com/1107383/109955417-e6751380-7cfb-11eb-988b-8d034bf7bbac.png)

They also sometimes see the edge's labels on the node body or labels on nodes where they dont belong.

I have a strong indication it can be related to https://github.com/cytoscape/cytoscape.js/issues/2290 and/or https://github.com/cytoscape/cytoscape.js/issues/2709.

**Desired behaviour**

Rendering the graph without misalignment, blurred or missing labels.

**Minimum steps to reproduce**

Unfortunately not able to reproduce myself, however I will update if I am able to.

janhartmann:
You mean, instead of setting:

```js
 if( eleCache && eleCache.invalidated ){ 
   eleCache.invalidated = false; 
   eleCache.texture.invalidatedWidth -= eleCache.width; 
 } 
```

We change this to

```js
 if( eleCache && eleCache.invalidated ){ 
   eleCache = null;
 } 
```

Or did I misunderstand?

Alexandr23:
Hi! I am suffering from almost the same problem.

Sometime labels are not being rendered properly. It can be a missing label, partially rendered label or other texts.

I tried to delve deeper into the texture caching mechanism, but have not yet found the cause of the problem. It looks as if the cache is taken with the correct dimensions, since the lengths of the labels are the same as expected, but coordinates or texture or  level is wrong.

The problem has never been reproduced locally, but it often appears among our users.

Can you tell me in which direction it is better to go deeper?

Alexandr23:
Hi @maxkfranz @janhartmann, 
I managed to create an environment where the issue is predictably reproducible. After a few days of investigation I found a problem with retired textures.

[...]

PR #2845, opened 2021-03-29
Issue-2828: Fixed the issue with retired textures

Alexandr23:
https://github.com/cytoscape/cytoscape.js/issues/2828

### DIFF
--- a/src/util/index.js
+++ b/src/util/index.js
@@ -121,12 +121,12 @@ export const defaults = defaults => {
   };
 };
 
-export const removeFromArray = ( arr, ele, manyCopies ) => {
-  for( let i = arr.length; i >= 0; i-- ){
+export const removeFromArray = ( arr, ele, oneCopy ) => {
+  for( let i = arr.length - 1; i >= 0; i-- ){
     if( arr[i] === ele ){
       arr.splice( i, 1 );
 
-      if( !manyCopies ){ break; }
+      if( oneCopy ){ break; }
     }
   }
 };


maxkfranz:
@Alexandr23, what do you think about making the PR stronger by changing the `removeFromArray()` function use `true` by default?  That would make it so the optimisation is opt-in rather than opt-out, and it may help with similar issues in future.

Alexandr23:
> Another point to consider is that the remove function seems to start at `array.length`, whereas it should start at `length - 1`. That may be worth addressing at the same time.

Yeah, good catch. I will do this.

Alexandr23:
> @Alexandr23, what do you think about making the PR stronger by changing the `removeFromArray()` function use `true` by default? That would make it so the optimisation is opt-in rather than opt-out, and it may help with similar issues in future.

Yes, I can do this. My only concern is that `removeFromArray` is used in 9 places and I'm only familiar with the caching mechanism. Should we put explicit `false` for the other 6 cases for now?

maxkfranz:
Ultimately, each call to `removeArray()` is going to be O(n) whether it can break early or not.  I think it's fine for all the calls to be more conservative for now, and they can have the flag switched later if there's a large benefit.

Alexandr23:
> Ultimately, each call to `removeArray()` is going to be O(n) whether it can break early or not. I think it's fine for all the calls to be more conservative for now, and they can have the flag switched later if there's a large benefit.

Ok, I updated the PR

Alexandr23:
Hi @maxkfranz, should we improve anything else in this PR?

maxkfranz:
Looks good.  I’ll publish a release later today

PR #2850, opened 2021-04-04
Update fCoSE extension definition

hasanbalci:
Emphasising the constraint support of fCoSE in extension definition may be a good idea to attract users. I forgot to add this to the previous PR, sorry!


maxkfranz:
Thanks 👍

PR #2857, opened 2021-04-21
feat(BaseRenderer): add dragpan, pinchzoom, and scrollzoom events

### LINKED ISSUE #2756: Distinguish between user and layout zoom/pan

I would like to distinguish between a layout zoom/pan and a user zoom/pan. 

As a default behaviour in my application, I want to fit everything to the screen using the layout `fit: true`. 

However, if the user wants to explore the graph closer, I want to change the layout options to `fit: false`, in order for the user to be in control. The reason for this is that my layout runs everytime new data is coming in to the graph and when the user is in control the graph should not fit again.

This switch should only happen on the user initiated zoom/pan, however with the current event handlers for `zoom` and `pan`, I cannot distinguish whether it was the user or the layout.

charles-stuart-appdetex:
Fixes #2756.

### DIFF
--- a/documentation/md/events.md
+++ b/documentation/md/events.md
@@ -102,6 +102,9 @@ These events are custom to Cytoscape.js, and they occur on the core.
  * `destroy` : when the instance of Cytoscape.js was explicitly destroyed by calling `.destroy()`.
  * `render` : when the viewport is (re)rendered
  * `pan` : when the viewport is panned
+ * `dragpan` : when the viewport is panned via dragging
  * `zoom` : when the viewport is zoomed
+ * `pinchzoom` : when the viewport is zoomed via pinch gesture
+ * `scrollzoom` : when the viewport is zoomed via the scrollwheel
  * `viewport` : when the viewport is changed (i.e. from a `pan`, a `zoom`, or from both when zooming about a point -- e.g. pinch-to-zoom)
  * `resize` : when the viewport is resized (usually by calling `cy.resize()`, a `window` resize, or toggling a class on the Cytoscape.js div)
--- a/src/extensions/renderer/base/load-listeners.js
+++ b/src/extensions/renderer/base/load-listeners.js
@@ -698,6 +698,7 @@ BRp.load = function(){
 
         }
 
+        cy.emit('dragpan');
         cy.panBy( deltaP );
 
         r.hoverData.dragged = true;
@@ -1084,6 +1085,7 @@ BRp.load = function(){
         newZoom = r.gestureStartZoom * e.scale;
       }
 
+      cy.emit('scrollzoom');
       cy.zoom( {
         level: newZoom,
         renderedPosition: { x: rpos[0], y: rpos[1] }
@@ -1602,6 +1604,7 @@ BRp.load = function(){
           }
         }
 
+        cy.emit('pinchzoom');
         cy.viewport( {
           zoom: zoom2,
           pan: pan2,
@@ -1740,6 +1743,8 @@ BRp.load = function(){
             r.data.bgActivePosistion = math.array2point( r.touchData.startPosition );
           }
 
+          cy.emit('dragpan');
+          
           if( r.swipePanning ){
             cy.panBy( {
               x: disp[0] * zoom,


src/extensions/renderer/base/load-listeners.js
```diff
@@ -1740,6 +1743,8 @@ BRp.load = function(){
             r.data.bgActivePosistion = math.array2point( r.touchData.startPosition );
           }
 
+          cy.emit('dragpan');
```
charles-stuart-appdetex, line 1746: I'm not sure if this emit call is necessary. Based on the code comments, etc. it looked like it 'dragpan' should be emitted here, but I wasn't able to see it fire in my testing.

maxkfranz:
Looks like a great addition.  I'll review & try it out tomorrow

maxkfranz:
Thanks for the PR!

I've reviewed this over the weekend, and there are two things we should consider:

(1) In the browser, wheel and pinch gestures are conflated on desktop.  Chrome, for example, takes Mac trackpad pinches and reports these as wheel events.  Safari has differentiated pinch events via `gesturechanged`, though these also go through our wheel handler function.  In the wheel handler, Safari pinches should be reported as `pinchzoom`.  In the docs, we should mention that pinchzoom will be reported only for browsers like Safari.

(2) The `pinchzoom` event should be fired after the zoom level is changed.  Otherwise, the consumer of the lib can't determine what the current zoom level is in a `pinchzoom` handler.  A similar approach should be used for `scrollzoom`.

charles-stuart-appdetex:
Thank you for the review. Let me know if you need any additional changes.

documentation/md/events.md
```diff
@@ -102,6 +102,9 @@ These events are custom to Cytoscape.js, and they occur on the core.
  * `destroy` : when the instance of Cytoscape.js was explicitly destroyed by calling `.destroy()`.
  * `render` : when the viewport is (re)rendered
  * `pan` : when the viewport is panned
+ * `dragpan` : when the viewport is panned via dragging
  * `zoom` : when the viewport is zoomed
+ * `pinchzoom` : when the viewport is zoomed via pinch gesture. This event is only supported for browsers which support the [gesturechange event](https://developer.mozilla.org/en-US/docs/Web/API/Element/gesturechange_event). Other browsers will fire the `scrollzoom` event.
```
maxkfranz, line 107: Or touch events: https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent

src/extensions/renderer/base/load-listeners.js
```diff
@@ -698,6 +698,7 @@ BRp.load = function(){
 
         }
 
+        cy.emit('dragpan');
```
maxkfranz, line 701: Should be after delta is applied (i.e. so can read value in handler)

maxkfranz (changes requested):
Looks good.  Some minor suggestions.  Once addressed, I'll test out the changes on a couple of devices.

src/extensions/renderer/base/load-listeners.js
charles-stuart-appdetex: It's been moved.

documentation/md/events.md
charles-stuart-appdetex: It's been updated, let me know if it needs further changes.

charles-stuart-appdetex:
> Looks good. Some minor suggestions. Once addressed, I'll test out the changes on a couple of devices.

Thanks!

src/extensions/renderer/base/load-listeners.js
```diff
@@ -1763,6 +1764,7 @@ BRp.load = function(){
             }
           }
 
+          cy.emit('dragpan');
```
maxkfranz, line 1767: On touch, input is imprecise.  The raw touch events that come in can have a delta over time, even when the user didn't intend to drag.  For this reason, you can't kick off gestures until you're reasonably sure that the user intended to perform the gesture.  Here, a threshold delta is used.

In order to make sure that `dragpan` is not erroneously emitted before the threshold is met, `dragpan` should be emitted only directly after each corresponding `panBy()` call.

maxkfranz (changes requested):
I tested it out, and things look good overall.  There's a subtle issue that should be resolved before merging.  Once that's addressed, I think it's ready for release.

src/extensions/renderer/base/load-listeners.js
charles-stuart-appdetex: Thank you for looking at it on touch devices. I've made the requested changes.

PR #2878, opened 2021-06-03
Added edge centrality to betweenness centrality

Sebubu:
**Issue type**
- Feature request

Adds the edge betweenness centrality to the betweenness centrality algorithm. Implemented according to the paper referenced in the code.

I needed to calculate edge betweenness centrality for a use-case. Edge centrality comes basically for free when doing node betweenness centrality.

It might be helpful for others as well.

### DIFF
--- a/documentation/docmaker.json
+++ b/documentation/docmaker.json
@@ -3332,12 +3332,12 @@
             {
               "name": "eles.betweennessCentrality",
               "pureAliases": ["eles.bc"],
-              "descr": "Considering only the elements in the calling collection, calculate the [betweenness centrality](https://en.wikipedia.org/wiki/Betweenness_centrality) of the nodes.",
+              "descr": "Considering only the elements in the calling collection, calculate the [betweenness centrality](https://en.wikipedia.org/wiki/Betweenness_centrality) of the nodes and edges.",
               "formats": [
                 {
                   "args": [
                     { "name": "options", "fields": [
-                      { "name": "weight: function(edge)", "descr": "A function that returns the positive weight for the edge.  The weight indicates the importance of the edge, with a high value representing high importance.", "optional": true },
+                      { "name": "weight: function(edge)", "descr": "A function that returns the positive weight for the edge.  The weight indicates the importance of the edge, with a low value representing high importance.", "optional": true },
                       { "name": "directed", "descr": "A boolean indicating whether the algorithm operates on edges in a directed manner from source to target (`true`) or whether the algorithm operates in an undirected manner (`false`, default).", "optional": true }
                     ] }
                   ]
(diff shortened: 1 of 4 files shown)


maxkfranz:
Great.

Could you update the docs with your changes?

Files:
- https://github.com/cytoscape/cytoscape.js/blob/unstable/documentation/docmaker.json#L3333
- https://github.com/cytoscape/cytoscape.js/blob/unstable/documentation/md/collection/betweennessCentrality.md

Are there some example results (e.g. from the paper) we could use as unit tests?  It would be good to have at least one.

Sebubu:
- I added a test with a simple example that calculated by hand to check.
- I added the necessary documentation.

**Important!**:
According to the [referenced paper](https://kops.uni-konstanz.de/bitstream/handle/123456789/5940/variants.pdf?sequence=1&isAllowed=y) last section on page 16, low weights are seen as more important than high weights.
I updated the documentation to reflect this. It was wrong before. Some users might have gotten the wrong results.

Would be great to see this in cyptoscape soon :)

maxkfranz:
> * I added a test with a simple example that calculated by hand to check.
> * I added the necessary documentation.
> 

Great.  It looks like there may be a problem with the test, as it's not passing on Travis.

> **Important!**:
> According to the [referenced paper](https://kops.uni-konstanz.de/bitstream/handle/123456789/5940/variants.pdf?sequence=1&isAllowed=y) last section on page 16, low weights are seen as more important than high weights.
> I updated the documentation to reflect this. It was wrong before. Some users might have gotten the wrong results.

The general description for betweenness centrality is:  The betweenness centrality for each node is the number of these shortest paths that pass through the node.  Therefore, the more central the node, the higher the centrality measure would be.

Are you sure you aren't looking at a particular use case mentioned in the paper, rather than the general one?

> 
> Would be great to see this in cyptoscape soon :)

Agreed

Sebubu:
Yep, I have to take a deeper dive into the test. I has something to do with initialising edges with 0.

> The general description for betweenness centrality is: For every pair of nodes in a connected graph, there exists at least one shortest path between the nodes such that either the number of edges that the path passes through (for unweighted graphs) or the sum of the weights of the edges (for weighted graphs) is minimised. The betweenness centrality for each node is the number of these shortest paths that pass through the node. Therefore, the more central the node, the higher the centrality measure would be.

I figured this out with my new test here
https://github.com/cytoscape/cytoscape.js/pull/2878/files#diff-f663b253c74b0337f8fae3768695213022f45831c621cc35161b2371c33f04d6R984

The algorithm gives edges with weight 1 more centrality than edges with weight 4.
The line in your quote `the sum of the weights of the edges (for weighted graphs) is minimised.` says that the sum of weights is minimised, not maximised. This indicates lower weights are more important than higher weights.

Just tell me if I confuse things here.

maxkfranz:
Yes, that's a typo.  The centrality should be higher the more central a node is.  The more paths it takes part in, the higher its total score should be (since it's just a sum of the paths).

If the concept were similarly applied to edges, the same should hold true.  A higher score means the edge is in more paths.

PR #2892, opened 2021-06-28
Add edge-style property

### LINKED ISSUE #2883: Triangle edge style

Draw edge segments as triangles between edge points, instead of lines.

Add new `edge-style` property, with possible values `line` (default) and `triangle` (ignores `line-style` and `curve-style`, it's always solid and straight).

Additional property might be needed for triangle direction. `width` can be used for triangle width at the base.

More prominently visualize the direction of the edge.

zakjan:
Adds a new `edge-style` property, may be `line` (default), or `triangle`.

Screen Shot 2021-06-28 at 9 47 56

Closes #2883

### DIFF
--- a/documentation/md/style.md
+++ b/documentation/md/style.md
@@ -424,7 +424,9 @@ A straight edge (`curve-style: straight`) is drawn as a single straight line fro
 
 For straight triangle edges (`curve-style: straight-triangle`, [demo](demos/edge-types)):
 
-A straight triangle edge (`curve-style: straight-triangle`) is drawn as a single straight triangle from the outside of the source node shape to the outside of the target node shape.  `width` defines triangle width at the source. Arrows are not supported on triangle edges.
+A straight triangle edge (`curve-style: straight-triangle`) is drawn as a single straight isosceles triangle in the direction from the source to the target.
+
+Property `width` defines line width at the triangle base, line width is `0` at the triangle apex.  Properties `line-style`, `line-cap`, `line-dash-pattern`, `line-dash-offset` are not supported.
 
 
 ## Taxi edges
--- a/documentation/md/style.md
+++ b/documentation/md/style.md
@@ -424,9 +424,9 @@ A straight edge (`curve-style: straight`) is drawn as a single straight line fro
 
 For straight triangle edges (`curve-style: straight-triangle`, [demo](demos/edge-types)):
 
-A straight triangle edge (`curve-style: straight-triangle`) is drawn as a single straight isosceles triangle in the direction from the source to the target.
+A straight triangle edge (`curve-style: straight-triangle`) is drawn as a single straight isosceles triangle in the direction from the source to the target, with a triangle base at the source and a triangle apex (a point) at the target.
 
-Property `width` defines line width at the triangle base, line width is `0` at the triangle apex.  Properties `line-style`, `line-cap`, `line-dash-pattern`, `line-dash-offset` are not supported.
+Property `width` defines width of the triangle base.  Properties `line-style`, `line-cap`, `line-dash-pattern`, `line-dash-offset` are not supported.
 
 
 ## Taxi edges
--- a/documentation/md/style.md
+++ b/documentation/md/style.md
@@ -336,7 +336,7 @@ The following is an example of valid background image styling using JSON. The ex
 These properties affect the styling of an edge's line:
 
  * **`width`** : The width of an edge's line.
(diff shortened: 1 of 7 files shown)


zakjan:
This is a PR

maxkfranz:
Looks good.  Since this is conceptually the same as `curve-style`, could this be amended to use the existing property?  It would be simpler for library consumers and less error-prone in the code in the long term.

zakjan:
Initially I thought so, but it's incompatible with most of line styles. Is it ok?

maxkfranz:
Isn't it incompatible with other properties in any case (e.g. control points)?  My line of thinking is that one property where the value seems to fit sensibly is simpler than two.  In any case, incompatibilities should be noted where they wouldn't be obvious (e.g. the edge offset distance from the node probably works, but some others may not in subtle ways).

If a property is already listed inclusively (e.g. "the following properties only work for `curve-style: foo | bar`), then those parts are fine as-is.  In cases where it may be ambiguous, we should try to make things clear for users.

zakjan:
Updated, it's `curve-style: straight-triangle` now

maxkfranz:
Great.  I've tested out your changes a bit.  Here are some questions:

(1) I added an `ef` edge in the debug demo to try it out.  Could you add an edge like that to your PR so that it could be more easily tested in future?

```
.selector('#ef')
        .style({ // e.g.
          'curve-style': 'straight-triangle',
          'width': 6

          // then you can play around with different styles to make sure they work in combination, e.g.:
          'source-endpoint': 'outside-to-node-or-label',
          'source-distance-from-node': 10,
        })
```

(2) The arrows seem to be shown even though you say in the documentation that they're unsupported.  Should the documentation be changed, or should the arrows be disabled in the code?

(3) The documentation doesn't explicitly say which direction the triangle goes, i.e. which is the small, pointed side (target) and which is the big, flat side (source).  You hint about the direction a bit, but it could be unclear what it means precisely to a novice of the library.  Could this be amended?

(4) Which [ edge endpoint](https://js.cytoscape.org/#style/edge-endpoints) properties are compatible with the new edge type?  If it's intended to support all of them, have you tested those cases?  Everything must be compatible with existing properties, or where not possible the documentation should explicitly note incompatibilities.

(5) Are there any [line](https://js.cytoscape.org/#style/edge-line) properties that are incompatible (e.g. properties that make sense to be compatible like `line-cap` should be tested and confirmed working)?  Which combinations have you tested?  Please note incompatibilities in the docs.

zakjan:
Thanks for highlighting possible conflicts with other existing features.

1) There is a testing edge in `edge-types` demo. Is it good enough, or do you mean another demo?
2) You're right, arrows work well. I removed the note about arrows not supported.
3) Updated docs.
4) Tested all edge endpoint properties. All values work well.
5) Tested all line properties. Solid fills including gradients work well. `line-style`, `line-cap`, `line-dash-pattern`, `line-dash-offset` are not supported. Updated docs.

There are no changes in the code, only in the docs. Does it cover all your concerns?

documentation/md/style.md
```diff
@@ -420,6 +420,15 @@ For straight line edges (`curve-style: straight`, [demo](demos/edge-types)):
 A straight edge (`curve-style: straight`) is drawn as a single straight line from the outside of the source node shape to the outside of the target node shape.  Endpoint and midpoint arrows are supported on straight edges.  Straight edges are not generally suitable for multigraphs.
 
 
+## Straight triangle edges
+
+For straight triangle edges (`curve-style: straight-triangle`, [demo](demos/edge-types)):
+
+A straight triangle edge (`curve-style: straight-triangle`) is drawn as a single straight isosceles triangle in the direction from the source to the target.
```
maxkfranz, line 427: Which side is source and which is target?  This should be described explicitly (pointy side vs. flat side, or small side versus large side, etc.)

documentation/md/style.md
```diff
@@ -420,6 +420,15 @@ For straight line edges (`curve-style: straight`, [demo](demos/edge-types)):
...
+
+Property `width` defines line width at the triangle base, line width is `0` at the triangle apex.  Properties `line-style`, `line-cap`, `line-dash-pattern`, `line-dash-offset` are not supported.
```
maxkfranz, line 429: > line width is `0` at the triangle apex

It's probably more intuitive to describe this as a point

maxkfranz (commented):
> There is a testing edge in edge-types demo. Is it good enough, or do you mean another demo?

All edge types should be included in the debug demo, as the debug demo is used by mostly everyone when making contributions.  This makes it very likely that someone working on the project would notice if there were any conflicts with a particular edge type and their work.

> You're right, arrows work well. I removed the note about arrows not supported.

Great

> Updated docs.

See comments

> Tested all edge endpoint properties. All values work well.

OK

> Tested all line properties. Solid fills including gradients work well. line-style, line-cap, line-dash-pattern, line-dash-offset are not supported. Updated docs.

What if a library consumer is looking at the individual line properties rather than the triangle curve style section?  Should these conflicts also be mentioned alongside those properties themselves?

documentation/md/style.md
zakjan: Updated, moved to the first paragraph

maxkfranz:
> Re demo, I agree with you, but as I mentioned, the new edge type is already present in the demo. This is the screenshot. Is this good enough? If not, I'm afraid that I'm still not sure what do you mean.
> 

Take a look here: https://github.com/cytoscape/cytoscape.js/tree/unstable/debug

zakjan:
Oh, thanks, I didn't know about this separate demo. Added.

PR #2928, opened 2021-10-01
Add underlay node/edge styles

### LINKED ISSUE #2882: Underlay node style

Draw a halo below the node. Rendering would be exactly the same as overlay, but below the node instead of above.

Add new `halo-*` properties similar to `overlay-*` properties: `halo-padding`, `halo-color`, `halo-opacity`, `halo-shape` (#2881).

Another way to highlight a node without overriding its style, e.g. after it is found by user using a search control.

zakjan:
Adds new `underlay-padding`, `underlay-color`, `underlay-opacity`, `underlay-shape` node/edge styles.

All existing `overlay-*` styles are duplicated into `underlay-*`, and underlay rendering is added before rendering the main shape.

127 0 0 1_8080_documentation_demos_node-types_

127 0 0 1_8080_documentation_demos_edge-types_

Closes #2882


src/extensions/renderer/canvas/drawing-nodes.js
```diff
@@ -342,6 +356,40 @@ CRp.drawNodeOverlay = function( context, node, pos, nodeWidth, nodeHeight ){
   }
 };
 
+CRp.drawNodeUnderlay = function( context, node, pos, nodeWidth, nodeHeight ){
```
maxkfranz, line 359: Could these underlay functions simply be made by making the overlay function generic?  E.g. a `prefix` parameter could be sent to the function so either `'overlay'` or `'underlay'` properties are used.  That way there's much less code

documentation/md/style.md
```diff
@@ -641,6 +641,16 @@ These properties allow for the creation of overlays on top of nodes or edges, an
...
 
+
+## Underlay
+
+These properties allow for the creation of underlays behind nodes or edges, and are often used in a highlighted state.
+
+ * **`underlay-color`** : The colour of the underlay.
+ * **`underlay-padding`** : The area outside of the element within which the underlay is shown.
+ * **`underlay-opacity`** : The opacity of the underlay.
+ * **`underlay-shape`** : The shape of the underlay; may be `round-rectangle` (default), `ellipse`.
```
maxkfranz, line 652: For overlay-shape and underlay-shape, we should note to the reader that they are only for nodes.  The edges don't get to use the shapes, right?

maxkfranz (commented):
Looks good.  Minor suggestions re. docs & refactoring

zakjan:
Yes, thanks, both suggestions make sense, I'll apply them this week.

documentation/md/style.md
zakjan: Updated

maxkfranz:
Excellent.  I'm away this upcoming week, but I'll push out the release.  It will be great to see all [these contributions](https://github.com/cytoscape/cytoscape.js/milestone/221?closed=1) in 3.20.0

PR #2947, opened 2021-11-12
Add functions to convert bend(anchor) points to relative positions

hasanbalci:
Add `segmentsToRelativePositions` and `controlsToRelativePositions` functions to core (cytoscape/cytoscape.js-elk#38)

The below two functions convert given segment/control points in absolute coordinates to segment/control distances and weights:

```
cy.segmentsToRelativePositions( segmentPoints, sourcePoint, targetPoint )
cy.controlsToRelativePositions( controlPoints, sourcePoint, targetPoint )
```
and returns:
`{distances: [distance1, distance2, ...], weights: [weight1, weight2, ...]}`

Currently a user can set relative segment/control point distances and weight via style properties such as `segment-distances` and `segment-weights` etc. and then can reach absolute coordinates of these points via `edge.segmentPoints()` and `edge.controlPoints()` functions. However, conversion from absolute coordinates to relative positions is currently not possible. With the help of these functions, user who has absolute coordinates can achieve this conversion and assign relative coordinates to corresponding edge styles.

### DIFF
--- a/documentation/docmaker.json
+++ b/documentation/docmaker.json
@@ -950,6 +950,58 @@
         },
 
         {
+          "name": "Edge points",
+          "fns": [
+            {
+              "name": "cy.segmentsToRelativePositions",
+              "descr": "Convert segment points in absolute positions into segment distances and weights and return them.",
+              "formats": [
+                {
+                  "args": [
+                    {
+                      "name": "segmentPoints",
+                      "descr": "The segment points to be converted"
+                    },
+                    {
+                      "name": "sourcePoint",
+                      "descr": "The source point of the edge."
+                    },
+                    {
+                      "name": "targetPoint",
+                      "descr": "The target point of the edge."
+                    }                                       
+                  ]
+                }
+              ],
+              "md": "core/segmentsToRelativePositions"
+            },
+            {
+              "name": "cy.controlsToRelativePositions",
+              "descr": "Convert control points in absolute positions into control point distances and weights and return them.",
+              "formats": [
+                {
+                  "args": [
+                    {
+                      "name": "controlPoints",
+                      "descr": "The control points to be converted"
+                    },
+                    {
+                      "name": "sourcePoint",
+                      "descr": "The source point of the edge."
+                    },
+                    {
+                      "name": "targetPoint",
+                      "descr": "The target point of the edge."
+                    }                                       
+                  ]
+                }
+              ],
+              "md": "core/controlsToRelativePositions"
+            }            
+          ]
+        },        
+
+        {
           "name": "Style",
 
           "fns": [
--- a/documentation/md/core/controlsToRelativePositions.md
+++ b/documentation/md/core/controlsToRelativePositions.md
@@ -1,0 +1,18 @@
+## Details
+
+This function converts given `controlPoints` for an edge with `sourcePoint` and `targetPoint` into corresponding `control-point-distances` and `control-point-weights` and returns them.
+
+`sourcePoint` and `targetPoint` can be either the source and target node positions as in the case of `edge-distances: 'node-position'` or the ends of the line from source to target which is from the outside of the source node’s shape to the outside of the target node’s shape as in the case of `edge-distances: 'intersection'`.
+
+## Examples
+
+```js
+var je = cy.$('#je');
+var controlPoints = je.controlPoints();
+var jPos = je.source().position();
+var ePos = je.target().position();
+
+var result = cy.controlsToRelativePositions(controlPoints, jPos, ePos);
+var controlPointDistances = result.distances;
+var controlPointWeights = result.weights;
+```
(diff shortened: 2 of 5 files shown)


hasanbalci:
Hi @maxkfranz, did you have a chance to review this PR?

maxkfranz:
I get a lot in my inbox, so sometimes there's a bit of a delay.

I think that the functions you have now are great, and we could add a couple of new functions on top of them to make things more streamlined.  It would be useful if we could make it so that it's easy to add support for custom points to other layouts, and it would also be useful to be able to enforce that there is consistency in how layouts use `scratch` for these points.

We currently have a [`nodes.layoutPositions()`](https://js.cytoscape.org/#nodes.layoutPositions) function.  Could we have functions similar to that one, using your existing functions for the underlying implementation? 

I propose the following:

- `edges.layoutSegmentPoints(layout, options, function(ele, i))`
- `edges.layoutBezierPoints(layout, options, function(ele, i))`

1. The passed in function returns the points for `ele` in absolute layout coordinates so it's easy for the layout to specify them.
2. The points are converted using your functions to the appropriate relative style property values.
3. The relative style value can be stored in scratch automatically since you have the layout name in `options.name`.  So you can automatically store the points in `scratch` based on the layout name (e.g. `scratch(options.name) = { segmentDistances: [], segmentWeights: [], ... })`, and the layout doesn't need to do any extra work.  It's automatic and always consistent across layouts.
4. With the higher level functions, we probably don't need the existing `segmentsToRelativePositions` and `controlsToRelativePositions` ones to be publicly available in the docs etc.
5. This should basically just be repurposing what you already have in this PR and in pieces from your Elk PR, so I hope it shouldn't be too much effort overall

What do you think?

hasanbalci:
Hi @maxkfranz,

I think, I've got the idea. The functionality of the `layoutSegmentPoints` and `layoutBezierPoints` functions will be to make the conversion and storing the corresponding style values in edges' scratch. This way seems more meaningful.

My only concern is that during the conversion inside these functions, I'll need the latest edge end points (they can be edge source and target node positions assuming the `edge-distances` will be set to `node-position`). However, they may not be updated till the end of the layout, so I cannot use current API functions for that purpose. That's why I wanted this information as parameters in this PR. Should I assume that this information will be returned via the passed in function, as well? What is your opinion on this?

Once I make this clear in my mind, I will update the PR as you suggested. Thanks!

maxkfranz:
Don’t worry about that too much to start.  We can get it to a production-ready, polished API in steps.

The first step is to build a layer on top of your existing functions.  This layer has the `edges` functions, e.g. `edges.layoutBezierPoints()`.  To start, these functions can just apply the positions to `scratch` directly when called.  That has the downside that you mentioned, where the order in which the functions are called is important.

The next step would be to automate when this application to `scratch` is carried out.  Because we would have functions like `nodes.layoutPositions(layout, options, function)`, we have access to all the layout options and the layout object itself.  So we can make it ‘magic’ by automating when the application is made using `layout`.

Layouts, like everything else in Cytoscape, are event emitters.  So we can internally use those events to make sure that the edge scratch points are applied only after the next `layoutstop` event for the passed in `layout`.  We also have access to all the internal state in the core library, so we can use some tricks.  For instance, you can see that when you register a layout with Cytoscape, it doesn’t use your layout class directly.  Instead, it creates its own layout class to wrap the one you pass in, so the layout is always cleaned up:

https://github.com/cytoscape/cytoscape.js/blob/ca717b56c690123211cb09b1aa2af48e5743067c/src/extension.js#L37-L136

So with these tools, we could do something like this:

- Track all calls to the edge point functions once the layout starts to `run()`.
- The edge point calls aren't actually applied when the author calls those functions, but they are saved for later.
- The edge point calls are used inside `nodes.layoutPositions()`, where you do know the final positions.

Some points to consider:

- There is the possibility that the layout author calls the edge functions after `nodes.layoutPositions()`.  That's a bit unnatural, but somebody could still do it.  We could detect that case and print a warning so the author knows to fix it.  We could still apply the points, since the nodes have their layout positions already, but it may not work exactly as expected in all cases (e.g. if the layout were animated, then the edge points would be applied only at the end making it look like they jumped).
- We could also consider a fallback for layouts that don’t use `nodes.layoutPositions()`, where the edge point functions are called after `layoutstop`.  This may not be that important, since I think pretty well all the layouts should be using `layoutPositions()` anyway.  An argument could be made that if you’re doing all the layout song-and-dance manually (without `nodes.layoutPositions()`), then it’s probably fine to also have to do the edge point stuff manually too.  So it may be that we don't have to make special accommodations for this case.

hasanbalci:
Thanks for your detailed answer and tips @maxkfranz. It seemed a bit complicated at first glance, but as you said, I will go step-by-step by following the explanations in your comment. I'll start to it as soon as possible. I may need your guidance again at the points where I am stuck.

hasanbalci:
Sorry for the delay, I'll complete this at the first opportunity.

Issue #2945, opened 2021-11-05
concentric layout: controlling node level from ipycytoscape


shadiakiki1986:
**Current (buggy) behaviour**

concentric layout: node levels are controlled from a javascript function through argument named "concentric"
(docs at: https://js.cytoscape.org/#layouts/concentric )

**Desired behaviour**

This default function would check for a field in node `data` first, in case it is set from ipycytoscape, before falling back to using node degrees. `ipycytoscape` cannot send a javascript function to the front-end since `ipywidgets/traitlets` json-serializes all data sent to the front-end.

**Minimum steps to reproduce**

Check issue in ipycytoscape: https://github.com/cytoscape/ipycytoscape/issues/238#issuecomment-962251735

maxkfranz:
This sounds like an ipycytoscape issue.  Have you tried posting your issue there?

shadiakiki1986:
Yes but it's not possible to fix it there because AFAIK ipywidgets doesn't
support sending a javascript function to the front-end.
OTOH it supports sending a string. So the ipycytoscape issue would be fixed
if cytoscape concentric's default function checked a key "concentric" in
the data field before falling back to the degree function.

maxkfranz:
Sounds like ipycytoscape should be converting this potential string option to a function on their JS/TS side.  Have you posted your issue with ipycytoscape?

shadiakiki1986:
It could be fixed on ipycytoscape's side in a complex way, or it could be fixed here in cytoscape.js with 2 lines of code.
The above PR #2951 does the job.

joshwatt22:
Was this issue ever fixed? I am facing the same problems.

PR #2951, opened 2021-11-16
Concentric layout: check data fields before falling back to node degree

shadiakiki1986:
**Issue type**

Feature request

[Concentric layout](https://js.cytoscape.org/#layouts/concentric) would allow the user by default to set the node level from a data field `concentricLevelValue` before falling back to the node degree. The same is for the `levelWidth` value and the `concentricLevelWidth` field

Documented in #2945

### DIFF
--- a/src/extensions/layout/concentric.js
+++ b/src/extensions/layout/concentric.js
@@ -16,7 +16,7 @@ let defaults = {
   width: undefined, // width of layout area (overrides container width)
   spacingFactor: undefined, // Applies a multiplicative factor (>0) to expand or compress the overall area that the nodes take up
   concentric: function( node ){ // returns numeric value for each node, placing higher nodes in levels towards the centre
-    return node.degree();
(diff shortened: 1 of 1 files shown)


maxkfranz:
It looks like it would be best for this sort of thing to be added to ipycytoscape, for one, because this is a ipycytoscape-specific issue.  Direct users of Cytoscape can simply use a function.

While it seems simple to add options like these to the main library, it more often than not does not work out well in the long term -- there's usually hidden complexity and expense.  For instance, as this PR stands, the method would fail for values of zero.

There's a great article by Evan Martin on this topic: http://neugierig.org/software/blog/2018/07/options.html

I'd like to see your feature made available, and the best place to do that would be in ipycytoscape.  Let's move forward with that, and I'll make myself available to advise to help move that forward.

Thanks for your feedback, and thanks for using Cytoscape

Issue #2948, opened 2021-11-12
Style selector on array field with index


zirkelc:
I'm retrieving my nodes and edges from the graph database Neo4j and this supports multiple labels per node. That means my data contains a field `labels`. I would like to apply certain styles based on a selector for the first label

```
{
  selector: `node[labels.0 = '${label}']`,
  style: { }
} 
```

This actually does seem to work, but I still get a warning in the console:
![image](https://user-images.githubusercontent.com/950244/141487879-e757739e-2e39-4c49-b713-ab0743c40238.png)

I'm also using the notation `labels.0` to read data from the node and it works fine.

**Environment info**

- Cytoscape.js version :    
"cytoscape": "^3.18.0",
"react-cytoscapejs": "^1.2.1",
- Browser/Node.js & version :
 Chrome: Version 95.0.4638.54 (Official Build) (x86_64)

maxkfranz:
I don’t recall any array support, and I don’t see it mentioned in the docs.  It looks like you may be relying on undefined behaviour

You’re welcome to put up a pull request if you’d like explicit selector support for array indices.  That should include unit tests and a consideration of different type scenarios (0 v.s. “0”, array v.s. object, etc.)

See https://github.com/cytoscape/cytoscape.js/blob/845be4bd9b9a738719960bf6541baf9fa0e26ea2/test/selectors.js <https://github.com/cytoscape/cytoscape.js/blob/845be4bd9b9a738719960bf6541baf9fa0e26ea2/test/selectors.js>

zirkelc:
Okay, I will try to come up with a PR in the next few days!

PR #2952, opened 2021-11-18
allow paths for get and set data field

zirkelc:
PR related to Issue https://github.com/cytoscape/cytoscape.js/issues/2948 regarding explicit selector support for array indices.

I added support for selectors via dot-notation `field.index` because it requires only a small change to the `tokens.variable` regex and doesn't possibly interfere with the `metaCompare` expressions which uses the double brackets `[[]]` . Moreover, when using dot-notation compared to bracket-notation, we don't have to handle different  cases like `field[0]`vs `field["0"]. 

In the data implementation I used [lodash.get](https://lodash.com/docs/4.17.15#get) and [lodash.set](https://lodash.com/docs/4.17.15#set) to get/set a data field via a path. That means support for bracket-notation would theoretically be possible in the future without a big change. In the getter/setter the field name is checked if it contains a dot and is therefore considered a path. That should avoid performance costs because we use lodash get/set only for paths and use the normal assignments for all other operations.

Please let me know what you think and if I missed something!


maxkfranz:
Some initial thoughts:

(1) It may make sense to have some tests that combine this dot operator with meta characters.  The parser supports some weird variable names if you escape them.  It would be good to ensure that there aren't conflicts.

(2) `_.set()` is great in that it accepts either an array of fields or a string.  It's convenient that you can pass in the full string to Lodash.  For the setting case, that means that Lodash would repeatedly parse the same string for N elements.  Would it make sense to pull that out of the loop by building up an array that can be passed for each iteration?

zirkelc:
1) I can add some more test cases. Could you give some more details which cases would make sense?

2) That was a good hint! I moved the path extraction to the top of the getter/setter block and added lodash `_.toPath()` method to  build the path in advance and pass the path array to `_.get()` and `_.set()`. The lodash get/set methods check internally if the given path (as string) exists as a normal field on the object (like the metachar test cases) and then get/set the field directly instead of using a path array. I handled this case as well by adding a check on `path && single._private[ p.field ][ name ] === undefined`.

maxkfranz:
(1) I was thinking of some basic cases with one or more of the meta characters defined in the regex with the indices interspersed (before, after, in between)

(2) Great

zirkelc:
I added some more test cases, for example with `'weird.0': 'weird.index'`, etc. I also added a combined test case with `'arrayval.0': [0, 1]` which does not work at the moment:




```js
{ 
  data: { 
    id: 'n2', 
    foo: 'two', 
    parent: 'nparent', 
    weight: 2, 
    'weird.name3': '"blah"^blah<blah>#blah', 
    'weird.1': '"blah"^blah<blah>#blah', 
    'arrayval.0': [0, 1] 
  }, 
  classes: 'cls1' 
}

itSelects('[arrayval\\.0.0 = 0]', 'n2');
itSelects('[arrayval\\.0.1 = 0]', 'n2');    
```

That's a tricky case as we are combining an array index access with a number in the field name. Should I try to support it or somehow discourage its use?

zirkelc:
I did some more tests and came up with two conflicting problems with the implementation via `_.get()`:

- accessing a field with a multi level path: 
````
arrayval: ['index0', 'index1']
````
access via `arrayval.0.1` would result in accessing the first (=0) array element and then the second =(1) character **i** in the `index0` string. 

- accessing a field with `.0` suffix and a single level path:
```
'arrayval.0': [0, 1]
```
accessing via `arrayval.0.1` would result in no result

The most feasible solution I think is to limit the dot-operator to access objects and arrays to avoid the incorrect result from the first example, and also limit the depth to one level so that only the last portion of the identifier is considered as dot-operator. I think it's already stated somewhere that the diff algorithm of `cy.data()` does only a shallow compare and therefore nested objects should be avoided anyway. 

Let me know what you think and if I should continue with this PR.

**Comment by maxkfranz** [comment]
> I did some more tests and came up with two conflicting problems with the implementation via `_.get()`:
> 
> * accessing a field with a multi level path:
> 
> ```
> arrayval: ['index0', 'index1']
> ```
> 
> access via `arrayval.0.1` would result in accessing the first (=0) array element and then the second =(1) character **i** in the `index0` string.

In some cases, this may be desirable.  

If we're unsure whether it would be a good idea to include character indexing as an explicit feature, it could simply be undocumented for now -- without code to forbid it.  If behaviour this isn't documented, then a consumer ought not to rely on it.  If we later decide it's a good idea, then it can just be documented.

What do you think?

> 
> * accessing a field with `.0` suffix and a single level path:
> 
> ```
> 'arrayval.0': [0, 1]
> ```
> 
> accessing via `arrayval.0.1` would result in no result
> 
> The most feasible solution I think is to limit the dot-operator to access objects and arrays to avoid the incorrect result from the first example, and also limit the depth to one level so that only the last portion of the identifier is considered as dot-operator. I think it's already stated somewhere that the diff algorithm of `cy.data()` does only a shallow compare and therefore nested objects should be avoided anyway.
> 
> Let me know what you think and if I should continue with this PR.

I would hope that no one would think that something like `"blah.0"` would be a good field name in a JSON object.

I suspect there may be valid use cases for multiple dots, e.g. `foo.bar.baz` where `foo` and `bar` are objects, so it would be great if that could be supported.

Would it be sufficient to document that dots are only supported for the cases that we think make sense (objects etc.)?  Other cases like `"blah.0"` will then not be expected to work and there doesn't have to be special accommodations in the code.

**Comment by zirkelc** [comment]
HI @maxkfranz, sorry for the later response, I somehow missed notification in this.

I think it's good to only document the cases that make sense (e.g. objects) and leave other cases unmentioned so people won't get encouraged to try them. 

How should we continue regarding the two failing test cases ([see this comment](https://github.com/cytoscape/cytoscape.js/pull/2952#issuecomment-989707501)) and the documentation? I could add some documentation for this selector. Just let me know where you think it should be documented.

**Comment by zirkelc** [comment]
I added documentation in the sections 'notes & caveats' and 'data' in *selectors.md*. I also added another test case for accessing an nested object property via dot-operator: 

```js  
itSelects('[bar.baz = "baz"]', 'n2');   
```

I did comment out the failing test case for mixing meta chars and dot operator (see https://github.com/cytoscape/cytoscape.js/pull/2952#issuecomment-989707501). Maybe a negated test case method `itSelectsNot` would make sense, or we just leave it like that.

maxkfranz:
Great.  I think that what we have now is sufficient for release.  If you'd like to improve the tests even more before the release, you're welcome to go ahead with that.  More tests are almost always a good thing.

Once you're happy with things, we can move forward with a [3.21.0 release](https://github.com/cytoscape/cytoscape.js/milestone/222)

zirkelc:
I'm happy I could contribute something to this great library! Thank you for your support and efforts! :-)

maxkfranz:
Great.  Let's get this released

PR #2960, opened 2021-12-12
Master

arthabus:
**Issue type**

Feature request

Custom renderer: allows overriding of base methods

When creating a custom renderer for more advanced use cases, it's sometimes required to have an ability to override methods of the BaseRenderer class.


maxkfranz:
It looks like these methods aren't intended to be overridden.  Would you describe what you're trying to do in more detail?

PR #2965, opened 2021-12-25
Find negative weight cycles on Bellman-Ford

TexKiller:
**Issue type**

Feature request

Adds a new option to the Bellman-Ford algorithm called `findNegativeWeightCycles`. If this option is set to `true`, negative weight cycles will be identified and returned.

The `negativeWeightCycles` property in the return of the algorithm should be an array with all negative weight cycles found. Each cycle is a collection in the form `[node, edge, node, edge, ...]`, the last element being the same node as the first.

### DIFF
--- a/documentation/docmaker.json
+++ b/documentation/docmaker.json
@@ -3184,8 +3184,7 @@
                     { "name": "options", "fields": [
                       { "name": "root", "descr": "The root node (selector or collection) where the search starts." },
                       { "name": "weight: function(edge)", "descr": "A function that returns the numeric weight for the edge.  The weight indicates the cost of going from one node to another node.  The weight may be positive or negative.", "optional": true },
-                      { "name": "directed", "descr": "A boolean indicating whether the algorithm should only go along edges from source to target (default `false`).", "optional": true },
-                      { "name": "findNegativeWeightCycles", "descr": "A boolean indicating whether the algorithm should find and return negative weight cycles (default `false`).", "optional": true }
+                      { "name": "directed", "descr": "A boolean indicating whether the algorithm should only go along edges from source to target (default `false`).", "optional": true }
                     ] }
                   ]
                 }
--- a/documentation/docmaker.json
+++ b/documentation/docmaker.json
@@ -3184,7 +3184,8 @@
                     { "name": "options", "fields": [
                       { "name": "root", "descr": "The root node (selector or collection) where the search starts." },
                       { "name": "weight: function(edge)", "descr": "A function that returns the numeric weight for the edge.  The weight indicates the cost of going from one node to another node.  The weight may be positive or negative.", "optional": true },
-                      { "name": "directed", "descr": "A boolean indicating whether the algorithm should only go along edges from source to target (default `false`).", "optional": true }
+                      { "name": "directed", "descr": "A boolean indicating whether the algorithm should only go along edges from source to target (default `false`).", "optional": true },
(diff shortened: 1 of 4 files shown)


maxkfranz:
Looks good.

Would you document your new option in the `docmaker.json` file?  Alternatively, could the new behaviour be automatic?

TexKiller:
@maxkfranz Sorry for the delay, and for forgetting to add the option to `docmaker.json` before.

I think this could be made the new behavior later, but I'm a little reluctant right now because it adds some computation and might affect performance (also it's not strictly backwards-compatible because the `negativeWeightCycles` property of the return was previously an empty array).

maxkfranz:
> I think this could be made the new behavior later, but I'm a little reluctant right now because it adds some computation and might affect performance (also it's not strictly backwards-compatible because the `negativeWeightCycles` property of the return was previously an empty array).

As a rule of thumb, minimising API surface area and maximising simplicity are generally more important than maximising performance -- especially in the long term, in my opinion.  It's easier to add a new option later than it is to remove one.  For example, we could make the default true on your option field and leave it undocumented.  Then if it's important for performance later, we can simply document it to make it public (minor version release).  If it were made public and then we decided to make it automatic later, then deleting that API would require a new major version (semver).

W.r.t. the naming of `negativeWeightCycles`: We don't need to use that exact name.  It doesn't look like that was public or documented, so it's not a compatibility concern.  We could use a better name (e.g. singular if only one path is ever returned).  

What do you think?

TexKiller:
I made some changes to improve the detection and make it the default behavior, allowing the user to disable the detection by passing the option `findNegativeWeightCycles: false`, which is now undocumented. This PR is now a breaking change, but if that is preferred then it's fine by me.

I've also pushed a commit adding a sixth node to the tests and changing the test that finds negative cycles to find two cycles (therefore, the property shouldn't be singular). For context, we couldn't have two separated cycles with only 5 nodes, and to add the extra node I had to update many other tests. I understand if that is not worth it, so just let me know and I can just revert that commit.

I'm open for naming the property something else, but I believe it should be an array of negative weight cycles.

maxkfranz:
> This PR is now a breaking change, but if that is preferred then it's fine by me.

If there was no public API for this, then there's nothing to break.  If it's not public, a library consumer is not permitted to use it.

> I've also pushed a commit adding a sixth node to the tests and changing the test that finds negative cycles to find two cycles (therefore, the property shouldn't be singular). For context, we couldn't have two separated cycles with only 5 nodes, and to add the extra node I had to update many other tests. I understand if that is not worth it, so just let me know and I can revert that commit.
> 
> I'm open for naming the property something else, but I believe it should be an array of negative weight cycles.

Great, let's merge this in for the next release

TexKiller:
Thanks for merging!

What I meant when I said this was a breaking change was the return of the function suddenly changed for the same parameters. If someone is responding to a request with the return, saving it in a DB or even snapshoting it on tests then this change will have consequences. But I'm ok with that, like I said.

maxkfranz:
That's a good observation.

This library (and npm) follow semver, so we go along with that norm.  New features like this one go into minor version numbers.  There's interesting nuance in your example in that a breaking change for that example app doesn't necessarily correspond to a semver breaking change (i.e. major version number) for the library.

For those sorts of cases, it may make sense for the library consumer to stick to patch releases rather than feature releases in routine upgrades.  It seems like a limitation of the expressiveness of semver.

Thanks for the PR.  We'll have a release soon

PR #2974, opened 2022-01-19
Add depthSort option to breadthfirst layout

vitalos:
**Issue type**

Feature request

This feature adds an option called `depthSort` to available options for the `breadthfirst` layout. The value is expected to be a sorting function (in the same sense as the `sort` option of the `circle` layout). When set, this function will be used to sort nodes that are at the same depth. If left unset then the original behavior will be the default (sort by "weighted percent" and then by element ID in the case of ties).

Because of their similarity and out of a desire to keep the language consistent, I reused most of the comment that documents the aforementioned `sort` option of the `circle` layout, but I modified it slightly to make it clear that the sort function will only apply to nodes at the same depth.

My use case is as follows: I am using a graph to visualize a question tree for the user. A question is a node in the tree, and its children are follow-up questions. The follow-up question that is asked depends on the answer that is given to the parent question. Both answers and follow-up questions are expected to be listed in an order that the user specifies, so I want the nodes on the graph to be displayed in an order consistent with the order of the questions + answers. Without this feature, I have no easy way to guarantee that the nodes will be displayed in the order that I desire.


maxkfranz:
Great addition.  Before it's merged, would you update the docs with your new option?  You can use `npm run docs` to test your changes

vitalos:
I had to fix a couple things in docmaker.js to for the docs task to run successfully. It appears that the marked package was bumped up several major versions recently and there were breaking changes.

Also, I think I missed a few files in my first commit (package-lock.json, .size-snapshot.json), so I committed those as well. Let me know if that was a mistake.

maxkfranz:
> I had to fix a couple things in docmaker.js to for the docs task to run successfully. It appears that the marked package was bumped up several major versions recently and there were breaking changes.

That's unfortunate.  It's great that you got it working.  It was probably one of those GH CVE PRs, which are often not applicable but people get scared of them if they're not merged in.

It's great that you solved it!

> 
> Also, I think I missed a few files in my first commit (package-lock.json, .size-snapshot.json), so I committed those as well. Let me know if that was a mistake.

I wouldn't worry about that.  Those would all be updated during a release build anyway

Let's merge this in.  Thanks for the PR

Issue #2995, opened 2022-03-12
CSS style selectors with multiple classes are not parsed correctly.


mkelly1495:
**Environment info**

- Cytoscape.js version : 3.21.0
- Browser/Node.js & version : Chrome/14.19.0

**Current (buggy) behaviour**

When specifying a style selector such as `node.class1.class2 { ... }` in a css file, cytoscape fails to parse the selector correctly. It actually parses the selector as ['node', 'class1.class2'] instead of ['node', 'class1', 'class2']. This appears to be caused by an incorrect regular expression ```^\\.((?:[\\w-.]|(?:\\\\[\\!\\\"\\#\\$\\%\\&\\'\\(\\)\\*\\+\\,\\.\\/\\:\\;\\<\\=\\>\\?\\@\\[\\]\\^\\`\\{\\|\\}\\~]))+)``` being used to match the class. The correct expression is ```^\\.((?:[\\w-]|(?:\\\\[\\!\\\"\\#\\$\\%\\&\\'\\(\\)\\*\\+\\,\\.\\/\\:\\;\\<\\=\\>\\?\\@\\[\\]\\^\\`\\{\\|\\}\\~]))+)``` (i.e. without the `.` after the `[\\w-` part.

**Desired behaviour**
Ability to specify more than one selector class in the css (e.g. `node.class1.class2`).

**Minimum steps to reproduce**
Create a style sheet file called `my-style.css` that defines a style with more than one class as a selector:
```
node.class1.class2 {
  background-fill: solid;
  background-color: #268dae;
}
```

Create a cytoscape graph that specifies the `style: require('./my-style.css')` property in the config (e.g. `var myGraph = cytoscape({ ..., style: require('./my-style.css'), ... })`.

Observe that the `consumeExpr` function in `src/selector/parse.js` fails to correctly parse the `.class1.class2` as two separate classes. Instead in parses it as a single class and therefore fails to match the style properly when the `match[Type.CLASS]` function (in `src/selector/query-type-match.js` is invoked. The class passed into the `match[Type.CLASS]` function is `.class1.class2` and therefore never matches anything since the `node.classes` property is an array with each class as a separate element.

maxkfranz:
@mkelly1495, can you confirm whether 3.20 is affected?  Are earlier versions affected, e.g. 3.10?

@zirkelc, could this be related to #2952?  Assuming this is a recent issue, this looks like the only recent change to selectors.

mkelly1495:
[...]

PR #2997, opened 2022-03-17
Fix css class selector.

mkelly1495:
[CSS style selectors with multiple classes are not parsed correctly](https://github.com/cytoscape/cytoscape.js/issues/2995)

### DIFF
--- a/src/selector/tokens.js
+++ b/src/selector/tokens.js
@@ -18,7 +18,7 @@ const tokens = {
 };
 tokens.variable = '(?:[\\w-.]|(?:\\\\' + tokens.metaChar + '))+'; // a variable name
 tokens.value = tokens.string + '|' + tokens.number; // a value literal, either a string or number
-tokens.className = tokens.variable; // a class name (follows variable conventions)
+tokens.className = '(?:[\\w-]|(?:\\\\' + tokens.metaChar + '))+'; // a class name can be letters, numbers, or dashes.
 tokens.id = tokens.variable; // an element id (follows variable conventions)
 
 (function(){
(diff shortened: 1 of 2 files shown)


maxkfranz:
@mkelly1495, looks good.

I moved the base branch to `unstable`, because it has GitHub Actions tests.  We've had some trouble with the reliability of Travis and currently only the `unstable` branch has GHA setup.  GHA was just configured recently.

What do you think about moving the `className` beside `variable` and having some sort of comment about the dot?  That would serve two purposes:  (1) It would help to prevent similar errors in future and more strongly show that the two variables are very similar.  (2) It should trigger GHA so we can verify it's working for PRs.

mkelly1495:
Sure - I can do that this weekend.

Michael Kelly

On Mar 18, 2022, at 1:59 PM, Max Franz ***@***.***> wrote:

﻿
@mkelly1495, looks good.

I moved the base branch to unstable, because it has GitHub Actions tests. We've had some trouble with the reliability of Travis and currently only the unstable branch has GHA setup. GHA was just configured recently.

What do you think about moving the className beside variable and having some sort of comment about the dot? That would serve two purposes: (1) It would help to prevent similar errors in future and more strongly show that the two variables are very similar. (2) It should trigger GHA so we can verify it's working for PRs.

mkelly1495:
@maxkfranz I added comments as you requested and moved the definition of `tokens.className` next to the definition of `tokens.variable`. Hopefully it will be clear now why they are different.

maxkfranz:
@mkelly1495 

Sounds good.  I've assigned myself to merge this in and apply it to the applicable branches.  Possibly tomorrow

maxkfranz:
Ref:

- #2999 
- #2995

PR #3002, opened 2022-03-23
Include underlay into bounding box

### LINKED ISSUE #2998: Underlay gets clipped on pan and zoom
**Environment info**

- Cytoscape.js version : 3.21.0
- Browser/Node.js & version : Firefox 98.0

**Current (buggy) behaviour**

Underlay is clipped when panning or zooming the graph.
It happens on last node on an axis direction: top, bottom, left, right.

**Desired behaviour**

Underlay staying intact during interaction.

Just pan (or zoom) the graph around.

[https://output.jsbin.com/lawerelaxi/1](https://output.jsbin.com/lawerelaxi/1)

zakjan:
Fixes #2998


maxkfranz:
@zakjan 

Is `includeUnderlays` in the docs somewhere?

Re. the patch, are there cases where the bug would still exist?  What if both an overlay and an underlay are applied, for instance?

zakjan:
Right, thanks, I didn't realized this is a public function. Added both, documentation and handling the situation with both overlay and underlay.

maxkfranz (approved):

Issue #3099, opened 2023-02-03
add acyclic option to breadthfirst layout


moser-clow:
**Description of new feature**

Currently, the `breadthfirst` layout extension is using a naive method of cycle detection when `maximal` is set to `true`. Namely, if the extension has had to shift a node once already, and if the extension has determined that it must shift the node again, then the extension issues a `double maximal shift` message to the console and stops shifting nodes. The resultant graph will still have upwards edges, which is not ideal when the user _knows_ that a given graph is a DAG and thus all edges could be laid out in the same direction (downwards).

Instead, I suggest that the extension provide an `acyclic` option that would have an effect only if `maximal` is set to `true` (but could be used for future code that would require knowing whether the current graph is acyclic). If `acyclic` is set to `false` (the default), then the code operates as it does prior to this code change. If `acyclic` is set to `true`, then allow more than one shift of a given node.

**Motivation for new feature**

I'm using Cytoscape.js to render graphs that I've ensured are DAGs before layout. The graph data involves dependencies; if a directed edge exists from A to B, then B depends on A. I have a requirement that all such dependencies be rendered in a downward direction; if C depends on B which depends on A, then B will be below A and C will be below B.

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [ ] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

PR #3100, opened 2023-02-03
add acyclic option to breadthfirst layout

moser-clow:
Associated issue: https://github.com/cytoscape/cytoscape.js/issues/3099

Implement a solution to above-linked issue.

Author:

- [X] The proper base branch has been selected.  New features go on `unstable`.  Bug-fix patches can go on either `unstable` or `master`.
- [X] Automated tests have been included in this pull request, if possible, for the new feature(s) or bug fix.  Check this box if tests are not pragmatically possible (e.g. rendering features could include screenshots or videos instead of automated tests).
- [X] The associated GitHub issues are included (above).
- [X] Notes have been included (above).

Reviewers:

- [ ] Just after this pull request is merged, it should be applied to both the `master` branch and the `unstable` branch.  Normally, this just requires cherry-picking the corresponding merge commit from `master` to `unstable` -- or vice versa.

### DIFF
--- a/src/extensions/layout/breadthfirst.js
+++ b/src/extensions/layout/breadthfirst.js
@@ -15,6 +15,7 @@ const defaults = {
   nodeDimensionsIncludeLabels: false, // Excludes the label when calculating node bounding boxes for the layout algorithm
   roots: undefined, // the roots of the trees
   maximal: false, // whether to shift nodes down their natural BFS depths in order to avoid upwards edges (DAGS only)
+  acyclic: false, // whether the tree is acyclic and thus a node could be shifted (due to the maximal option) multiple times without causing an infinite loop; if you are uncertain whether a tree is acyclic, set to false to avoid potential infinite loops
   depthSort: undefined, // a sorting function to order nodes at equal depth. e.g. function(a, b){ return a.data('weight') - b.data('weight') }
   animate: false, // whether to transition the node positions
   animationDuration: 500, // duration of animation in ms if enabled
@@ -22,8 +23,7 @@ const defaults = {
   animateFilter: function ( node, i ){ return true; }, // a function that determines whether the node should be animated.  All nodes animated by default on animate enabled.  Non-animated nodes are positioned immediately when the layout starts
   ready: undefined, // callback on layoutready
   stop: undefined, // callback on layoutstop
-  transform: function (node, position ){ return position; }, // transform a given node position. Useful for changing flow direction in discrete layouts
-  acyclic: false // whether the tree is acyclic; if uncertain, set to false to avoid potential infinite loops
+  transform: function (node, position ){ return position; } // transform a given node position. Useful for changing flow direction in discrete layouts
 };
 /* eslint-enable */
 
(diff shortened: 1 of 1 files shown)


moser-clow:
@maxkfranz - just making sure you saw this. Thank you.

maxkfranz:
Minor:  I think the docs for built-in layout options is automatic (from the code), so your new option should probably be defined somewhere closer to the other option field it affects.  Some kind of explicit description of how `acyclic` affects `maximal` in the comments there would help consumers of the library to understand what to expect when using them.

Otherwise, looks good to go

maxkfranz:
> @maxkfranz: As requested, I've moved the definition of the `acyclic` option to be next to the `maximal` option and added a bit more of a description. 

Great

> Random thought: Should it be named something different such as `enableMultipleShifts`?

That’s a good point.  The name is important.

Whether the graph is a DAG or not could be detected properly during the early step where it’s running `eles.bfs()`.  In that case, might the best name for your option be no name at all?

You could use the same `acyclic` name only internally, and it would be set automatically for the API consumer.  

The API consumer wouldn’t be able to make mistakes:  They couldn’t mis-set `acyclic`, because they can’t set it at all.  They couldn’t mis-set `maximal`, because it could be safely ignored for cyclic graphs (maybe with a proper warning message).

moser-clow:
> the early step where it’s running `eles.bfs()`

Where is this step? I'm having trouble finding it.

What do you imagine the proper detection of DAG-ness would look like?

maxkfranz:
I looked into this more, and I've given this more thought.

The `maximal` feature originated from someone who, early in the library's life, wanted a hierarchical layout that avoided upwards edges in a DAG.  

At the time, the only hierarchical layout in Cytoscape was breadthfirst.  Where else would that kind of feature go but breadthfirst?

The goal of breadthfirst was to have a very simple, builtin layout for hierarchical graphs.  Aligning the nodes based on a simple BFS traversal makes sense for that:  The main use case is that the user wants to see how many hops away, at minimum, a given node is from the root (directed or undirected).  The `maximal` option doesn't really fit in that paradigm, but rightly or wrongly, it was added anyway.

Today, there are more sophisticated hierarchical layouts that have all sorts of options to customise the layout -- like Dagre, Klay, and Elk.

Frankly, I think the `maximal` option should be deprecated.  It's not a great feature for breadthfirst, and that kind of feature lends itself better to other layouts today.

With regards to your PR specifically, I see these points:

(1) We shouldn't add any new API re. `maximal`.

(2) `maximal` should be removed from the documentation.  Deprecated.

(3) You're welcome to improve the PR to make the `maximal` automatic with cycle detection.  The implementations for BFS and DFS [are the same, with minor adjustments](https://github.com/cytoscape/cytoscape.js/blob/be10eed0050681fc9cc53c0ceeabe48d8be6a162/src/collection/algorithms/bfs-dfs.js).  That function could be expanded to allow for returning a collection that represents a topological ordering of the graph (DFS postorder, rather than preorder).  If every edge in the graph respects the ordering, then it should be a DAG.  You can use that detection to amend the `maximal` functionality as you suggested, but we couldn't guarantee for how much longer the feature would exist (i.e. v4). 

(4) As an even better alternative to (3), you could copy the existing breadthfirst layout into a new, external layout extension that uses DFS -- perhaps called `depthfirst`.  The DFS ordering would be better suited to that use case generally, and if you have the need for further adjustments, being able to get the topological ordering from (3) may be useful here too.

(5) As an even more minimalistic alternative to (3) and (4), you could use Dagre, Klay, or Elk instead.

moser-clow:
@maxkfranz - I've updated the PR so that `maximal` is made automatic with cycle detection. I believe that deprecation of `maximal` should be a separate change.

maxkfranz:
Thanks.  I'll look into this early next week

Issue #3021, opened 2022-05-10
Events style is not dynamically applied


Ahmed-username:
**Before you post**

A request for help or a requests for a how-to should be directed to Stack Overflow:
http://stackoverflow.com/questions/tagged/cytoscape.js

**Environment info**

- Cytoscape.js version : 3.20.0
- Browser/Node.js & version : Google Chrome Version 101.0.4951.59 (Official Build) (64-bit)

**Current (buggy) behaviour**

[Events style](https://js.cytoscape.org/#style/events) is not applied when dynamically added/removed from an element.

**Desired behaviour**

Applying "events: no" or "events: yes" should be considered even when applied after the element has been created. 

**Minimum steps to reproduce**

To replicate the bug check this [JsBin](https://jsbin.com/relarapina/edit?console,output)
When the node is clicked, a log is printed conforming that the tap event is handled normally.
However when the "Add .no-event" button is clicked, a class is added that adds the attribute "events: no" to the node, yet when the node is clicked the events are handled normally.

On the other hand, this [JsBin](https://jsbin.com/busupogewu/1/edit?console,output) has "events:no" applied when the node is created, and the tap event is not picked up as expected.
But when the "Remove .no-event" button is clicked, the click event is still not picked up although the "events:no" style has been removed.

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [x] Ensure that the reporter has included a reproducible demo.  They can easily fork this JSBin demo: http://jsbin.com/fiqugiq
- [x] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [x] The commits have been incorporated into the corresponding branches.  Bug-fix patches go on
    - [x] `master`,
    - [x] `unstable`, and
    - [x] the previous feature release branch (e.g. `1.1.x` if the current release is 1.2).
- [x] The issue has been labelled as a [`bug`](https://github.com/cytoscape/cytoscape.js/labels/bug), if necessary.

maxkfranz:
It's important that the issue template is used and filled in completely in order for issues to be addressed properly and promptly.  Please edit your issue according to the template.  I've started it for you.

maxkfranz:
@mikekucera, could you take a look at this one?  Now that the issue is following the template, there should be enough information to reproduce this in the debug page.

Ahmed-username:
Hello,
Any updates about this issue?

thanks

[...]

PR #3151, opened 2023-08-11
Clear z-order cache on style change.

mikekucera:
Associated issues: #3021

The pull request fixes issue #3021 

The base renderer keeps a cache of "interactive" elements along with the z-order cache. Changing 'events' style property of an element changes its interactivity, but this cache is not updated. This pull request clears the cache when the 'style' event occurs.

The cache is created here: https://github.com/cytoscape/cytoscape.js/blob/658aed7234ef9d66e546af23c5134b3571c3c502/src/extensions/renderer/base/coord-ele-math/z-ordering.js#L43

Its used here:
https://github.com/cytoscape/cytoscape.js/blob/658aed7234ef9d66e546af23c5134b3571c3c502/src/extensions/renderer/base/coord-ele-math/coords.js#L94

And affects this return value:
https://github.com/cytoscape/cytoscape.js/blob/658aed7234ef9d66e546af23c5134b3571c3c502/src/extensions/renderer/base/load-listeners.js#L412

That's why the tap event fires when it shouldn't.

Note: I'm not sure how to write a unit test for this. The cache is internal to the base renderer and I don't see a way through the API to check the state of that cache. Any advice on how to write a test for this is appreciated.

**Checklist**

Author:

Reviewers:

### DIFF
--- a/src/extensions/renderer/base/index.js
+++ b/src/extensions/renderer/base/index.js
@@ -140,7 +140,6 @@ BRp.notify = function( eventName, eles ){
     || eventName === 'load'
     || eventName === 'zorder'
     || eventName === 'mount'
-    || eventName === 'style'
   ){
     r.invalidateCachedZSortedEles();
   }
--- a/src/extensions/renderer/base/index.js
+++ b/src/extensions/renderer/base/index.js
@@ -134,11 +134,13 @@ BRp.notify = function( eventName, eles ){
   }
 
   if(
-    eventName === 'add' || eventName === 'remove'
+    eventName === 'add' 
+    || eventName === 'remove'
     || (eventName === 'move' && cy.hasCompoundNodes())
     || eventName === 'load'
     || eventName === 'zorder'
     || eventName === 'mount'
+    || eventName === 'style'
   ){
     r.invalidateCachedZSortedEles();
   }
--- a/debug/tests.js
+++ b/debug/tests.js
@@ -766,4 +766,35 @@
       de.move({ source: 'd', target: 'e' });
     }
   });
+
+  test({
+    name: "events:no",
+    displayName: "events: no",
+    description: "Apply events:no style to all nodes. Clicking on nodes should no longer affect the node.",
+    
+    setup: function(){
+      cy.nodes().style(
+        { events: 'no' }
+      );
+    },
+    teardown: function(){
+      cy.nodes().removeStyle();
+    }
+  });
+
+  test({
+    name: "text-events:yes",
+    displayName: "text-events: yes",
+    description: "Apply text-events:yes style to all nodes. Clicking on node labels should select the node.",
+    
+    setup: function(){
+      cy.nodes().style(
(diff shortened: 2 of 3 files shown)


src/extensions/renderer/base/index.js
```diff
@@ -134,11 +134,13 @@ BRp.notify = function( eventName, eles ){
...
 
   if(
-    eventName === 'add' || eventName === 'remove'
+    eventName === 'add' 
+    || eventName === 'remove'
     || (eventName === 'move' && cy.hasCompoundNodes())
     || eventName === 'load'
     || eventName === 'zorder'
     || eventName === 'mount'
+    || eventName === 'style'
```
maxkfranz, line 143: Great diagnosis.

If a missed z-sorting is the issue, then this approach would work -- but it is a bit of a shotgun approach. It would undo some performance optimisations, since the z-order would be recalculated every time any style property is changed.  The z-order sorting can be expensive if it happens frequently, and not every property change necessitates a recalculation (e.g. just changing node colour).  This helps keep things like animations on properties like node colour inexpensive.

Defining a `triggersZOrder ` condition in the property spec would be a bit more precise:

https://github.com/cytoscape/cytoscape.js/blob/9e388771139c10cff330dcc586661f8471b000bc/src/style/properties.js#L232

E.g. for the `display` property:

https://github.com/cytoscape/cytoscape.js/blob/9e388771139c10cff330dcc586661f8471b000bc/src/style/properties.js#L237

Aside:  The `text-events` property might also be affected by this bug, since it's analogous to `events`.

maxkfranz (commented):
Great.  Looks like you've identified the root cause.

Here's a few points of feedback before we merge this in:

> Note: I'm not sure how to write a unit test for this. The cache is internal to the base renderer and I don't see a way through the API to check the state of that cache. Any advice on how to write a test for this is appreciated.

It's difficult to write good, automated rendering tests.  Sometimes, it's best to add to [the dropdown on the debug page](https://github.com/cytoscape/cytoscape.js/blob/9e388771139c10cff330dcc586661f8471b000bc/debug/tests.js), which just makes it a bit easier to test things manually with a setup and a teardown.  Basically, you select an entry from the dropdown and then see if everything works as expected manually.

For this bug, one automated test in the test suite could be this:  Don't test the bug completely.  That's difficult in a headless environment.  However, [a test with the styling enabled](https://github.com/cytoscape/cytoscape.js/blob/9e388771139c10cff330dcc586661f8471b000bc/test/collection-style.js) could test for the `zorder` event being fired on an element when it's value for the `events` property is changed.  This test wouldn't be perfect, but it's better than nothing and it's simple.

mikekucera:
I made the changes you suggested.

maxkfranz:
Perfect.  Thanks

maxkfranz:
Cherry-picked to master

maxkfranz:
Cherry-picked to 3.25.x

Issue #3191, opened 2023-11-28
Add edge <position>-arrow-width for hollow shapes 


EliotRagueneau:
**Description of new feature**

Hi, we would like cytoscape.js to allow us to customize how wide the stoke of hollow shapes are, because at the moment they do not blend with the rest of our developped style, but we need them to follow SBGN standards. We were thinking of a property like <position>-arrow-width, but of course it can be simpler than that, like using the edge width on the arrow hollow shapes as well.
<img width="1008" alt="Screenshot 2023-11-28 at 11 17 46" src="https://github.com/cytoscape/cytoscape.js/assets/42945601/488cf346-7ec1-4068-9f88-4df61a08cbda">
<img width="506" alt="Screenshot 2023-11-28 at 11 17 55" src="https://github.com/cytoscape/cytoscape.js/assets/42945601/56e39e09-012c-4aac-a1f3-e5bec21cf2a7">

**Motivation for new feature**

We are developing a new [reactome-cytoscape-style](https://github.com/reactome-pwp/PathwayBrowser/tree/main/projects/reactome-cytoscape-style), similar to SBGN-stylesheet, in order to use it for the new Reactome Pathway-Browser we are developing, but also in our new curator tool.

We need to go by the SBGN standards, but we would like to do so without compromising our newly developed visual identity

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [ ] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

maxkfranz:
Is the requirement for your app that the stroke with of hollow arrows match the main line width, or do they need to vary?

The simplest option may be to expose another width prop for the arrows. Another option is an enum, with values like 'default' and 'match-line'. 

A hybrid approach would be to support both, i.e. custom px values or a special 'match-line' enum value.

maxkfranz:
It looks like Ugur may be away. Let's move forward with the hybrid approach (px value or 'match-line' enum value). 

[...]

Issue #2019, opened 2017-11-10
Cytoscape v4 changes


maxkfranz:
Brainstorming of improvements that would cause breaking changes for v4

- [ ] `ele.json()` returns `classes` as an array

pmackay:
>ele.json() returns classes as an array

Does this mean that it would _include_ classes as well as other data? 

Is there any way currently to get the list of classes an element has?

maxkfranz:
> Does this mean that it would include classes as well as other data?

`classes: ['foo', 'bar']` versus `classes: 'foo bar'`

> Is there any way currently to get the list of classes an element has?

`ele.json().classes`

StevenLangbroek:
Since a `class` is a concept in Javascript now, maybe it's better to use `classNames`? Ecosystem seems to be standardising on this.

maxkfranz:
Yes, but that wouldn't have to be a breaking (v4) change.  The old key could still be supported.

theFrawley:
Our app would really benefit from Pointer Events.
I believe that web is moving that way - with pointer events now supported in all the main browsers.

Please please please could you add them to Cytoscape.js? :)

maxkfranz:
This issue is for discussing v4 API -- i.e. breaking API changes.  If you have a general feature in mind that is additive w.r.t. API, then feel free to make an ordinary feature request and the corresponding pull request.

PR #3194, opened 2023-12-02
:sparkles: Add <position>-arrow-width property for hollow shapes

EliotRagueneau:
Associated issues: #3191 

- This PR adds a new style property on edges `<position>-arrow-width` which allow user to customise the thickness of the hollowed arrow shapes stroke. 
  - default value : 'match-line' ==> the stroke width is the same as the edge width
  - number value ==> considered as a px width
  - string value ==> `${number}${unit}` where unit can either be `px`, `em` or `%`. 
    - `%` value is based on edge width. e.g. 100% ahas the same behaviour as 'match-line'
- This PR includes modifications to the documentation to present the new property
- This PR includes modifications to the debug to easily visualise the impact of the property

Screenshot 2023-12-02 at 13 09 20

**Checklist**

Author:

Reviewers:

### DIFF
--- a/debug/init.js
+++ b/debug/init.js
@@ -17,22 +17,25 @@ var cy, defaultSty, options;
           'target-arrow-shape': 'triangle',
           'mid-target-arrow-shape': 'triangle',
           'mid-source-arrow-shape': 'triangle-backcurve',
-          'target-arrow-fill': 'hollow',
-          'source-arrow-fill': 'hollow',
-          'source-arrow-width': 2,
-          'target-arrow-width': '50%',
         })
       .selector('#ab')
         .style({
           'curve-style': 'unbundled-bezier',
           'control-point-distances': [ 20, -100, 20 ],
           'control-point-weights': [ 0.25, 0.5, 0.75 ],
+          'source-arrow-fill': 'hollow',
+          'source-arrow-width': 2,
+          'target-arrow-fill': 'hollow',
+          'target-arrow-width': 'match-line',
         })
       .selector('#bc')
         .style({
           'curve-style': 'segments',
           'segment-distances': [ 20, -80 ],
           'segment-weights': [ 0.25, 0.5 ],
+          'source-arrow-fill': 'hollow',
+          'source-arrow-width': '50%',
+          'target-arrow-fill': 'hollow',
         })
       .selector('#ef')
         .style({
--- a/debug/init.js
+++ b/debug/init.js
@@ -16,7 +16,11 @@ var cy, defaultSty, options;
           'source-arrow-shape': 'triangle-backcurve',
           'target-arrow-shape': 'triangle',
           'mid-target-arrow-shape': 'triangle',
-          'mid-source-arrow-shape': 'triangle-backcurve'
+          'mid-source-arrow-shape': 'triangle-backcurve',
+          'target-arrow-fill': 'hollow',
+          'source-arrow-fill': 'hollow',
+          'source-arrow-width': 2,
+          'target-arrow-width': '50%',
         })
       .selector('#ab')
         .style({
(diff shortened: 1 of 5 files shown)


debug/init.js
```diff
@@ -16,7 +16,11 @@ var cy, defaultSty, options;
           'source-arrow-shape': 'triangle-backcurve',
           'target-arrow-shape': 'triangle',
           'mid-target-arrow-shape': 'triangle',
-          'mid-source-arrow-shape': 'triangle-backcurve'
+          'mid-source-arrow-shape': 'triangle-backcurve',
+          'target-arrow-fill': 'hollow',
+          'source-arrow-fill': 'hollow',
+          'source-arrow-width': 2,
+          'target-arrow-width': '50%',
```
maxkfranz, line 23: Would you apply these new styles only to a subset of the edges?  The default style here should ideally show as many combinations of styles upfront.  

That way, anyone who looks at the debug page graph would immediately notice if any of these styles are broken.  E.g., you can see right away that both filled and hollow edges are OK.

maxkfranz:
Are there any cases that would affect compatibility of existing apps?

Let's say I have an app today and I upgrade Cytoscape.js to the latest version (i.e. the version with your feature, when it is released).  In my app, let's say I have hollow edges.  Would my graph look the same as before?

debug/init.js
EliotRagueneau: In my case, and more generally in SBGN in general, yes, only some edges are hollow, the rest is filled.

Here in the debug, you can see filled inside the "mid" arrows, and hollow at the "end" arrows. Then indeed, this test doesn't represent the full extent of all possibilities as it doesn't include the default "match-line", nor the usage of em units.
Do you want me to include them as well?

EliotRagueneau:
@maxkfranz At the moment no, since I set the default to "match-line" as you suggested, 
Therefore, if the edge width is not `1`, the arrow stroke-width will not be `1`  as it is currently (default = `1`).
I can put back the default to `1` to have consistent behaviour if you want. 

It's just not necessarily the best looking solution, but I understand that consistency can be more important than improving default appearance.

debug/init.js
EliotRagueneau: Done in https://github.com/cytoscape/cytoscape.js/pull/3194/commits/69843f90fb10cd9cc7a47f00bc9026e1fd446e58

maxkfranz: Great.  Thanks

maxkfranz:
@EliotRagueneau, I agree that the match/100% is a better default generally.  Making sure that, when a dev updates Cytoscape, everything works the same as before is an important requirement.  People really appreciate easy upgrades, and we don't want to discourage people from upgrading by giving them extra work to carry out upgrades.

Let's plan to change the default to the better match/100% value for v4:

- https://github.com/cytoscape/cytoscape.js/issues/2019

maxkfranz (approved):

maxkfranz:
@EliotRagueneau, I tested out the PR locally, and it looks very good.

@mikekucera, @chrtannus, I've added you on as reviewers.  Let me know if you think anything could be improved generally -- or specifically re. Cy3 compatibility.

EliotRagueneau:
@maxkfranz Thanks a lot! 
Happy to have contributed :D

maxkfranz:
> > re. Cy3 compatibility.
> 
> Desktop doesn't have a visual property for "open" (ie hollow) arrow shape stroke thickness, and I don't think theres any plans to add one. The default is like 'match-line', the arrow shape size is relative to the edge thickness. Note, the set of arrow shapes in desktop is different from cy.js so its already not fully compatible.

Great.  It sounds like there aren't any compatibility blockers, so let's merge this in as-is.

Issue #3199, opened 2023-12-05
Support edge line properties for node borders


EliotRagueneau:
**Description of new feature**

It would be awesome if we could apply the folowing list of edge-line properties on the node borders

- `width` ==> `border-width` **(already in place)**
- `line-color`  ==> `border-color` **(already in place)**
- `line-style` ==> `border-style` **(already in place)**
- `line-opacity` ==> `border-opacity` **(already in place)**
- `line-cap` ==> `border-cap`
- `line-dash-pattern` ==> `border-dash-pattern`
- `line-dash-offset` ==> `border-dash-offset`
- `line-fill` ==> `border-fill` (More difficult than the original one)
-  *not existing* ==> `border-position`  (*center*, `inside`, `outside`)

**Motivation for new feature**

In reactome-cytoscape-style, we have designed ndoes that should look like this

<img width="723" alt="Screenshot 2023-12-05 at 19 59 10" src="https://github.com/cytoscape/cytoscape.js/assets/42945601/7952c6bf-2277-49a2-a7ad-4c6da4484a08">

But the closest I can get border-wise is this 
![Screenshot 2023-12-05 at 19 59 55](https://github.com/cytoscape/cytoscape.js/assets/42945601/55daa0ba-8858-4f4f-82e4-bf5c0f85597d)

The `border-position` and `border-fill` are more complex to implement, so I don't have much expectation about them, but they would be nice.

Also in our specific case, we don't need the `border-fill`, but the `border-position` is definitely important to achieve the aspect as our mockup

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [ ] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

maxkfranz:
Everything that is supported natively by canvas are clearly good things to support -- mostly overlapping with existing edge props.

> The border-position and border-fill are more complex to implement, so I don't have much expectation about them, but they would be nice.

[...]

PR #3202, opened 2023-12-13
Border properties

EliotRagueneau:
Associated issues: 

- #3199 

- This PR adds the following properties
 * **`border-cap`** : The cap style of the node's border; may be `butt`, `round` or `square`.
 * **`border-dash-pattern`** : The `dashed` line pattern which specifies alternating lengths of lines and gaps. (e.g. `[6, 3]`).
 * **`border-dash-offset`** : The `dashed` line offset (e.g. `24`). It is useful for creating edge animations.
 

**Checklist**

Author:

Reviewers:

- [x] All automated checks are passing (green check next to latest commit).
- [x] At least one reviewer has signed off on the pull request.
- [x] For bug fixes:  Just after this pull request is merged, it should be applied to both the `master` branch and the `unstable` branch.  Normally, this just requires cherry-picking the corresponding merge commit from `master` to `unstable` -- or vice versa.


EliotRagueneau:
I've started to implement `border-position` using the strategy defined at https://stackoverflow.com/a/34234589/11028828

However, I didn't want to create a new canvas for each nodes with this type of border for performance reason, and as the  author explained in the previous ticket, just by using it within the global context does create a lot of issues with the rest of rendering, so I've put on hold those changes for now, as you suggested @maxkfranz 

Let me know if you are aware of an easy way and performat way to do this. (inside can be done using clips also I beileve, but not outside)

maxkfranz:
> I've started to implement `border-position` using the strategy defined at https://stackoverflow.com/a/34234589/11028828
> 
> However, I didn't want to create a new canvas for each nodes with this type of border for performance reason, and as the author explained in the previous ticket, just by using it within the global context does create a lot of issues with the rest of rendering, so I've put on hold those changes for now, as you suggested @maxkfranz

Yes, I wouldn't use GCO for this.

> 
> Let me know if you are aware of an easy way and performat way to do this. (inside can be done using clips also I beileve, but not outside)

Inside -- regular clip of node shape

Outside -- possibly an even-odd clip, with one box for the full bounds and the node shape or some other combination

See https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clip

EliotRagueneau:
Just added the support to `border-position` as you suggested using clips and evenodd, thanks for the advice!

EliotRagueneau:
This strategy might create issues if users are using a custome polygon which cross itself, like an hourglass for instance

EliotRagueneau:
I've tested a custom hourglass shape and it still works great 🥳 
While doing so, I've realised that we didn't have any lineJoin property, and it becomes useful with outer border in my opinion, so I have added it

EliotRagueneau:
Sorry about those missing semi colons

maxkfranz (approved):
@EliotRagueneau, anything else before this is merged in?

EliotRagueneau:
No I think we're good on this one

Issue #3201, opened 2023-12-12
New edge type: smooth-segments


EliotRagueneau:
**Description of new feature**
It would be awesome to internally support a new edge type: smooth segments.

I have a working implementation using unbundled bezier edges already. 
You can see a demo of it at https://reactome-pwp.github.io/PathwayBrowser/R-HSA-453279 
However, to do this, I have to define 3 points for each segment end point, which is not super user-friendly, specially when using a tool like https://github.com/iVis-at-Bilkent/cytoscape.js-edge-editing.

Also, at the moment, I'm using quadratic curves to do the rounding of the corners, but I think it would look even better if we could use arcs instead.

The principle I used so far is represented in the following schema
![smooth-segments](https://github.com/cytoscape/cytoscape.js/assets/42945601/17dff004-5b08-489d-8799-63b6bf6916b7)

In terms of added properties, I would consider adding the following:

- `curve-style`: `smooth-segments`
- `control-point-distances` : same as bezier
- `control-point-weights`: same as bezier 
- `control-point-radii`: `number` | `number[]` | `string` 
  - If one number is provided, apply it to all control points. 
  - If more than one radius is provided, each radius is for a specific control point. 
  - The string should be parsed like distances and weights are, and otherwise follow the same logic
- `smoothing-type`: `arc` | `quadratic`

The code implementing it the following:
```typescript
import {Position} from "cytoscape";
import {add, array, NDArray, subtract} from "vectorious";

/**
 * Add roundness to a segment edge by adding for each anchor position an additional point just before and just after
 * at a distance of radius from the original anchor point.
 *
 * With Cytoscape Bezier edges, this renders as a segmented edges with smoothed corner.
 * For a schematic explanation, @see https://raw.githubusercontent.com/reactome-pwp/PathwayBrowser/main/doc/smooth-segments.png
 *
 * @param source The position of the source node
 * @param target The position of the target node
 * @param anchors The intermediate segment points
 * @param radius How far away the new points need to be from the anchor points.
 * A bigger radius means smoother corners,
 * but it shouldn't be bigger than the segment
 * @private
 */
export function addRoundness(source: Position, target: Position, anchors: Position[], radius = 40): Position[] {
  const output: NDArray[] = []
  if (anchors.length === 0) return output;

  const points = [source, ...anchors, target];
  for (let i = 1; i < points.length - 1; i++) {
    let previousPoint = toVector(points[i - 1]);
    let currentPoint = toVector(points[i]);
[...]

Issue #3198, opened 2023-12-05
border-radius property for node shapes with rounded corners


EliotRagueneau:
**Description of new feature**

It would be really great if users could easily specify what is the border-radius of all the "rounded" node shapes.
Additionally, adding this property would enable effectively new shapes, such as the pill shape which is just a rounded rectangle with a corner radius of half the height

![corner-radius](https://github.com/cytoscape/cytoscape.js/assets/42945601/a7c00cf0-44cf-4eeb-b889-cd880b1c92d7)

**Motivation for new feature**

In the reactome-cytoscape-style library we're developing, we need bigger border-radius than the ones provided by default . 
At the moment we're using background-image svgs to do it, but that is extremely heavy for something that could just be a simple property of a shape, without impacting too much performances (I believe).   

We would like to boost our performances by using native shape renderers instead of custom SVGs

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [ ] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

EliotRagueneau:
This is I guess somehow related to #1167 , but it never got through. 
In my context I don't care about parent nodes specifically, just nodes in general

maxkfranz:
Adding initial thoughts while I'm online just now:  A large border radius on compound parent nodes would be troublesome.  The bounds of the parent are defined by the collective bounds of the descendants -- a rectangle.  A large border radius on a parent would allow a child to visually escape the bounds of the parent at the corners, possibly escaping entirely or possibly only overlapping the border.

For non-parent nodes, large border radii are probably without conflict.

Some initial options:

(1) Forbid custom border radii on parent nodes.
- Inexpensive to implement
- Consumers of the library can't make mistakes

(2) Allow custom border radii on parent nodes and put harsh warnings and mitigations in the documentation.
- Inexpensive to implement
- Consumers of the library can easily make mistakes
- One mitigation: Set a padding that's appropriate to the radius

[...]

PR #3203, opened 2023-12-15
Border radius

EliotRagueneau:
Associated issues: 

- #3201 
- #3198 

- This PR adds border-radius support, both for rounded polygons and rounded rectangle
- This PR adds `round-segments` & `round-taxi`, as well as their respective radius properties

**Checklist**

Author:

Reviewers:


EliotRagueneau:
I'm going in hollidays until the 08/01/2024, so I posted this pull request though not yet fullly complete.

#### TODO
- [ ] Store corner data rather than recalculating it for the `draw`, `intersectLine` and `checkPoint` functions in node-shapes.js `generateRoundPolygon()`. 
  - I've prepared what to be stored with round.js `getRoundCorner()`, but I wasn't sure where to store it.
  - The previous way of pre calculating points and unit vectors was at the node shape generator level, therefore prior to any access to radius data. Need to find a place to store. them and reuse them wherever needed
- [ ] Add support for padding with border radius as discussed in #3203  
- [ ] Generate documenation for the new properties
- [ ] Generate demo for the new edge types
- [ ] Do further testing to make sure that there are never problems


Of course, feel free to complete some of those tasks while I'm away, and to give me feedback on some aspects.
This is a much bigger PR than all the previous ones I've made, so help would be really appreciated

maxkfranz:
Pinned

EliotRagueneau:
Hi @maxkfranz , I hope you had nice holidays.

Could you please give me some advice on where to store pre-calculated results for rounded node shapes ?
> Store corner data rather than recalculating it for the draw, intersectLine and checkPoint functions in node-shapes.js generateRoundPolygon().
> > I've prepared what to be stored with round.js getRoundCorner(), but I wasn't sure where to store it.
> > The previous way of pre calculating points and unit vectors was at the node shape generator level, therefore prior to any access to radius data. Need to find a place to store. them and reuse them wherever needed

EliotRagueneau:
Reminder: 
- [ ] Check if the new roundness types are well supporting underlay
- [ ] Check if the new roundness types are well supporting overlay

maxkfranz:
Some initial notes:

1. Cached points can be stored in `ele._private.rstyle`.
2. The caching of the points is initiated here: https://github.com/cytoscape/cytoscape.js/blob/unstable/src/extensions/renderer/base/coord-ele-math/edge-control-points.js#L928-L932
3. `allpts` is also used for calculations and it may need to be updated, especially for a large radius on an edge.

EliotRagueneau:
For the edges, I've managed to cache what is need I believe.
`allpts` I don't think I touched it, gonna have a look

EliotRagueneau:
As you can see bellow, I've just tweaked a bit the outline formula to adjust the roundness of the outline, so that it looks neat.

Also I've checked for overlay and underlay, and both are working great for the new edge curvatures
Screenshot 2024-01-11 at 11 36 17

src/style/properties.js
```diff
@@ -356,9 +358,11 @@ const styfn = {};
     { name: 'control-point-weights', type: t.numbers, triggersBounds: diff.any },
     { name: 'segment-distances', type: t.bidirectionalSizes, triggersBounds: diff.any },
     { name: 'segment-weights', type: t.numbers, triggersBounds: diff.any },
+    { name: 'segment-radii', type: t.numbers, triggersBounds: diff.any },
```
maxkfranz, line 361: Let's make this `segment-radius` to be consistent in naming with your other properties.  A lot of the other, existing properties are singular even though they accept multiple values (e.g. `background-image`).

maxkfranz (commented):
Just a point to consider re. naming before we merge this in:

src/style/properties.js
maxkfranz: I didn't notice you had aliases.  Disregard

maxkfranz:
Looks reasonable.  There are some minor linter errors to resolve. 

Are there any other tests or points of consideration before merging this in?

EliotRagueneau:
The thing that worries me the most really is the rounded node shapes which are now less optimised than before because we are not caching the unit vectors, I was hoping to be able to cache the rounded corners themselves within the nodes.

You said it was okay, but I'm a bit suprised because we are calculating them in 3 different functions for each round shapes, and potentially, if the user moves a node, we need to recalculate the roundness again and again to calculate the arrow head position using the intersection function. But if you say that's alright, I'm fine with it.

maxkfranz:
Of course, it’s better to cache.However, the nodes have texture caching applied at a level above the points caching.  So, the points only need to be recalculated for drawing when the texture cache misses.One area that would be affected is the hit tests, for determining when you click on a node, for instance.It’s up to you if you want to apply caching of the points now.  I think it would be better with caching, but it could also be added later.  Only people who use the new, custom border radius would be affected, right?  If so, it’s suboptimal but not a regression.On Jan 12, 2024, at 05:22, Eliot Ragueneau ***@***.***> wrote:﻿
The thing that worries me the most really is the rounded node shapes which are now less optimised than before because we are not caching the unit vectors, I was hoping to be able to cache the rounded corners themselves within the nodes.
You said it was okay, but I'm a bit suprised because we are calculating them in 3 different functions for each round shapes, and potentially, if the user moves a node, we need to recalculate the roundness again and again to calculate the arrow head position using the intersection function. But if you say that's alright, I'm fine with it.

EliotRagueneau:
Well no, with the current code, the new way to calculate corners is applied to all the rounded polygons, not just those for which the user specified the border radius. I did this to avoid having yet other shapes like "customisable-round-rectangle", and because the current implementation of the rounded shapes is failing when the width and the height are dissimilar. 
Therefore the lack of node caching would is harmful anytime a user use a "round-xxx" shape

EliotRagueneau:
I've implemented the node caching, it's a bit more dirty than the previous caching since it is at the node level instead of the abstract shape definition, but the cache is well being used whenever we can, and updated whenever we need. 
I think this covers all the concerns I had.

maxkfranz:
Looks great.  Merged.

If you'd like to add anything else to this feature, or you notice any issues, just let me know.

IntAct-EBI:
I noticed while testing it that for `round-segments`, it tends to create loops if a point is collinear to the previous and the following point, e.g. if A, B and C are on a single line, then B will have a loop on top of it.

In theory, users shouldn't put unnecessary control points, and these are unnecessary since they do not make the edge  move in any way, but it can be confusing. Do you think I should prevent this issue from happening? Like by ignoring collinear points? Is it too late to fix it for next release?

EliotRagueneau:
I noticed while testing it that for `round-segments`, it tends to create loops if a point is collinear to the previous and the following point, e.g. if A, B and C are on a single line, then B will have a loop on top of it.

Screenshot 2024-01-29 at 18 41 12

In theory, users shouldn't put unnecessary control points, and these are unnecessary since they do not make the edge  move in any way, but it can be confusing. Do you think I should prevent this issue from happening? Like by ignoring collinear points? Is it too late to fix it for next release?

maxkfranz:
There are all sorts of cases where people can create invalid beziers.

We print some warnings in the console for the worst cases that get detected with no additional computational cost (e.g. the bezier can't be drawn at all).

If you think there can be an inexpensive check to detect this case, then we can either automatically correct the points or we can print a warning in the console.  Someone might create collinear points when creating control points interactively.  Consider the case where you have a few points and it's easier to just drag one point to be collinear rather than to delete it.  That example demonstrates that autocorrection would generally be preferable.

> Is it too late to fix it for next release?

We don't need to be too strict about the release date of 3.29.0.  Let's consider 2024-02-12.  Would that be enough time to add your proposed fix?

EliotRagueneau:
yes that's probably more than enough, thanks a lot!
Will fix that ASAP :D

EliotRagueneau:
It seems my new commits are not showing in this PR, so I'll create a new PR

Issue #3243, opened 2024-05-15
Styling the edge lines with line-outline properties (border)


xkdgo:
**Description of new feature**

In addition to existing overlays/underlays add new properties to edges
`line-outline-width` and `line-outline-color`

Those properties are add new border visualisation of edge when `line-outline-width` > 0
![line-ouline-dashed](https://github.com/cytoscape/cytoscape.js/assets/38592352/7f136552-d96a-430d-bd0b-a57626531d49)

**Motivation for new feature**

This feature is usable when use library for network weathermap. [There was a dialogue about this feature.](https://github.com/cytoscape/cytoscape.js/discussions/3145)
When there is zero traffic between nodes it can be shown in white, but this is impossible now when the background is also white.
Underlay can resolve this problem only when line is not dashed. 
When the feature is implemented , overlay/underlay can be used when highlighting a line.
In example below
```
      "style": {
          // existed properties
          "width": 3,
          "line-color": "white",
          "underlay-opacity": "0.3",
          "underlay-color": "red",
          "underlay-padding": "3px",
          "line-style": "dashed",
          "line-dash-pattern": [15, 15],
          
          // new feature proposal
          "line-outline-width": 2, // default 0 mean switched off
          "line-outline-color": "#000" // default "#000"
      }
```
visualisation will be
![line-ouline-dashed-with-undelay](https://github.com/cytoscape/cytoscape.js/assets/38592352/51d2361b-a385-4e16-bb2d-d8c291c272b0)

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [ ] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

PR #3244, opened 2024-05-16
Feature/line outline

xkdgo:
Associated issues: 

- https://github.com/cytoscape/cytoscape.js/issues/3243

[Demo with outlines](https://github.com/cytoscape/cytoscape.js/assets/38592352/2e23b3fa-82cc-4280-8377-be90e45dde08)
[line-outline-demo.zip](https://github.com/cytoscape/cytoscape.js/files/15334183/line-outline-demo.zip)

**Checklist**

Author:

Reviewers:


mikekucera:
Hi, thanks for the pull request. 

I have one concern,

This screenshot shows the styles from the demo, the line-outline-width is set to 2 and the underlay has a padding of 3px. You can see that the underlay is slightly wider than the outline (as expected).
Screenshot 2024-05-16 at 9 53 47 AM


However, when I set the underlay padding to 2px it looks like below. If the outline and the underlay are both set to 2, shouldn't they line up perfectly? I think It should be possible to make the underlay and outline have the exact same width so that they match.
Screenshot 2024-05-16 at 9 39 53 AM

xkdgo:
> It should be possible to make the underlay and outline have the exact same width so that they match.

To test this i changed line's width calculation
from
```
context.lineWidth = edgeWidth + lineOutlineWidth;
```
to
```
context.lineWidth = edgeWidth + lineOutlineWidth / 2;
```
[line-outline-demo-w.zip](https://github.com/cytoscape/cytoscape.js/files/15338651/line-outline-demo-w.zip)

xkdgo:
@mikekucera I was wrong when I changed line's width calculation.
> It should be possible to make the underlay and outline have the exact same width so that they match.
But why we should do it ?

lets compare our width calculations
Underlay width calculate as double of its padding
```
    let padding = edge.pstyle(`${overlayOrUnderlay}-padding`).pfValue;
    let width = 2 * padding;
```
Result line width calculated either
edge's width if our outline is zero
or edgeWidth + lineOutlineWidth if lineOutlineWidth > 0. After commit [6ad45bf](https://github.com/cytoscape/cytoscape.js/pull/3244/commits/6ad45bf2acc3c3d899813176dd1e341aa46350cd) calculation changed to edgeWidth + lineOutlineWidth / 2.

Now when we set in demo edge's width to 3, and outline and the underlay are both set to 2, we will get the following values

underlay-width = 2 * 2 = 4
edge's width = 3 + 2 / 2 = 4

In this case we got equal values.
But if we change some of these values, we will no longer get a match, or we will get it by accident, or we will simply count the values ​​so that they coincide.

In my opinion underlay property serves to highlight a line. But line outline serves to show a line when line-color and the color of the background match. And dimensions of underlay should always be greater than edge's width with or not with outline.
Its up to user to control it.

I will be grateful for any advice on how to improve something in the PR.

mikekucera:
I think you're right. 
The total width of the line should be: line width + outline width, then the underlay padding should be in addition to the total width. That way to make the outline and underlay line up perfectly the underlay padding needs to be set to 0. 

It already works this way with just dashed line and underlay. Here the underlay padding is set to 0...

Screenshot 2024-05-17 at 9 15 51 AM

But if I turn on the outline, it should look like this...
The total edge width is greater, but the underlay still matches because the padding is 0.

Screenshot 2024-05-17 at 9 23 11 AM

What do you think, does that make sense?

xkdgo:
When underlay-padding is 0 I think it should be switched off ?
May be when we have dashed line we can see underlay padding but it swittched off. May be its a bug and not a feature ?

xkdgo:
> How about we treat the total width of the line as: line width + outline width

Could this confuse the user?
When user tuning outline, it simple to tune dimentions when he know width of his line, and he add some outline with color.

But of course it is possible to subtract the thickness of the outer line from the total line width. But in the library outline of nodes does not have same behaviour. We can check it when set outline-width : The size of the node’s outline.

mikekucera:
You're right again, node underlay padding does not appear to take the node border width or outline into consideration at all. I find this strange because the node's bounding box does appear to include the border. But in any case the way padding and width are related should be consistent with node borders.

I'm going to see if Max is available to weigh in on this. But so far it looks like your original PR is good. Thanks.

xkdgo:
@mikekucera before the next discussion I returned the calculation to the original. Thank you.

Issue #3273, opened 2024-09-10
Support mouseover events in Shadow DOM


FlorianHockmann:
**Description of new feature**

We are using Cytoscape.js in a React Web Component built with [`@r2wc/react-to-web-component`](https://github.com/bitovi/react-to-web-component). Now we want to configure a `mouseover` event, but it unfortunately does not fire.

Our first idea was that this is maybe something that can simply be handled different by `react-to-web-component` so we first asked on [their Discord server](https://discord.gg/J7ejFsZnJ4) for help. But they responded that the events don't fire from within the shadow DOM because of how Cytoscape registers the events.
Since the response included technical details and already a suggestion how this could be handled differently in Cytoscape, I simply quote it here:

> I looked into how the mouse events are handled in cytoscape and I don't think you'll have much success with using shadow DOM.  Let me explain the technical details:
> * Events that are generated within a Shadow DOM root can cross the barrier into the containing DOM, for most native event types.  "mouseover" events in cytoscape use "mousemove" under the hood, but this is a supported event type; on the technical level, the event's default properties include `composed: true`
> * However, when an event crosses the barrier from a Shadow DOM to the containing DOM, the event target is updated to the element containing the shadow root.  In your codesandbox case, once the mousemove bubbles up past the shadow root, its target becomes `<hello-wc>`
> * cytoscape sets all of these native event listeners up on the window, so they all see the component element as the target, and this becomes an issue when trying to determine when a mousemove is relevant to the cytoscape canvas.
> * mousemove events are discarded, and not turned into mouseover/mouseout events if the mousemove target is not a descendant of the container.  see https://github.com/cytoscape/cytoscape.js/blob/0a113f524d6e055eafcc43e13c1ebf80df64d6f1/src/extensions/renderer/base/load-listeners.js#L354
> * When the event leaves the shadow DOM, as stated before, its target becomes `<hello-wc>`.  This is an ancestor element, not a descendant element, of the `__________cytoscape_container` div that contains the canvases.  So mousemoves within the canvas are always discarded.
> * **The way to fix this, for cytoscape devs specifically, is to break out the event root from the window and make it configurable.  currently it's using `cy.window()` for several purposes, including choosing where to attach event listeners, and  accessing `getComputedStyle()`.  you can't just override `cy.window()` even with Proxy magic because the event handlers are set up before you get a reference to `cy`**

[emphasis added]

[...]

PR #3277, opened 2024-09-26
Attach the 'mousemove' handler to the window and the shadow DOM root.

mikekucera:
Associated issues: 

- #3273 

Notes re. the content of the pull request. Give context to reviewers or serve as a general record of the changes made. Add a screenshot or video to demonstrate your new feature, if possible.

- This PR attaches the 'mousemove' handler to both the window and the shadow DOM root of the cytoscape container if there is one.

The problem is described in detail in #3273, but I will recap here. The 'mousemove' handler is attached to the window, but it uses a function called `eventInContainer()` to test if the mouse is over the cytoscape container. The problem is that when the canvas is inside a shadow DOM the event.target field is retargeted to the shadow root, and the `eventInContainer()` function can't tell that the event was over the container. This prevents the 'mouseover' node event from ever firing.

I believe the reason the handler is attached to the window is so it can detect when a node is dragged beyond the bounds of the canvas. Like in this video...

https://github.com/user-attachments/assets/7e1f5391-72e9-4eb5-8bad-13198c8984e0

That means that making the event root for 'mousemove' configurable as suggested in the bug report will break this behaviour. (It would also add API just to handle this case which is undesirable.) Here is an example where I tried attacing the listener to the shadow root instead of the window.

https://github.com/user-attachments/assets/aeadf438-d05e-43bf-8671-3f129cd3396d

The solution I landed on is to test for a shadow root, and if there is one attach the listener to both the window and the shadow root. When the mouse moves over the canvas the listener attached to the shadow root is called, when it moves off the canvas the listener attached to the window is called. Since dragging is initiated by clicking on the canvas the dragging behaviour works correctly. This solution also doesn't require any API change.

**Checklist**

Author:

Reviewers:

### DIFF
--- a/src/extensions/renderer/base/load-listeners.js
+++ b/src/extensions/renderer/base/load-listeners.js
@@ -8,8 +8,20 @@ var BRp = {};
 
 BRp.registerBinding = function( target, event, handler, useCapture ){ // eslint-disable-line no-unused-vars
   var args = Array.prototype.slice.apply( arguments, [1] ); // copy
-  var b = this.binder( target );
 
+  if( Array.isArray(target) ){
+    let res = [];
+    for( var i = 0; i < target.length; i++ ){
+      let t = target[i];
+      if( t !== undefined ){
+        var b = this.binder( t );
+        res.push( b.on.apply( b, args ) );
+      }
+    }
+    return res;
+  }
+
+  var b = this.binder( target );
   return b.on.apply( b, args );
 };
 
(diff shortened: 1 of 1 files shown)


mikekucera:
Here is a test file that doesn't require any dependencies.
[test_shadow.html.zip](https://github.com/user-attachments/files/17165132/test_shadow.html.zip)

chrtannus (approved):
The `mouseover` event is fired when hovering a node in the canvas with the Shadow DOM. I also zoomed in and out a few times and dragged nodes around, which worked just fine.
I tested it on: Chrome, Safari and FF on macOS; Edge on Windows 11.

Issue #3251, opened 2024-07-01
Unbundled bezier edge disappear or flip when nodes are close


anthonyhtyao:
**Environment info**

- Cytoscape.js version : 3.30.0
- Browser/Node.js & version : firefox 126.0.1

**Current (buggy) behaviour**

Unbundled bezier edges disappear (2nd) or flip (3rd) when two nodes are next to each other or overlap as shown below
![Screenshot from 2024-07-01 12-21-19](https://github.com/cytoscape/cytoscape.js/assets/13106293/2ab3e1d8-3ae1-4ddc-9ac1-56301973f4c9)

p.s. there is an ancient issue about the flip [#411](https://github.com/cytoscape/cytoscape.js/issues/411) 

**Desired behaviour**

The second and third edges should display similarly to the first one

**Minimum steps to reproduce**

Here is the js code to reproduce the image above

```javascript
cytoscape({
    "container": document.getElementById("cy"),
     "elements": [
         {"data": {"id": 1}, "position": {"x": 0, "y": 0}},
         {"data": {"id": 2}, "position": {"x": 40, "y": 0}},
         {"data": {"id": 3}, "position": {"x": 0, "y": 40}},
         {"data": {"id": 4}, "position": {"x": 20, "y": 40}},
         {"data": {"id": 5}, "position": {"x": 0, "y": 80}},
         {"data": {"id": 6}, "position": {"x": 19, "y": 80}},
         {"data": {"source": 1, "target": 2}}, // normal
         {"data": {"source": 3, "target": 4}}, // disappear
         {"data": {"source": 5, "target": 6}}, // flip
    ],
    "style": [
         {"selector": "node", "style": {"width": 20, "height": 20}},
         {"selector": "edge", "style": {"curve-style": "unbundled-bezier", "control-point-weight": 0.5, "control-point-distance": 20}},
    ],
    "layout": {"name": "preset"},
})
```

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has included a reproducible demo.  They can easily fork this JSBin demo: http://jsbin.com/fiqugiq
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the corresponding branches.  Bug-fix patches go on
    - [ ] `master`,
    - [ ] `unstable`, and
[...]

PR #3280, opened 2024-09-30
Fix/3251

d2fong:
**Cross-references to related issues.**  If there is no existing issue that describes your bug or feature request, then [create an issue](https://github.com/cytoscape/cytoscape.js/issues/new/choose) before making your pull request.

Associated issues: 

- #3251

### Notes
- Some progress made on the flipping bezier/non existent bezier issue 
- Mostly looking for feedback on whether this is the right direction w.r.t fixing the bezier flipping/non existent bezier issue.
- It seems like the code producing the bug is in `src/extensions/renderer/base/coord-ele-math.js` specifically the `findEdgeControlPoints` function

##### Invisible bezier problem
When calculating the vector norm and vector norm inverse, it is possible for `dx` `dy` and `distance (l)` to be `0`, leading to the norm/norm inverse to be `NaN`.  I found another place in the edge rendering code that also mentions handling the cases of impossible beziers using a small constant value `0.01`.

##### Flipping bezier problem
I noticed that `dx` and `dy` could be negative.  Changing the computation to be absolute value changes the rendering to always make the bezier on one side.  There is probably nuance to this problem that I am missing.  

##### Testing
I _temporarily_ edited the debug init code to be the example provided in issue #3251.  Run the code by running `npm run watch` and toggling the various element json lines in `debug/init.js`

Author:

Reviewers:

### DIFF
--- a/debug/init.js
+++ b/debug/init.js
@@ -3,182 +3,48 @@
 
 var cy, defaultSty, options;
 
-(function(){
-  defaultSty  = cytoscape.stylesheet()
-
-      .selector('node')
-        .style({
-          'background-opacity': 0.4,
-          'label': 'data(id)',
-          'outline-offset': 5,
-          'outline-color': 'red',
-          'outline-opacity': 0.5,
-          'outline-width': 10,
-          'outline-style': 'solid',
-          'border-width': 5,
-          'border-opacity': 0.5,
-          'border-color': 'blue',
-          'border-position': 'inside',
-          'width': 220,
-          'height': 60
-        })
-      .selector('node#a')
-        .style({
-          'shape': 'round-rectangle',
-          'width': 220,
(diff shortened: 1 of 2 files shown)


mikekucera:
Hi Dylan,

The invisible bezier case from the test from issue #3251 is still happening for me. I took the test code posted by @anthonyhtyao and put it in an html file. The middle case with the invisible edge still happens. The flipping case is fixed. Here is the file if you want to take a look at it. Put it in the root of the project and run an http server. 
[test_edge_flip.html.zip](https://github.com/user-attachments/files/17210640/test_edge_flip.html.zip)

I think using small constants to avoid NaNs is a fine direction. Max would know better than me but he isn't available right now. But it looks good to me so far.

Thanks Dylan!

Issue #3270, opened 2024-08-27
Node label flickering when animating font-size


likeamike:
**Environment info**

- Cytoscape.js version : 3.30.2
- Browser/Node.js & version : Chrome Version 128.0.6613.85 (64-bit), Firefox 129.0.2 (64-bit), Node v20.13.1

**Current (buggy) behaviour**

_What does the bug do?_

When I try to add a hover effect to a node to enlarge the font size, the node's label blinks. Specifically, when I hover over one node, the label of another node or both nodes start blinking.

https://github.com/user-attachments/assets/a9eb0de2-eff2-4a3b-a634-2a0bcf9a4bb4

**Desired behaviour**

_What do you expect Cytoscape.js to do instead?_

Node labels should remain stable and not blink when hovering over another node.

**Minimum steps to reproduce**

_What do you need to do to reproduce the issue?_

1. Hover over the nodes.
2. If everything appears fine initially, refresh the page and hover over the nodes again.

The issue is reproduced here: https://jsbin.com/narayik/edit?html,css,js,output

_Fork/clone this JSBin demo and reproduce your issue so that your issue can be addressed quickly and effectively:_
http://jsbin.com/fiqugiq

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has included a reproducible demo.  They can easily fork this JSBin demo: http://jsbin.com/fiqugiq
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the corresponding branches.  Bug-fix patches go on
    - [ ] `master`,
    - [ ] `unstable`, and
    - [ ] the previous feature release branch (e.g. `1.1.x` if the current release is 1.2).
- [ ] The issue has been labelled as a [`bug`](https://github.com/cytoscape/cytoscape.js/labels/bug), if necessary.

mikekucera:
By "blinking" you are referring to the non-smooth resizing animation that sometimes happens?

mikekucera:
I'm still going to look deeper into this issue, but there appears to be a workaround that makes it at least somewhat better. The approach is to call the `stop()` function on the node before starting the new animation. I still get blinking sometimes, so there's still a bug, but it seems to happen much less often.

```javascript
  cy.on('mouseover', 'node', function(event) {
    var node = event.target;
  
    node.stop(); // stop current animation

    node.animate({
[...]

PR #3298, opened 2024-11-11
Fix: labelDimCache

maxkfranz:
**Cross-references to related issues.**  If there is no existing issue that describes your bug or feature request, then [create an issue](https://github.com/cytoscape/cytoscape.js/issues/new/choose) before making your pull request.

Associated issues: 

- #3270

- Work in progress
- Currently uses a single value

**Checklist**

Author:

Reviewers:

### DIFF
--- a/src/extensions/renderer/base/coord-ele-math/labels.js
+++ b/src/extensions/renderer/base/coord-ele-math/labels.js
@@ -471,14 +471,11 @@ BRp.calculateLabelDimensions = function( ele, text ){
 
   var document = containerWindow.document;
 
-  let cacheKey = util.hashString( text, ele._private.labelDimsKey );
+  let cacheKey = util.hashString(text, ele._private.labelDimsKey);
+  let cache = r.labelDimCache || (r.labelDimCache = { key: null, value: null });
 
-  let cache = r.labelDimCache || (r.labelDimCache = []);
-
-  let existingVal = cache[ cacheKey ];
-
-  if( existingVal != null ){
-    return existingVal;
+  if (cache.key === cacheKey) {
+    return cache.value;
   }
 
   let padding = 0; // add padding around text dims, as the measurement isn't that accurate
@@ -522,10 +519,13 @@ BRp.calculateLabelDimensions = function( ele, text ){
   width += padding;
   height += padding;
 
-  return ( cache[ cacheKey ] = {
+  cache.key = cacheKey;
+  cache.value = {
     width,
     height
-  } );
+  };
+
+  return cache.value;
 };
 
 


maxkfranz:
@likeamike, does this address #3270 for you?

maxkfranz:
@mikekucera, do you think this would be worth merging in for the next patch release?

I'd like to have it decided which PRs we're going to include so I can coordinate with the upcoming PR for the build system.  I don't want to deal with lots of merge conflicts, so applying or (reapplying) my changes on top of everything we want included would be best.  Thanks

maxkfranz:
Just a note: This issue might be more widespread than just the use case indicated in #3270.  If the old approach can generate invalid cache values for the label dimensions, then it could mess up all the label texture calculations for canvas and WebGL.  So I suspect it would be better to include this PR in the upcoming release, unless you have any reservations.

mikekucera:
Hi Max,

The labelDimCache used to have one entry for every label (when there's no animation). That way label dimensions are only calculated once per label. 
The proposed replaces the labelDimCache with an object with a single entry. Won't this result in label dimensions being recalculated when they used to be cached? 
I assumed that a fix for this issue would change how the cache keys are calculated as to avoid collisions.

maxkfranz:
Yes, that's a good point.

Here's the approach in the PR:

(1)
- Uses one cached value per element (32 bit int cache key).
- Collides erroneously only if the hash of the label for the element accidentally matches the previous label's hash for the same element.

The old approach:

(2)
- Uses a global cache.
- Grows unbounded over time.
- More likely to erroneously collide than (1).  Gets worse as the cache size grows.
- Uses one 32 bit int for the cache key per label.

Another approach:

(3)
- Uses a global cache like (2) but use two hash algorithms with two hash values.  The hash tuple (h1, h2) is used as the key.
- Uses two 32 bit ints for the cache key.
- Far less likely to erroneously collide than (2).
- Still has the problem of unbounded growth.  Would need something like an LRU cache.
- More complicated than (1).

An expensive approach:

(4) 
- Use a global cache.
- Use a string as the cache key, including the label itself.
- Won't collide but uses much more memory.
- Might still be cheaper than recalculating the bounds each frame, at least in terms of CPU.

What do you think?

mikekucera:
I think using one cached value per element in (1) would probably be the simplest and most reliable, even if it’s not the most efficient. (Do you even need a key if the bounds are stored on the element itself?) It has the disadvantage that if different nodes have the same label style then the bounds are recalculated for each instance, but I think that’s probably ok. The need is to avoid recalculating bounds on successive frames, not necessarily avoiding recalculating the same labels on the same frame. Lots of networks have unique labels for each node anyway.

I do think this solution is more addressing the symptom than the root cause. Hash keys shouldn’t be colliding so easily. The hash function should be producing keys with a uniform distribution. But changing the hashing is high risk and I shied away from it myself.

P.S. Edges actually need up to 3 labels right?

maxkfranz:
> I think using one cached value per element in (1) would probably be the simplest and most reliable, even if it’s not the most efficient. (Do you even need a key if the bounds are stored on the element itself?) It has the disadvantage that if different nodes have the same label style then the bounds are recalculated for each instance, but I think that’s probably ok. The need is to avoid recalculating bounds on successive frames, not necessarily avoiding recalculating the same labels on the same frame. Lots of networks have unique labels for each node anyway.

Looking at this a bit more, I think the simplest approach along the lines of (1) would be to remove the caching/hashing in this function altogether.

It looks like it may be a carryover from when we didn't have higher-level caching / rendered style invalidation.  Currently, this function is only called in two ways:

(a) For calculating the label dimensions for rendering.  These values are stored/cached per-element and used in higher-level functions.

(b) For calculating multi-line labels (wrapping, automatic ellipsis).

For the normal flow (a), removing the caching (i.e. labelDimCache) would only incur a performance penalty when the style of an element changes but the label happened to remain exactly the same in text and styling.

For multi-line calculations (b), you can't use a per-element cached value anyway.  You have to check based on the words.  I'm not sure caching the intermediate multi-line values would be worth the space/time trade-off either.

So what do you think about just removing the caching from this function (`BRp.calculateLabelDimensions`)?

> 
> I do think this solution is more addressing the symptom than the root cause. Hash keys shouldn’t be colliding so easily. The hash function should be producing keys with a uniform distribution. But changing the hashing is high risk and I shied away from it myself.

Yes.  The hash function also has to be cheap and work with the limitations of JS (e.g. bitwise ops treat a number as a 32-bit int).

> 
> P.S. Edges actually need up to 3 labels right?

Yes

maxkfranz:
Ref:

https://github.com/cytoscape/cytoscape.js/blob/5623dea7ea1935699aa976ab1c4d200a1f0ac83f/src/extensions/renderer/base/coord-ele-math/labels.js#L298-L320

Issue #3305, opened 2024-12-13
Initial, experimental WebGL rendering support


maxkfranz:
**Description of new feature**

_What should the new feature do?  For visual features, include an image/mockup of the expected output._

Add support for experimental rendering using WebGL.  Not all rendering features will be initially supported with WebGL, and improved WebGL support will be added over time.

**Note that until the WebGL rendering feature has been marked as final, the API and behaviour of the renderer will be subject to change.**

This initial version in 3.31.0 will be enabled via `options.webgl: true`.  It will support:

- Node styles
- Labels
- Straight edges
- Bezier edges
- Triangle arrows only

If you notice bugs or missing features that are important to you, file an issue on Github.

See also:

- #3261 

**Motivation for new feature**

_Describe your use case for this new feature._

- Improved rendering performance
- Getting community feedback for areas of improvement

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [x] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [x] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [x] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

mikekucera:
Work is happening in the `feature/gpu` branch.

pmxujiawei:
@maxkfranz maxkfranz When will version 3.31.0 be released?

Issue #3261, opened 2024-07-20
WebGPU Support


apowers313:
**Description of new feature**

networkx has an ability to [use GPUs to accelerate algorithms](https://docs.rapids.ai/api/cugraph/nightly/nx_cugraph/nx_cugraph/) resulting in [orders of magnitude faster processing time](https://developer.nvidia.com/blog/accelerating-networkx-on-nvidia-gpus-for-high-performance-graph-analytics/) for algorithms like centrality.

There is an emerging [WebGPU](https://www.w3.org/TR/webgpu/) browser standard with [some desktop implementations](https://caniuse.com/webgpu) that [enables the use of the GPU in the browser for parallelizing algorithms](https://surma.dev/things/webgpu/).

Just throwing it out there that it might be interesting for cytoscape.js to either integrate GPU acceleration or expose a backend plugin API [similar to networkx](https://networkx.org/documentation/stable/reference/backends.html) to enable others to do so.

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [ ] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

mikekucera:
I'm currently working on WebGL support full time, primarily to support large networks. WebGL is much more broadly adopted and works on most mobile browsers.

apowers313:
My understanding is that it's two different purposes. WebGL is for graphics and will help visualization run faster, especially with large networks. WebGPU can also do graphics (I'm not advocating that) but it can also be used as thousands of tiny CPUs to parallelize graph algorithms and make them run faster. For example, calculating betweenness centrality on a graph with 3M nodes and 17M edges takes 513 seconds to run on a CPU and 16 seconds to run on a GPU.

maxkfranz:
@apowers313, does your team currently have capacity to contribute in this direction?

kelciocasemiro:
Waiting for this...

apowers313:
[...]

PR #3314, opened 2025-01-06
Feature/gpu

mikekucera:
Associated issues: 

- #3305
- #3261

- This PR adds a WebGL rendering mode to the Canvas renderer. This improves the frame rate for large networks.
- The WebGL mode is opt-in. It must be enabled when creating the cy object...

```javascript
cy = cytoscape({
  container: document.getElementById('cy’),
  elements: [ /* … */ ],
  style: [ /* … */ ],

  renderer: {
    name: 'canvas',  // still uses the canvas renderer
    webgl: true, // turns on WebGL mode
    showFps: true,  // (optional) shows the current FPS at the top-left 
    webglDebug: true, // (optional) prints debug info to the browser console

    // additional options (provisional, may change in future releases)
    webglTexSize: 4096,
    webglTexRows: 24, 
    webglBatchSize: 2048,
    webglTexPerBatch: 16,
});
```

All node and label styles are fully supported. Nodes should look exactly the same with the WebGL renderer. The existing Canvas renderer is reused by the WebGL renderer. Nodes are rendered off-screen to a “sprite sheet” using the Canvas renderer and then used as textures by the WebGL renderer. This allows nodes to look the same while leveraging hardware acceleration. Each node only needs to be drawn to the sprite sheet once, then offloaded to video memory where it can be rendered repeatedly by the GPU. This approach also has the advantage that future contributions of new visual styles to the Canvas renderer should automatically work with the WebGL renderer as well.

Creating the sprite sheets does take some time when the network is first loaded. The preview release does not have a progress bar for this initial load time, so keep in mind that if your network doesn’t show up right away that just means it needs a few seconds to load.

Edge rendering has some limitations. It would not be advantageous to use the same sprite-sheet strategy for edges that is used for nodes. The strategy for this first release is to render straight edges as long skinny rectangles, and bezier edges as a sequence of straight segments that approximate a curve.

- Only straight-line, haystack and bezier edges are supported. Other less frequently used edge types such as taxi or segmented edges will show up as bezier edges.
- Dashed lines, overlays and underlays are not supported.
- Edges can only have one label in the centre. Source and target labels are not supported yet.
- Only triangle shaped edge arrows are supported. Other arrow shapes will show up as triangles. Hollow arrows are not supported.
- Only solid colours are supported, no colour gradients. The source and target arrows and the edge line can have different colours though. Opacity is fully supported.

**Checklist**

Author:

Reviewers:


maxkfranz (approved):
This looks great.  I'll just note this for later:  The current approach looks like it draws the overlays as a layer on top of everything else.  

Currently, in the canvas-only mode of the renderer, the overlays are drawn in the z-index order.  This doesn't make a difference for most use cases, but it might be different for cases like compound networks.

Some people use the overlays as a highlight for many elements, but most people just use it to show the active state (usually one element).

It's fine as-is, and it's a good trade-off -- but we might want to do something about it in future.  Some options for future:

- Note that WebGL handles overlays differently in the documentation.
- Change the WebGL approach to be like the canvas renderer.
- Change the canvas approach to be like the WebGL approach.

maxkfranz:
@mikekucera, looks good to merge

maxkfranz:
@d2fong. Hi, Dylan. Just wanted to bring this to your attention re. the overlays. Whatever approach we take in future could be informed by what works best for Cytoscape Web. Does/will Cytoscape Web use overlays for things like highlighting/selections/search?

mikekucera:
> The current approach looks like it draws the overlays as a layer on top of everything else.

I'm pretty sure that's not the case. Why do you think that?

It draws elements in z-index order, and for each node it draws underlay, body, label, overlay in that order. The code is here: 

https://github.com/cytoscape/cytoscape.js/blob/ba83c4ec3abc63cb1ccd53c9671ec990536ee802/src/extensions/renderer/canvas/webgl/drawing-redraw-webgl.js#L450C9-L453C14

maxkfranz:
Not enough sleep.  My mistake

mikekucera:
No worries. Merging now.

Issue #3320, opened 2025-01-15
display and selection problems with GPU support


KonradHoeffner:
As described in <https://github.com/cytoscape/cytoscape.js/pull/3314>, I activated webgl as follows using the newest Cytoscape.js 3.31.0 and the WebGL mode successfully activates but all hidden edges and node labels are still visible. I also cannot select using mouse clicks (multi-select using SHIFT works) and I also cannot move nodes. I am not using features unsupported by GPU such as background color gradient, edge source labels or non-straight edges.

```js
        cy = cytoscape({
            container: [...],
            style:  [...],
            renderer: {
                name: "canvas",
                webgl: true,
                showFps: true,
                webglDebug: true,
            },
```
## Without WebGL with hidden nodes, this is how it should look, live on https://www.snik.eu/graph/

![Screenshot From 2025-01-15 13-37-32](https://github.com/user-attachments/assets/79e8c2fa-8911-452e-aca7-e532f363d418)

## Without WebGL with the hidden nodes shown
![image](https://github.com/user-attachments/assets/452ed376-4960-4bac-bd62-5b0acf25141d)

## With WebGL, live on https://www.snik.eu/gpu/
![Screenshot From 2025-01-15 13-36-52](https://github.com/user-attachments/assets/cf0acd3a-64a9-4af8-98e1-a27fb1375f7e)

The code is in <https://github.com/snikproject/graph/>, master and gpu branch respectively.

KonradHoeffner:
Sorry, it seems there was a caching issue, closing this for now.

maxkfranz:
@mikekucera, do you think it would be a good idea to have some feedback in the debug console for the WebGL feature, e.g. 


```js
console.warn('Provisional WebGL support enabled.  Please be advised that this feature may have stability issues and its API is subject to change... and so on...');
```

KonradHoeffner:
@mikekucera: Thanks for fixing this! Can you create a new release so that this is usable? Or is it possible to install the current master branch of cytoscape.js over npm?

mikekucera:
Hi @KonradHoeffner, 
I'm working on improving the memory usage of the webgl rendrer. I should have a PR ready this week. Then we can discuss doing a point release. Generally we are pretty quick with these releases.
You can install packages directly from github using npm. However, the cytoscape.js build artifacts are in the /dist folder and I don't think that gets updated on every commit. So installing from the unstable branch might not work.

KonradHoeffner:
[...]

PR #3321, opened 2025-01-15
fix webgl mode showing hidden nodes/edges

mikekucera:
Associated issues: 

- #3320

- Fixes issue where the webgl mode was still displaying nodes/edges that had the `visibility:hidden` property.
- Selection seems to work fine. I think it wasn't working for the nodes that were supposed to be hidden.

**Checklist**

Author:

Reviewers:

### DIFF
--- a/src/extensions/renderer/canvas/webgl/atlas.js
+++ b/src/extensions/renderer/canvas/webgl/atlas.js
@@ -528,7 +528,7 @@ export class AtlasManager {
     }
   }
 
-  isRenderable(ele, type) {
+  isVisible(ele, type) {
     const opts = this.getRenderTypeOpts(type);
     return opts && opts.isVisible(ele);
   }
--- a/src/extensions/renderer/canvas/webgl/defaults.js
+++ b/src/extensions/renderer/canvas/webgl/defaults.js
@@ -12,6 +12,6 @@ export const renderDefaults = defaults({
   getRotation: null,
   getRotationPoint: null,
   getRotationOffset: null,
-  isVisible: null,
+  isVisible: () => true,  // this is an extra check for visibility in addition to ele.visible()
   getPadding: null,
 });
--- a/src/extensions/renderer/canvas/webgl/drawing-elements-webgl.js
+++ b/src/extensions/renderer/canvas/webgl/drawing-elements-webgl.js
@@ -377,7 +377,7 @@ export class ElementDrawingWebGL {
 
   drawTexture(ele, eleIndex, type) {
     const { atlasManager } = this;
-    if(!atlasManager.isRenderable(ele, type)) {
+    if(!ele.visible() || !atlasManager.isVisible(ele, type)) {
       return;
     }
     if(!atlasManager.canAddToCurrentBatch(ele, type)) {
@@ -434,6 +434,9 @@ export class ElementDrawingWebGL {
 
 
   drawEdgeArrow(edge, eleIndex, prefix) {
+    if(!edge.visible()) {
+      return;
+    }
     // Edge points and arrow angles etc are calculated by the base renderer and cached in the rscratch object.
     const rs = edge._private.rscratch;
 
@@ -453,6 +456,7 @@ export class ElementDrawingWebGL {
       return; 
     }
 
+    // check shape after the x/y check because pstyle() is a bit slow
     const arrowShape = edge.pstyle(prefix + '-arrow-shape').value;
     if(arrowShape === 'none' ) {
       return; 
@@ -504,14 +508,20 @@ export class ElementDrawingWebGL {
 
 
   drawEdgeLine(edge, eleIndex) {
+    if(!edge.visible()) {
+      return;
+    }
+    const points = this.getEdgePoints(edge);
+    if(!points) {
+      return;
+    }
+
     // line style
     const baseOpacity = edge.pstyle('opacity').value;
     const lineOpacity = edge.pstyle('line-opacity').value;
     const width = edge.pstyle('width').pfValue;
     const color = edge.pstyle('line-color').value;
     const opacity = baseOpacity * lineOpacity;
-
-    const points = this.getEdgePoints(edge);
 
     if(points.length/2 + this.instanceCount > this.maxInstances) {
       this.endBatch();
@@ -591,6 +601,11 @@ export class ElementDrawingWebGL {
   
   getEdgePoints(edge) {
     const rs = edge._private.rscratch;
+
+    // if bezier ctrl pts can not be calculated, then die
+    if( rs.badLine || rs.allpts == null || isNaN(rs.allpts[0]) ){ // isNaN in case edge is impossible and browser bugs (e.g. safari)
+      return;
+    }
     const controlPoints = rs.allpts;
     if(controlPoints.length == 4) {
       return controlPoints;
--- a/src/extensions/renderer/canvas/webgl/drawing-redraw-webgl.js
+++ b/src/extensions/renderer/canvas/webgl/drawing-redraw-webgl.js
@@ -1,6 +1,3 @@
-// import { EdgeDrawing } from './drawing-edges-webgl';
-// import { EdgeBezierDrawing } from './drawing-edges-curved-webgl';
-// import { NodeDrawing } from './drawing-nodes-webgl';
 import { ElementDrawingWebGL } from './drawing-elements-webgl';
 import { RENDER_TARGET, renderDefaults } from './defaults';
 import { OverlayUnderlayRenderer } from './drawing-overlay';
@@ -50,16 +47,6 @@ CRp.initWebgl = function(opts, fns) {
     return label && label.value;
   };
 
-  // r.edgeDrawing = new EdgeBezierDrawing(r, gl, opts, renderDefaults({
-  //   getKey: fns.getLabelKey,
-  //   getBoundingBox: fns.getLabelBox,
-  //   drawElement: fns.drawLabel,
-  //   getRotation: getLabelRotation,
-  //   getRotationPoint: fns.getLabelRotationPoint,
-  //   getRotationOffset: fns.getLabelRotationOffset,
-  //   isVisible: isLabelVisible,
-  // }));
-
   r.eleDrawing = new ElementDrawingWebGL(r, gl, opts);
   const our = new OverlayUnderlayRenderer(r);
   
@@ -67,7 +54,6 @@ CRp.initWebgl = function(opts, fns) {
     getKey: fns.getStyleKey,
     getBoundingBox: fns.getElementBox,
     drawElement: fns.drawElement,
-    isVisible: ele => ele.visible(),
   }));
 
   r.eleDrawing.addTextureRenderType('node-label', renderDefaults({


mikekucera:
Hi @KonradHoeffner,

I created a branch with a fix: `fix/3320`

Could you please verify that the fix works for you. 
- check out the cytoscape.js fix/3320 branch
- run `npm install`
- run `npm run build`
- run `npm run dist:copy`
- change the cytoscape dependency in your package.json to `"cytoscape": "file:../cytoscape.js"`

I already tested it locally with your repository, but I want to make sure everything is working as you expect it to.

@maxkfranz Is there a better way for bug reporters to test a fix before we merge it in?

Thanks.

maxkfranz:
> Hi @KonradHoeffner,
> 
> I created a branch with a fix: `fix/3320`
> 
> Could you please verify that the fix works for you.
> 
> * check out the cytoscape.js fix/3320 branch
> * run `npm install`
> * run `npm run build`
> * run `npm run dist:copy`
> * change the cytoscape dependency in your package.json to `"cytoscape": "file:../cytoscape.js"`
> 
> I already tested it locally with your repository, but I want to make sure everything is working as you expect it to.
> 
> @maxkfranz Is there a better way for bug reporters to test a fix before we merge it in?
> 
> Thanks.

Not really.  

I suppose you could attach a zip of the build dir as an attachment here so bug reporters don't have to build.  They could overwrite the files in `node_modules/cytoscape/dist` in their project with your zip.  If a bug reporter is already using npm for their project (i.e. most people), it's not hard for them to do a build themselves.  Six of one.

maxkfranz:
@mikekucera

Hi, Mike.  The code looks fine to me.  Would you like me to test this out?

mikekucera:
> Hi, Mike. The code looks fine to me. Would you like me to test this out?

No, just want to double check with the bug poster, if they respond.

KonradHoeffner:
@mikekucera: Thank you for the fast response and the fix! Your instructions are very helpful, the only thing I struggled a bit was the package.json change, because I first pointed it directly at the dist folder and the various files, before I understood that I need to use the parent folder and npm by itself is finding the dist folder.

I'm currently trying it with my laptop with only 8 GB of RAM where Firefox always crashes with `webgl: true` due to being out of memory. When webgl is set to false, the OS only uses 2.5-3 GB in total (Vite in dev mode + Firefox).

KonradHoeffner:
P.S.: On my other PC I also needed `npm install --force` to ignore the conflicting peer deps.

KonradHoeffner:
On the desktop PC I can now run it successfully, there the PC is using 7 GB of RAM in total with webgl set to true and 3.33 GB with webgl set to false. When it does not crash due to being out of memory the performance is vastly improved as the FPS more than doubles!

However I am still unable to select anything, neither node nor edge. Maybe it doesn't work for left handers?


|PC|resolution|webgl|RAM|FPS|
|----|---------------|---------|------|--------|
|Intel Core i9 12900k|4k|true|~7 GB|42|
|Intel Core i9 12900k|4k|false|3.3 GB|16|
|Intel Core i3 1115G4|full HD|true|OOM|-|
|Intel Core i3 1115G4|full HD|false|||

mikekucera:
Hi @KonradHoeffner,

Thanks for the feedback, its very helpful.

The way the WebGL renderer works is that it creates textures for the node bodies and labels. Each texture object holds several node bodies. I believe its the textures that are using the memory. There are options to control the size of the textures that might help.

```javascript
cy = cytoscape({
  ...
  renderer: {
    name: 'canvas',  // still uses the canvas renderer
    webgl: true, // turns on WebGL mode
    showFps: true,  // (optional) shows the current FPS at the top-left 
    webglDebug: true, // (optional) prints debug info to the browser console

    // additional options (provisional, may change in future releases)
    webglTexSize: 4096, 
    webglTexRows: 24, 
    webglBatchSize: 2048,
    webglTexPerBatch: 16,
});
```

The options `webglTexSize` and `webglTexRows` can be used to adjust how big textures are. 
- Lowering `webglTexSize` to 1024 or 2048 will reduce the size of each texture. 
- `webglTexRows` controls how many node bodies fit into each texture. Using a higher value will fit more nodes into each texture which reduces the total number of textures.

But if you make the textures too small then nodes will start to look blurry. You will have to try tuning it to your application.

We will have to figure out a way to manage memory better in the WebGL renderer, but that might take a while.

mikekucera:
I'm on a Macbook and I don't have any issues selecting or dragging nodes in SNIK. I've tried in chrome and firefox and it works fine. I may need to get access to a PC to test it there. 

How do I view the entire network including the hidden nodes?

In the meantime I could add another option to turn off the WebGL click detection and fall back to the canvas renderer. We are using WebGL for click detection because its faster, but if there's issues then maybe there needs to be an option to turn it off. I'll look into it.

mikekucera:
BTW, if you're interested in saving the network as a PDF then take a look at this extension: https://github.com/cytoscape/cytoscape.js-pdf-export

KonradHoeffner:
> P.S. How do I view the entire network including the hidden nodes?

For measuring the FPS and RAM I first hit CTRL+ALT+R (or Layout->Reset Main Graph) to unhide the hidden nodes and then CTRL+ALT+L (Layout -> Recalculate Layout) and then zoom out a bit and constantly move the graph by click-holding the background and moving the mouse around a lot.

KonradHoeffner:
> We will have to figure out a way to manage memory better in the WebGL renderer, but that might take a while.

Thank you for explaining the options, I will try how much I can lower RAM consumption with them! I'm trying it with my third computer right now which has an Intel i9 12900k but no dedicated GPU and the RAM consumption is even high with over 14 GB, this is definitely more than our target audience is guaranteed to have.

KonradHoeffner:
Updated the table with more tested parameters. Increasing webglTexRows to 256 or more did make it a bit blurry but significantly reduced RAM consumption. Also tried various higher values for webglTexPerBatch but FPS didn't seem to change.

KonradHoeffner:
@mikekucera: Is the large texture requirement maybe due to each node being saved separately? As we only have 3 different node shapes (* a few different colors), would it be possible to save CPU/GPU RAM by reusing the shapes?

KonradHoeffner:
BTW Chromium fares way better in both RAM consumption and FPS.

mikekucera:
> @mikekucera: Is the large texture requirement maybe due to each node being saved separately? As we only have 3 different node shapes (* a few different colors), would it be possible to save CPU/GPU RAM by reusing the shapes?

It creates textures for both node bodies and labels. I does reuse textures when it can, so the node bodies are probably being reused. I think its the labels that are using the texture memory.

KonradHoeffner:
>  I think its the labels that are using the texture memory.

Could the library save RAM but saving a texture for each letter of the alphabet? I would imagine that 26 (or even 100 or more if you count lowercase + uppercase and so on) characters would need a fraction of the RAM compared to having several thousand textures of sharp text.

mikekucera:
I'm not sure how far we could get with that approach. Labels can have backgrounds and outlines and other fancy formatting. (eg https://js.cytoscape.org/demos/linkout-example/). And we would have to consider things like kerning. I think it would get very complicated.

KonradHoeffner:
Maybe there could be a simple/performance mode and a fancy mode?
At least for our use case, fancy font features like kerning and ligatures are not necessary.
I'm not an OpenGL expert and there seem to be multiple common methods but most of them seem to have in common that each letter or "glyph" is saved separately.

<https://stackoverflow.com/a/36065835/398963>:

> Texture atlas
> We could raster characters for every frame and re-create the textures, but that is not very efficient, specially if characters have a fixed size.
> The more efficient approach is to raster all characters you plan on using and cram them on a single texture.
> And then transfer that to the GPU once, and use it texture with custom uv coordinates to choose the right character.
> This approach is called a [texture atlas](https://en.wikipedia.org/wiki/Texture_atlas) and it can be used not only for textures but also other repeatedly used textures, like tiles in a 2D game or web UI icons.

More stuff I found:

* this YouTube lecture series seems relevant: <https://youtube.com/playlist?list=PLEMXAbCVnmY6zCgpCFlgggRkrp0tpWfrn>
* https://faultlore.com/blah/text-hates-you/
* detailed explanation last updated in 2019 https://stackoverflow.com/a/5278471/398963

maxkfranz:
> We will have to figure out a way to manage memory better in the WebGL renderer, but that might take a while.

I don't remember: Are we using the node body style hashes to lookup the textures for WebGL, or does the WebGL code put every node in the texture atlas?  

Using the node style hashes reduces memory usage when many nodes are styled the same.  A lot of networks have the same node style for many nodes, e.g. SBGN.

We definitely shouldn't be reimplementing text rendering using WebGL, but node body optimisations would generally be more useful and tractable  in future.

maxkfranz:
Other future node body enhancements in future might include things like:

Key the texture atlas for nodes by { general body style key ignoring colour, include aspect ratio but not size } => texture.  Then you get more hits.  Node bodies could be tinted with the colour at draw time, and similarly they could be scaled at draw time as long as the aspect ratio is the same.

maxkfranz:
@KonradHoeffner I'll also note that the best way to optimise labels is to not draw so many in the first place.  For large networks, you don't need to see all the labels when you're zoomed out since they're too small to read anyway. The `min-zoomed-font-size` property is the current  way to use this, but I don't know if we support this in WebGL.  You can always implement something similar in your app by only setting the labels when you're zoomed in far enough to see them.

Generally, the WebGL renderer isn't feature-complete yet, and we'd probably focus on getting missing features in (e.g. more edge types) before adding all of the possible optimisations.  We'll have to find a balance between the two going forward.

maxkfranz:
Sorry to spam the thread with so many comments in a row.  I just wanted to get down some thoughts about WebGL so that they're written down for future reference

maxkfranz (approved):
👍

mikekucera:
The WebGL renderer uses the same style keys as the ElementTextureCache. It actually passes the same functions that generate the keys to the WebGL renderer:
https://github.com/cytoscape/cytoscape.js/blob/unstable/src/extensions/renderer/canvas/index.js#L282-L283

I don't think the zoom level is a problem once the textures are uploaded to the gpu. The mipmaps should handle that with good performance. 

I'm currently doing some research into using compressed texture formats. I think (and I'm just guessing right now) that the main issue is that the textures are uncompressed. I can see on my machine that for the example EM-web network the "Google Chrome Helper (GPU)" process uses between 3-4 GB, for only 17 textures, which seems like a lot. BTW This 3-4GB does not show up in the chrome dev tools memory tab, which is why I never realized how much memory was getting used.

WebGL supports _loading_ compressed textures. It does not support doing the actual compression. It seems like the standard approach is to load texture files that have been pre-compressed, which makes sense for a game, but we are generating textures at runtime using offscreen canvases, so I don't know how to do the compression. This MDN page suggests to use a library called Basis Universal, which I'm looking into, but it seems kind of complicated: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#consider_compressed_texture_formats

Will keep you posted.

KonradHoeffner:
> but we are generating textures at runtime using offscreen canvases, so I don't know how to do the compression. This MDN page suggests to use a library called Basis Universal, which I'm looking into, but it seems kind of complicated: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#consider_compressed_texture_formats

At least in our application we don't target mobile users, so just desktop would be enough, where WEBGL_compressed_texture_s3tc seems to be the best supported one.

KonradHoeffner:
P.S.: Are textures even needed at all? At least in our case the nodes are all basic primitives (triangle, square, circle) with a solid color.
Admittedly my OpenGL knowledge is very outdated (we had an OpenGL exercise in 2005) but we build a whole application using colored polygons with no textures at all.

mikekucera:
One of the reasons cytoscape.js is popular is because of its excellent support for styling. The idea with the WebGL mode was to support as many of the existing styles as possible. By using textures for nodes and labels we are able to support all node and label styles. There are currently limitations on edge styles in the WebGL mode, but ideally we would be able to figure out how to remove those limitations in the future.

There could be a use for an "ultra-performance" mode, where nodes are just simple polygons and labels are limited to simple fonts with no extra styling. But the idea with the current approach is to hopefully address the needs of the greatest number of users. We also need to support some specific applications that we are developing that allow for large range of styles to be applied.

I will look into pixijs, I'm curious how they handle (or avoid) these issues with textures.

mikekucera:
Some initial findings...

I used this library to find out how much memory each texture uses: [greggman / webgl-memory](https://github.com/greggman/webgl-memory)

For a texture of node labels from the EM-web network set to 4096x4096, it uses 89mb of video memory.

I then saved the texture as an image file and used Basis Universal from the command line to compress it. The resulting file was about 650k.

That's a huge difference. How exactly this would translate to video memory savings I can't know for sure. The overhead of compressing at runtime could also be an issue. Using a compression library would introduce dependencies. But theoretically, this looks like it could be a big help if we can get it to work.

maxkfranz:
Let's split off the discussion for future performance enhancements: https://github.com/cytoscape/cytoscape.js/discussions/3325

This PR is not going to be visible long-term after it's merged in.

mikekucera:
The original bug with hidden nodes being visible has been fixed. I've merged this PR. Further work on optimizing memory usage will happen in another branch.

Issue #3337, opened 2025-02-26
WebGL: clicking a label to select doesn't work the same as canvas


mikekucera:
**Desired behaviour**

When `text-events: yes` is enabled clicking anywhere within the bounding box of a label will select the node (or click-hold to drag). 

For performance reasons the WebGL renderer uses its own logic for handling hit detection (ie picking).

In the WebGL mode if `text-events: yes` is set and you click on any of the transparent space between letters it won't detect the click. If `text-events: no` is set and you click on the text of a label it will register the click.

**Minimum steps to reproduce**

Steps are for developers...
- Open the file `debug/webgl/network-styles.json`. Add `"text-events": "yes"` to the "node" style section.
- Open the WebGL debug page (/debug/webgl/index.html).
- Enable the WebGL mode (should be on by default)
- Open the "Style Test" network.
- Click on the "Lorem Ipsem" label somewhere in the whitespace between letters. 
  - Notice how the node isn't selected.

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has included a reproducible demo.  They can easily fork this JSBin demo: http://jsbin.com/fiqugiq
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the corresponding branches.  Bug-fix patches go on
    - [ ] `master` and
    - [ ] `unstable`.
- [ ] The issue has been labelled as a [`bug`](https://github.com/cytoscape/cytoscape.js/labels/bug), if necessary.

maxkfranz:
I suppose for some events like click, we could fall back on the expensive canvas method.

For some events like mousemove or mouseover, the canvas method is really expensive for large networks.

maxkfranz:
> In the WebGL mode if you click on any of the transparent space between letters it won't detect the click. This is kind of strange and is inconsistent with how the canvas renderer works.

I suppose as a hacky workaround, we could use a nearly-transparent (e.g. opacity 0.00001) background box on the labels.  If the pixels aren't perfectly white/transparent, then it should allow for the first phase of the hit detection passing, right?

mikekucera:
Yes I think that would work, thanks for the idea.

PR #3346, opened 2025-03-04
WebGL: fix label picking

mikekucera:
- This PR fixes bug #3337 
- Unfortunately the idea to set the background opacity of labels to 0.0001 did not work. I don't know why, but it would not detect the clicks when I tried this. Maybe something in WebGL ignores small alpha values. I'm not sure.
- Instead I took an approach of reusing the "simple shapes" mode that draws rectangles. If `text-events: no` then it will ignore the label, if `text-events: yes` then it will render the bounding box of the label as an opaque rectangle that can be clicked on. Of course you can't see this because its happening in an offscreen framebuffer.
  - I didn't try measuring but this may actually be more performant. It doesn't have to render label textures anymore for picking. Labels are either ignored or rendered as rectangles.

**Checklist**

Author:

Reviewers:

### DIFF
--- a/debug/webgl/network-styles.json
+++ b/debug/webgl/network-styles.json
@@ -64,7 +64,8 @@
         "background-gradient-stop-colors": "cyan magenta yellow",
         "text-valign": "top",
         "text-rotation": 0.7853981633974483,
-        "label": "n3 - Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod"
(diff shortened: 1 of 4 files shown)


maxkfranz (approved):
Works well

I tried with multiline labels and it worked for that too.

Unrelated: The label became blurry when it was multiline:

Screenshot 2025-03-05 at 11 15 38

mikekucera:
Thanks.

> Unrelated: The label became blurry when it was multiline:

I'll open a bug for this. Multiline labels should have a separate texture for each line.

Issue #3406, opened 2025-08-12
WebGL: group selection in bounding boxes broken


svozza:
**Environment info**

- Cytoscape.js version : 3.33.0
- Browser/Node.js & version : Node 22

**Current (buggy) behaviour**

I have a diagram where nodes are contained in multiple bounding boxes, when I click on some (but not all) of the bounding boxes the nodes in the bounding box do not get selected.

**Desired behaviour**

When I click on a bounding box, all the nodes in the box should get selected. When I disable WebGL, everything works as expected. 

**Minimum steps to reproduce**

_What do you need to do to reproduce the issue?_

1. Create a graph with several grouped nodes, i.e., parent child relationships.
2. Attempt to select any of the groups.
3. The select event does not appear to register

See the below screen recordings.

WebGL enabled:

![Image](https://github.com/user-attachments/assets/d4219a87-05cc-486e-9ea2-fa414f9c7a96)

WebGL disabled

![Image](https://github.com/user-attachments/assets/ddadb484-5931-4fb3-8c4f-e9e04c8e1a7e)

Literally the only difference between these two examples are that I have set the WebGL boolean:

```js
 // Create a new Cytoscape instance with conditional WebGL rendering options
        const cy = cytoscape({
            container: containerRef.current,
            elements: [], // Start with empty graph
            style: graphStyle,
            renderer: {
                name: 'canvas',
                webgl: false, // <= changing this to true causes the bug
            },
        });
```

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has included a reproducible demo.  They can easily fork this JSBin demo: http://jsbin.com/fiqugiq
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the corresponding branches.  Bug-fix patches go on
    - [ ] `master` and
    - [ ] `unstable`.
- [ ] The issue has been labelled as a [`bug`](https://github.com/cytoscape/cytoscape.js/labels/bug), if necessary.

mikekucera:
I did a quick diagnosis for this bug. The issue is not unique to compound nodes, it happens for all nodes that have `background-opacity: 0`. Clicking the background does not select the node, but clicking the border will. The solution for the WebGL renderer is probably to just ignore the `background-opacity` property when in `RENDER_TARGET.PICKING` mode. I should have a PR soon.

[...]

PR #3410, opened 2025-08-21
fix: webgl: selection of nodes with background-opacity: 0

mikekucera:
Associated issues: 

- #3406

- When the WebGL renderer is in `PICKING` mode, if a node passes the visibility check its `background-opacity` is ignored and always set to 1. This ensures the node is still clickable/selectable.

Note: I did not thoroughly test this. Its a good idea to do some extra testing before merging.

**Checklist**

Author:

Reviewers:

### DIFF
--- a/debug/webgl/network-compound-nodes.json
+++ b/debug/webgl/network-compound-nodes.json
@@ -16,7 +16,8 @@
         "text-halign": "center",
         "shape": "round-rectangle",
         "corner-radius": "10",
-        "padding": 10
+        "padding": 10,
+        "background-opacity": 0.0
       }
     },
     {
--- a/debug/webgl/network-styles.json
+++ b/debug/webgl/network-styles.json
@@ -79,7 +79,7 @@
         "border-color": "black",
         "border-style": "dashed",
         "border-width": 2,
-        "background-opacity": 0.5
+        "background-opacity": 0.0
       }
     },
     {


mikekucera:
This is actually tricker to fix than I first thought. The issue is that the webgl PICKING mode tests if a pixel has `alpha = 0` to determine if an element exists at that pixel. That means if `alpha = 0` for any reason, clicking on that pixel will not register as a "hit". 

If a node body cannot be rendered directly by webgl as a `"simple shape"` (ie `ellipse`, `rectangle`, `roundrectangle` or `bottomroundrectangle` with solid border and nothing else) then it will fall back on the canvas renderer to render it as a texture. When this happens the webgl renderer basically hands control over to the canvas renderer which could theoretically render anything it wants to the texture. There's no way to know if a pixel in the texture with `alpha = 0` is because of a transparent background, an opacity gradient, a weirdly shaped polygon, or even a background image that has transparent pixels in the image file. In other words, I don't know how to easily fix this in a way that will always work for all corner cases. We would need to rethink how picking mode works.

As a temporary mitigation I made a couple simple changes to relax what styles count as a "simple shape" when in webgl PICKING mode. We already have something like this for node labels, that allows hit detection to work anywhere inside the bounding box of the label. Using the bounding box won't work for node bodies (I won't explain why here) so just adding a check for corner cases that are easy to handle is what i've ladened on.
- When in picking mode ignore border styles, background images, and gradients when testing for a "simple shape", and then render the simple shape with an opaque background.
- Unfortunately a bunch of supported node shapes currently won't work with this strategy. These are: `diamond`, `pentagon`, `hexagon`, `heptagon`, `octagon`, `star`, `vee`, `rhomboid`, `polygon`. These always get rendered as a texture.
- Its possible to add SDF functions for all theses shapes so they could be treated as "simple shapes", however that would be a lot of work and is outside the scope of this PR... and it will never work for `polygon`.

Issue #3408, opened 2025-08-15
Font display screw up when specify fallback font family


Rey-Wang:
**Environment info**

- Cytoscape.js version : 3.33.0
- Browser/Node.js & version : Chrome Version 138.0.7204.49 (Official Build) (arm64)

**Current (buggy) behaviour**
When specify a fallback family, the font display is not collapsed

<img width="584" height="145" alt="Image" src="https://github.com/user-attachments/assets/1ffd58d7-28e3-4f48-92d8-7b5165fc9d06" />

**Desired behaviour**
Should display different font family correctly

**Minimum steps to reproduce**
When change the font-family style from `'font-family': 'Roboto Mono'` to be `'font-family': 'Roboto Mono, Inter'`

I also notice this issue happen on another issue, even its few years ago

Please check this simple demo. The issue occurs on initial page load, so I suspect it’s related to lazy font loading.
https://codesandbox.io/p/sandbox/bai-frontend-react-hfpnt

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has included a reproducible demo.  They can easily fork this JSBin demo: http://jsbin.com/fiqugiq
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the corresponding branches.  Bug-fix patches go on
    - [ ] `master` and
    - [ ] `unstable`.
- [ ] The issue has been labelled as a [`bug`](https://github.com/cytoscape/cytoscape.js/labels/bug), if necessary.

slava-class:
We hit this same class of bug downstream in an Electron/Cytoscape.js graph. The useful detail was that `<link rel="preload">` and hosted font assets were not enough: Cytoscape was still being instantiated before the FontFaceSet had resolved, so canvas text measurement/layout used fallback metrics on initial render.

The workaround that fixed it was to gate `cytoscape(...)` creation on the actual font readiness contract:

```js
const graphFonts = [
  '400 16px "Your Graph Font", sans-serif',
  '700 16px "Your Graph Font", sans-serif',
];

await Promise.all([
  ...graphFonts.map((font) => document.fonts.load(font)),
  document.fonts.ready,
]);

if (!graphFonts.every((font) => document.fonts.check(font))) {
  throw new Error("Graph fonts failed to reach ready state.");
}

const cy = cytoscape({ container, elements, style });
```

[...]

PR #3418, opened 2025-10-14
Feat: Improve the precision of label bounding box calculations and rendering

### LINKED ISSUE #3417: Improve the accuracy of label bounding box calculations and rendering

Improves the precision of label bounding box calculations and rendering by introducing new style properties and refining the handling of label metrics.

Currently, different fonts introduce extra space for hanging characters (e.g., 'g', 'p', 'j', 'q'), which expands the bounding box unnecessarily. This additional space makes it difficult to select labels when overlaps occur. The extra space becomes quite substantial when the text is rotated, making the bounding box even larger and more difficult to manage.
Here is an example:

Image

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has adequately described their idea.  If not, elicit more information about the use case.  You should iteratively build a spec together.
- [ ] Ensure that the issue is a good fit for the core library.  Some things are best done in extensions (e.g. UI-related features that aren't style-related).  Some things are best done by app authors themselves -- instead of in Cytoscape libraries.
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the `unstable` branch via pull request.  The corresponding pull request is cross-referenced.

fzaman-linz:
This PR improves the precision of label bounding box calculations and rendering by introducing new style properties and refining label metrics handling.

- `text-metrics (font | glyph)`
Allows to choose between browser-default text metrics or actual font metrics (actualBoundingBoxAscent / Descent) for more accurate label height calculations.

- Label Dimension Calculation Enhancements
`calculateLabelDimensions()` now supports `text-metrics: glyph`, using `actualBoundingBoxAscent` and `actualBoundingBoxDescent` for more precise vertical alignment and height.

Associated issues: 

- #3417 

**Checklist**

Author:

- [x] For new or updated API, the `index.d.ts` Typescript definition file has been appropriately updated.

Reviewers:

- [x] All automated checks are passing (green check next to latest commit).
- [x] At least one reviewer has signed off on the pull request.
- [x] For bug fixes:  Just after this pull request is merged, it should be applied to both the `master` branch and the `unstable` branch.  Normally, this just requires cherry-picking the corresponding merge commit from `master` to `unstable` -- or vice versa.

### DIFF
--- a/src/collection/dimensions/bounds.mjs
+++ b/src/collection/dimensions/bounds.mjs
@@ -291,7 +291,8 @@ let updateBoundsFromLabel = function( bounds, ele, prefix ){
     let borderWidth = ele.pstyle( 'text-border-width' ).pfValue;
     let halfBorderWidth = borderWidth / 2;
     let padding = ele.pstyle( 'text-background-padding' ).pfValue;
-    let marginOfError = 2; // expand to work around browser dimension inaccuracies
+    let marginOfErrorX = ele.pstyle( 'bbox-margin-error-x' ).pfValue; // expand to work around browser dimension inaccuracies
+    let marginOfErrorY = ele.pstyle( 'bbox-margin-error-y' ).pfValue; // expand to work around browser dimension inaccuracies
 
     let lh = labelHeight;
     let lw = labelWidth;
@@ -341,10 +342,10 @@ let updateBoundsFromLabel = function( bounds, ele, prefix ){
     }
 
     // shift by margin and expand by outline and border
-    let leftPad  = marginX - Math.max( outlineWidth, halfBorderWidth ) - padding - marginOfError;
-    let rightPad = marginX + Math.max( outlineWidth, halfBorderWidth ) + padding + marginOfError;
-    let topPad   = marginY - Math.max( outlineWidth, halfBorderWidth ) - padding - marginOfError;
-    let botPad   = marginY + Math.max( outlineWidth, halfBorderWidth ) + padding + marginOfError;
+    let leftPad  = marginX - Math.max( outlineWidth, halfBorderWidth ) - padding - marginOfErrorX;
+    let rightPad = marginX + Math.max( outlineWidth, halfBorderWidth ) + padding + marginOfErrorX;
+    let topPad   = marginY - Math.max( outlineWidth, halfBorderWidth ) - padding - marginOfErrorY;
+    let botPad   = marginY + Math.max( outlineWidth, halfBorderWidth ) + padding + marginOfErrorY;
 
     lx1 += leftPad;
     lx2 += rightPad;
--- a/src/collection/dimensions/bounds.mjs
+++ b/src/collection/dimensions/bounds.mjs
@@ -291,8 +291,7 @@ let updateBoundsFromLabel = function( bounds, ele, prefix ){
     let borderWidth = ele.pstyle( 'text-border-width' ).pfValue;
     let halfBorderWidth = borderWidth / 2;
     let padding = ele.pstyle( 'text-background-padding' ).pfValue;
-    let marginOfErrorX = ele.pstyle( 'bbox-margin-error-x' ).pfValue; // expand to work around browser dimension inaccuracies
-    let marginOfErrorY = ele.pstyle( 'bbox-margin-error-y' ).pfValue; // expand to work around browser dimension inaccuracies
+    let marginOfError = 2; // expand to work around browser dimension inaccuracies
 
     let lh = labelHeight;
     let lw = labelWidth;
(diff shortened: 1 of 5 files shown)


fzaman-linz:
Hi guys, could you please have a look

> cc: @maxkfranz @mikekucera

fzaman-linz:
Thanks for the feedback @mikekucera.
I've made the changes you suggested and here is a demo: https://codepen.io/fzaman-linz/pen/pvEYvxj
cc: @d2fong @maxkfranz

maxkfranz:
> Screenshot 2026-02-16 at 2 26 10 PM

What happens if you have a label like 'apple' instead of just 'a'?

If you have several nodes like the above image, the text (baseline) on all the labels should line up.  That's important.

Try these labels.  They should all line up as long as the label positions are the same and the node dimensions are the same:

- 'yes' (descender)
- 'no' (no ascender nor descender)
- 'not' (ascender)

fzaman-linz:
@maxkfranz Yes, label text baselines won't line up across the board. Tight bounding boxes need per‑word heights (which shifts baselines), while consistent baselines need fixed heights (which reintroduces the gaps we're trying to avoid). That's why keeping `text-metrics: default | actual` makes sense so it leaves the choice with the user.
Are you ok with that?
image

maxkfranz:
@fzaman-linz, the team will discuss this on a call on Thursday 23 April.

maxkfranz:
@fzaman-linz, in the meantime, would you please update the OP with the current status of the feature (i.e. only currently proposed props)?

fzaman-linz:
@maxkfranz I've updated the OP and also updated the example URL to showcase default vs actual.
https://codepen.io/fzaman-linz/pen/pvEYvxj
image

maxkfranz:
Great. Thanks

maxkfranz:
The team is discussing this feature and the details of the properties etc., and we'll have an update some time next week.

src/collection/dimensions/bounds.mjs
```diff
@@ -291,7 +291,8 @@ let updateBoundsFromLabel = function( bounds, ele, prefix ){
     let borderWidth = ele.pstyle( 'text-border-width' ).pfValue;
     let halfBorderWidth = borderWidth / 2;
     let padding = ele.pstyle( 'text-background-padding' ).pfValue;
-    let marginOfError = 2; // expand to work around browser dimension inaccuracies
+    let marginOfErrorX = ele.pstyle( 'bbox-margin-error-x' ).pfValue; // expand to work around browser dimension inaccuracies
```
maxkfranz: Are larger values than 2 needed for specific cases?

src/style/properties.mjs
```diff
@@ -91,6 +91,7 @@ const styfn = {};
     valign: { enums: [ 'top', 'center', 'bottom' ] },
     halign: { enums: [ 'left', 'center', 'right' ] },
     justification: { enums: [ 'left', 'center', 'right', 'auto' ] },
+    textMetrics: { enums: [ 'default', 'actual' ] },
```
maxkfranz: These values should be more descriptive so they're obvious as to what they do.  E.g. 'aligned' | 'tight'.  Or whatever names would be clear.

src/extensions/renderer/base/coord-ele-math/labels.mjs
```diff
@@ -312,14 +312,14 @@ BRp.applyPrefixedLabelDimensions = function( ele, prefix ){
 
   let labelDims = this.calculateLabelDimensions( ele, text );
   let lineHeight = ele.pstyle('line-height').pfValue;
+  let size = ele.pstyle('font-size').pfValue;
```
maxkfranz, line 315: Could you upload some screenshots of some line-height:1 multiline labels?  We should verify that this change does not cause regressions.  There are likely edge cases that `normPerLineHeight` was handling.

src/collection/dimensions/bounds.mjs
maxkfranz: From @mikekucera : 

> Remove the bbox-margin-error-x / bbox-margin-error-y properties, the default of 2 is fine. Its not worth it to add two VPs just for this.

src/style/properties.mjs
fzaman-linz: How about text-metrics: `font` | `glyph`, or `auto` | `precise`? 
Default would be `font` / `auto`, but happy to go with whatever's preferred.

src/collection/dimensions/bounds.mjs
fzaman-linz: Fair, will remove bbox-margin-error-x/y, keep the hardcoded 2. The tradeoff isn't worth it. Once we decide on the `text-metrics` arguments name I'll update it.

src/extensions/renderer/base/coord-ele-math/labels.mjs
fzaman-linz: https://codepen.io/fzaman-linz/pen/pvEYvxj
image

src/style/properties.mjs
maxkfranz: font | glyph sounds intuitive/clear

src/extensions/renderer/base/coord-ele-math/labels.mjs
maxkfranz: Are the blue multiline cases comparable to the current prod version?

fzaman-linz: Yes, they're comparable to the current prod version. Also I've made the suggested changes.

slava-class:
One downstream datapoint that may be useful while testing this PR: we recently hit a separate initial-render label issue that looked like a Cytoscape label metrics bug, but the root cause was webfont readiness. Cytoscape was created before custom fonts had loaded, so the initial canvas label metrics/bounds were based on fallback fonts; a later style refresh made the labels correct.

For any demo/test cases here that use webfonts, it may be worth loading/checking the font before creating the graph. Preload was not sufficient in our case; we had to wait for `document.fonts.load(...)` plus `document.fonts.ready`, check the exact label font descriptors with `document.fonts.check(...)`, then instantiate Cytoscape. I added the concrete snippet on the directly related issue: https://github.com/cytoscape/cytoscape.js/issues/3408#issuecomment-4412270646

maxkfranz (approved):

PR #3465, opened 2026-05-20
`eles.withinBox()` and `eles.hit()` for hit test use cases

fzaman-linz:
Cytoscape currently only exposes axis-aligned bounding boxes (AABB) for labels. This does not work well for rotated labels, which require polygon-based bounds for accurate hit-testing and intersection detection.
These features are already implemented as part of Cytoscape's box selection, but the functions were private and not accessible to users.
This PR exposes them as part of the public API.

- **withinBox(box)** - filter elements by position inside a rectangular box
- **labelsContainPoint(point)** - hit-test which nodes' label polygons contain a point
- **polygonIntersection(polygon)** - filter elements whose label bounds intersect a given polygon (using SAT)

Associated issues: 

- #3466

- This PR adds this.
- This PR does that.
- This PR allows us to do some other thing.

**Checklist**

Author:

Reviewers:

## Summary by CodeRabbit

* **New Features**
  * Added APIs to retrieve precise polygon bounds for main, source, and target labels, including rotated labels.
  * Added collection filtering by rectangular regions and polygon intersections.
  * Added position-based hit testing for element bodies and main, source, or target labels.
  * Added touch-aware hit testing and configurable hit-test behavior.
  * Added TypeScript declarations for the new geometry, filtering, and hit-testing APIs.

### DIFF
--- a/src/collection/dimensions/bounds.mjs
+++ b/src/collection/dimensions/bounds.mjs
@@ -27,8 +27,8 @@ elesfn.renderedBoundingBox = function( options ){
   };
 };
 
-elesfn.renderedActualLabelBoundingBox = function(){
-  let polygon = this.actualLabelBoundingBox();
+elesfn.renderedActualLabelBoundingBox = function( options ){
+  let polygon = this.actualLabelBoundingBox( options );
   let cy = this.cy();
   let zoom = cy.zoom();
   let pan = cy.pan();
@@ -39,36 +39,45 @@ elesfn.renderedActualLabelBoundingBox = function(){
   }));
 };
 
-elesfn.actualLabelBoundingBox = function() {
-  let bounds;
+elesfn.actualLabelBoundingBox = function( options ) {
+  let label = (options && options.label) || 'main';
+  let prefix = label === 'main' ? undefined : label; // 'source' | 'target' | undefined
 
   let isDirty = memoize(ele => {
     let _p = ele._private;
+    let cache = _p.labelPolygonCache;
 
-    return _p.labelPolygonCache == null || _p.styleDirty || _p.labelPolygonPosKey !== getLabelPolygonPosKey(ele);
+    return cache == null || cache[label] == null || _p.styleDirty || _p.labelPolygonPosKey !== getLabelPolygonPosKey(ele);
   }, ele => ele.id());
 
   if (this.length === 1 && !isDirty(this[0])) {
-    bounds = this[0]._private.labelPolygonCache;
-  } else {
-    let eles = this;
-    let cy = eles.cy();
-    let styleEnabled = cy.styleEnabled();
-    this.edges().forEach(isDirty);
-    this.nodes().forEach(isDirty);
+    return this[0]._private.labelPolygonCache[label];
+  }
 
-    if(styleEnabled) {
-      this.recalculateRenderedStyle();
-    }
+  let eles = this;
+  let cy = eles.cy();
+  let styleEnabled = cy.styleEnabled();
+  this.edges().forEach(isDirty);
+  this.nodes().forEach(isDirty);
 
-    for (let i = 0; i < eles.length; i++) {
-      let ele = eles[i];
-      let _p = ele._private;
-      let polygon = getRotatedLabelBox(ele);
-      _p.labelPolygonCache = polygon;
+  if(styleEnabled) {
+    this.recalculateRenderedStyle();
+  }
+
+  let bounds;
+
+  for (let i = 0; i < eles.length; i++) {
+    let ele = eles[i];
+    let _p = ele._private;
+
+    if( _p.labelPolygonCache == null || _p.labelPolygonPosKey !== getLabelPolygonPosKey(ele) ){
+      _p.labelPolygonCache = {};
       _p.labelPolygonPosKey = getLabelPolygonPosKey(ele);
-      bounds = polygon;
     }
+
+    let polygon = getRotatedLabelBox(ele, prefix);
+    _p.labelPolygonCache[label] = polygon;
+    bounds = polygon;
   }
 
   return bounds;
@@ -990,14 +999,18 @@ let cachedBoundingBoxImpl = function( ele, opts ){
 };
 
 let getLabelPolygonPosKey = function(ele) {
-  let pos = ele.position();
   let angle = ele.pstyle('text-rotation').pfValue || 0;
 
-  return hashIntsArray([
-    Math.round(pos.x),
-    Math.round(pos.y),
-    Math.round(angle * 1000) 
-  ]);
+  if( ele.isEdge() ){
+    let p1 = ele.source().position();
+    let p2 = ele.target().position();
+
+    return hashIntsArray([ Math.round(p1.x), Math.round(p1.y), Math.round(p2.x), Math.round(p2.y), Math.round(angle * 1000) ]);
+  } else {
+    let pos = ele.position();
+
+    return hashIntsArray([ Math.round(pos.x), Math.round(pos.y), Math.round(angle * 1000) ]);
+  }
 };
 
 let defBbOpts = {
@@ -1100,7 +1113,7 @@ elesfn.dirtyBoundingBoxCache = function(){
     _p.arrowBounds.target = null;
     _p.arrowBounds['mid-source'] = null;
     _p.arrowBounds['mid-target'] = null;
-    _p.labelPolygonCache = null;
+    _p.labelPolygonCache = {};
     _p.labelPolygonPosKey = null;
   }
 
--- a/src/collection/dimensions/bounds.mjs
+++ b/src/collection/dimensions/bounds.mjs
@@ -27,6 +27,53 @@ elesfn.renderedBoundingBox = function( options ){
   };
 };
 
+elesfn.renderedActualLabelBoundingBox = function(){
+  let polygon = this.actualLabelBoundingBox();
+  let cy = this.cy();
+  let zoom = cy.zoom();
+  let pan = cy.pan();
+
+  return polygon.map(p => ({
+    x: p.x * zoom + pan.x,
+    y: p.y * zoom + pan.y
+  }));
+};
+
+elesfn.actualLabelBoundingBox = function() {
+  let bounds;
+
+  let isDirty = memoize(ele => {
+    let _p = ele._private;
+
+    return _p.labelPolygonCache == null || _p.styleDirty || _p.labelPolygonPosKey !== getLabelPolygonPosKey(ele);
+  }, ele => ele.id());
+
+  if (this.length === 1 && !isDirty(this[0])) {
+    bounds = this[0]._private.labelPolygonCache;
+  } else {
+    let eles = this;
+    let cy = eles.cy();
+    let styleEnabled = cy.styleEnabled();
+    this.edges().forEach(isDirty);
+    this.nodes().forEach(isDirty);
+
+    if(styleEnabled) {
+      this.recalculateRenderedStyle();
+    }
+
+    for (let i = 0; i < eles.length; i++) {
+      let ele = eles[i];
+      let _p = ele._private;
+      let polygon = getRotatedLabelBox(ele);
+      _p.labelPolygonCache = polygon;
+      _p.labelPolygonPosKey = getLabelPolygonPosKey(ele);
+      bounds = polygon;
+    }
+  }
+
+  return bounds;
+};
+
 elesfn.dirtyCompoundBoundsCache = function(silent = false){
   let cy = this.cy();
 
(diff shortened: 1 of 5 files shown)


maxkfranz:
After a quick initial look:  One potential point of consideration is that edges can have up to three labels.  How would this PR account for that?  Could you specify a particular one with these new functions?  Are there opportunities to match some of the options or API patterns to `eles.boundingBox(options)`?

fzaman-linz:
Good point on edges, updated to address this.

actualLabelBoundingBox() and renderedActualLabelBoundingBox() now accept { label: 'main' | 'source' | 'target' } option (defaulting to 'main')

On aligning with `boundingBox(options)`: considered includeMainLabels / includeSourceLabels / includeTargetLabels but since this function returns a single polygon rather than a merged bbox, label selector felt more appropriate

src/collection/filter.mjs
```diff
@@ -378,7 +379,65 @@ let elesfn = ({
       value: min,
       ele: minEle
     };
-  }
+  },
+
+  withinBox: function(box) {
+    let eles = this;
+    const { x1, y1, x2, y2 } = box;
```
maxkfranz: You can't assume the box has x2,y2.  It might have w,h.  Normalise with the math function that already exists.

src/collection/filter.mjs
```diff
@@ -378,7 +379,65 @@ let elesfn = ({
...
+
+      if (pointInsidePolygonPoints(point.x, point.y, flatPoints)) {
+        inside.push(ele);
+      }
+    }
+
+    return this.spawn(inside);
+  },
+
+  polygonIntersection: function(polygon) {
```
maxkfranz, line 404: This only considers labels.

maxkfranz:
What's the use case for these added functions?  In hindsight, the library is already handling hit tests for you so doing your own seems to be redundant.

fzaman-linz:
> What's the use case for these added functions? In hindsight, the library is already handling hit tests for you so doing your own seems to be redundant.

@maxkfranz `labelsContainPoint` would be really useful for stacked labels selection which native AABB not able to. See this example:
https://codepen.io/fzaman-linz/pen/pvNdrZQ (try hovering "Hello World")
There are other use cases, but they all relate to improvements when dealing with rotated labels, where AABB is a nightmare to work with.


labelContainsPoint

src/collection/filter.mjs
fzaman-linz: updated

src/collection/filter.mjs
fzaman-linz: updated

maxkfranz:
Two points of feedback, upon further consideration:

- The withinBox function could be generally useful, but it only consider nodes and only the node positions (not dimensions).
- Rather than essentially duplicating this logic, it would probably be best to just have two functions: `eles.withinBox(box)` and `eles.hit(pos, options)`.  These functions can just proxy out to the corresponding renderer functions that already do hit tests and box selection logic.

So we'd have:

- eles.withinBox(box): returns subset of collection that is within the box (as per box selection rules)
- eles.hit(pos, options): returns subset of collection that passes the hit test (or maybe just the topmost one)
  - pos: the position {x,y} of the hit test
  - returned eles: probably sorted by z-index/depth for order of hit test, or maybe just return the topmost one as per the renderer hit test
  - options: same sort of options as eles.bb(), includeBody, includeSourceLabel, ... that kind of thing

Maybe eles.withinBox would need eles.withinBox(box, options) in case different use cases need different options

fzaman-linz:
@maxkfranz Thanks I've added both as suggested:

- eles.withinBox(box): proxies to the renderer's getAllInBox
- eles.hit(pos, options):  returns all matching elements sorted topmost-first (not just the topmost) coz we need the full ranked list
   - options: includeBody, includeMainLabels, includeSourceLabels, includeTargetLabels, isTouch

src/collection/filter.mjs
```diff
@@ -378,7 +379,64 @@ let elesfn = ({
...
+
+  withinBox: function(box) {
+    const cy = this.cy();
+    const r = cy.renderer();
+    if( cy.styleEnabled() ){ this.recalculateRenderedStyle(); }
+    const x1 = box.x1 != null ? box.x1 : box.x;
+    const y1 = box.y1 != null ? box.y1 : box.y;
+    const x2 = box.x2 != null ? box.x2 : x1 + box.w;
+    const y2 = box.y2 != null ? box.y2 : y1 + box.h;
+    const allInBox = r.getAllInBox(x1, y1, x2, y2);
```
maxkfranz, line 392: Not sure whether the null renderer has a simplified version of this for headless mode.

src/collection/filter.mjs
```diff
@@ -378,7 +379,64 @@ let elesfn = ({
...
+    const col = this;
+    return this.spawn( allInBox.filter(ele => col.has(ele)) );
+  },
+
+  hit: function(pos, options) {
+    const cy = this.cy();
+    const r = cy.renderer();
+    if( cy.styleEnabled() ){ this.recalculateRenderedStyle(); }
+    return this.spawn( r.hitTestAt(pos.x, pos.y, this, options) );
```
maxkfranz, line 401: Same as above re. null renderer

src/collection/dimensions/bounds.mjs
```diff
@@ -28,6 +27,62 @@ elesfn.renderedBoundingBox = function( options ){
   };
 };
 
+elesfn.renderedActualLabelBoundingBox = function( options ){
```
maxkfranz: Are these functions necessary now that we have `eles.withinBox()` and `eles.hit()`?

src/extensions/renderer/base/coord-ele-math/coords.mjs
```diff
@@ -72,6 +72,111 @@ BRp.invalidateContainerClientCoordsCache = function(){
...
+    ];
+    return math.pointInsidePolygonPoints( x, y, points );
+  } else {
+    return math.inBoundingBox( bb, x, y );
+  }
+};
+
+// Returns all elements from eles that are hit by (x, y), sorted topmost first.
+// options: { includeBody, includeMainLabels, includeSourceLabels, includeTargetLabels, isTouch }
+BRp.hitTestAt = function( x, y, eles, options ){
```
maxkfranz, line 123: We should be reusing `r.findNearestElements()` (refactoring findNearestElements a little bit if needed)

maxkfranz (changes requested):
Looks pretty good.  Reuse (and possibly minor refactoring of r.findNearestElements) is important.

We can also remove the unneeded functions given that `eles.hit()` and `eles.withinBox()` should suffice for the use case you mentioned.

fzaman-linz:
@maxkfranz I've:
- added `getAllInBox` and `hitTestAt` stubs to the null renderer
- refactored checkLabel to use the new `BRp.checkLabelHitAt` directly. `hitTestAt` uses the same
- removed `renderedActualLabelBoundingBox`, `actualLabelBoundingBox` used by polygonIntersection so we need that
- updated hit() options all defaulting to true.

maxkfranz:
Low hanging fruit from a Claude review.  All of the points from Claude are fair.

-----

## PR Review: `eles.hit()`, `eles.withinBox()`, `eles.polygonIntersection()`

Overall this adds three useful public spatial-query methods plus a supporting `actualLabelBoundingBox()` and a renderer hit-test refactor. The `findNearestElements` label-check refactor (extracting `checkLabelHitAt`) is a genuine improvement. But there's substantial **duplicated geometry code**, a couple of **real correctness bugs**, **dead code**, and **no tests**. Details below, roughly ordered by severity.

---

### 🔴 Correctness bugs

**1. `polygonIntersection` silently skips node bodies (`bodyBounds` may be null).**
`src/collection/filter.mjs:416` reads `ele._private.bodyBounds` directly, but `bodyBounds` is populated only inside `cachedBoundingBoxImpl` during `boundingBox()` (`bounds.mjs:834`) — **not** by `recalculateRenderedStyle()`. On elements whose `boundingBox()` hasn't been computed yet, `bodyBounds` is `null`, so the body check is skipped and only the label is tested. Note the canvas renderer does it correctly: `let getElementBox = ele => { ele.boundingBox(); return ele[0]._private.bodyBounds; }` (`canvas/index.mjs:152`). Fix: call `ele.boundingBox()` before reading `bodyBounds` (or use the bbox directly).

**2. Dead null-guard + NaN polygon in the new `getRotatedLabelBox` (bounds.mjs).**
```js
let bb = bbs[bbPrefix] = bbs[bbPrefix] || {};
if (!bb) { return null; }   // can never be true — bb is always an object
```
`bb` is either the existing bounds or a freshly created `{}`, so the `if (!bb)` guard (and its large explanatory comment) is dead code. Worse, when `labelBounds[prefix]` was null, this now (a) mutates `_p.labelBounds` with an empty object and (b) computes `lx1 = bb.x1 - th - ox` = `undefined - … ` = **NaN**, returning a polygon of NaN points instead of `null`. Compare the sibling copy in `getAllInBox` which does `var bb = _p.labelBounds[prefix||'main']; if (!bb) return null;` — that one is correct. This copy diverged and broke.

**3. `actualLabelBoundingBox` returns only the *last* element's polygon for multi-element collections.**
`bounds = polygon` is reassigned each loop iteration, so a collection of N elements returns only `eles[N-1]`'s box (the rest are merely cached). Every other `*BoundingBox` method on collections aggregates/unions. Either restrict the API to singular (and document/return for `this[0]`), or aggregate — but returning "the last one" is a footgun.

---

### 🟠 Duplicated code (the main theme of the request)

The rotated-label-polygon construction (`bb ± th ± margin`, then rotate by `theta`) now exists in **three** near-identical copies:

- `BRp.checkLabelHitAt` (new, `coords.mjs:75`)
- `getRotatedLabelBox` local to `BRp.getAllInBox` (`coords.mjs:398`, pre-existing)
- `getRotatedLabelBox` (new, `bounds.mjs:272`)

These should be unified into a single shared helper (e.g. in `math.mjs` or a label-geometry util) returning the 4-point polygon, with the hit-testers and the bounding-box getter all consuming it. As-is, they've already drifted (the `text-events`/`label.value` gating in `checkLabelHitAt` vs. `label.strValue` and no gate in `bounds.mjs`; the dead null-guard in one and correct guard in another), which is exactly the maintenance hazard duplication causes.

Additionally, **node-body hit logic is duplicated**: `hitTestAt` (`coords.mjs`) re-implements the same `outerWidth/outerHeight + threshold → bb check → shape.checkPoint` block that already lives in `findNearestElements`'s `checkNode` (`coords.mjs:236`). The PR refactored the *label* check to share but not the *node* check. Worth extracting a `checkNodeHitAt` analogous to `checkLabelHitAt`.

---

### 🟡 Design / consistency

**4. Naming.** `actualLabelBoundingBox` returns a polygon (`Position[]`), not a `BoundingBox` (`{x1,y1,x2,y2,w,h}`). The name collides conceptually with `boundingBox`/`renderedBoundingBox` while returning a different shape, and "actual" is unexplained. Consider `labelPolygon()` / `boundingPolygon()`. (The d.ts type alias `PolygonBoundingBox = Position[]` has the same naming tension.)

**5. Zoom-dependent value baked into a "model coordinate" geometry.** `getRotatedLabelBox` (bounds) adds `th = 2/zoom` to the polygon. That hit-test affordance makes a supposedly model-space geometry getter shift with zoom level — surprising for `actualLabelBoundingBox`, which the d.ts documents as "in model coordinates." The threshold belongs in the hit-testers, not the geometry getter.

**6. `polygonIntersection` doc says "nodes" but tests all elements.** The d.ts comment says "containing nodes whose polygonal bounds intersect," but the loop iterates the whole collection including edges (edges have `bodyBounds` too). Align doc and behavior, or `this.nodes()`-filter.

**7. `hit` iterates the entire z-sorted graph regardless of collection size.** `hitTestAt` walks all of `getCachedZSortedEles()` and filters by an id `Set`. For `someSmallCollection.hit(pos)` on a large graph this is O(graph), not O(collection). Fine for whole-graph use; worth a note/comment, or sort the (usually smaller) input collection by z instead.

**8. `withinBox`/`hit` only ever return interactive elements** (via `getAllInBox`/`getCachedZSortedEles().interactive` and box-selection style rules). That's defensible given the "renderer's box-selection rules" doc, but should be called out explicitly in the d.ts so callers aren't surprised when `display:none`/`box-selection:none` elements are missing.

---

### 🟢 Minor / style

- **No tests.** Three new public API methods + `actualLabelBoundingBox` and zero `test/` changes. These are very testable (construct a graph, assert membership for hit/box/polygon cases incl. rotated labels). Please add coverage.
- `getLabelPolygonPosKey(ele)` is recomputed up to 3× per element per call (memoized `isDirty`, the cache-invalidation `if`, and the assignment). Cache it in a local.
- `getRotatedLabelBox` in `bounds.mjs` mixes `var` and `let` within one function; the surrounding `bounds.mjs` uses `let`. The `coords.mjs` copies use `var`. Pick one per file.
- `actualLabelBoundingBox` returns `undefined` when headless (`getRotatedLabelBox` early-returns), while the cache stores `undefined`. Confirm the `=== null` cache-validity checks behave with `undefined` vs `null` (the `cache[label] == null` check covers both, OK, but `bounds = polygon` then returns `undefined`).
- Trailing whitespace / double blank line added at top of `bounds.mjs:30`.

---

### Summary

The feature set is reasonable and the `checkLabelHitAt` extraction is a step in the right direction, but I'd ask for another pass before merge:

1. Fix the `bodyBounds`-null skip in `polygonIntersection` (real bug).
2. Fix the dead guard / NaN-polygon path in the new `getRotatedLabelBox`.
3. Decide and fix multi-element semantics of `actualLabelBoundingBox`.
4. Unify the 3 copies of rotated-label-box construction (and ideally the node-body check) into shared helpers.
5. Add tests.
6. Reconsider naming and the zoom-dependent threshold in the geometry getter.

Want me to draft the unified label-geometry helper and refactor the three call sites, or post these as inline PR comments?

fzaman-linz:
@maxkfranz I've addressed both Claude's and CodeRabbit's comments and updated the PR. Plz take another look. Also could this be included in an upcoming milestone, so we have a rough idea of when it might be available for us. Cheers!

Issue #3336, opened 2025-02-26
Improved wheel-zoom support for mice/OSes that report inaccurate/imprecise scroll wheel deltas


ubmarco:
**Environment info**

- Cytoscape.js version : 3.31.0
- Browser/Node.js & version : On Arch Linux x64 XFCE, Firefox, Chromium, Brave, Edge

**Current (buggy) behaviour**

Zoom too fast or too much on a single mouse wheel event.

https://github.com/user-attachments/assets/aab88383-8f46-4223-a93a-dc6320e1d4a2

Browsers:
```
Chromium   Brave
Firefox    Edge
```

**Desired behaviour**

Cross OS and cross browser smooth zoom behavior.

**Minimum steps to reproduce**

1. Go to https://cytoscape.org/cytoscape.js-dagre/
2. Move the mouse wheel to zoom the image

When inserting
```js
window.addEventListener("wheel", event => console.info(event.deltaY));
```
into the browser console, I get values of +120 and -120 or +132 and -132 on the different browsers.
A user on Mac just gets values between -2 and +2 and smooth zooming.

Changing the Firefox setting `mousewheel.system_scroll_override.vertical.factor` from the default of 200 to 20 helps. It makes the overall scrolling super slow, so no solution.

Changing https://js.cytoscape.org/#init-opts/wheelSensitivity helps, but it only helps on one system and breaks others.
I think the code should normalize the `deltaY` value of the `wheel` event.

**For reviewers**

_Reviewers should ensure that the following tasks are carried out for incorporated issues:_

- [ ] Ensure that the reporter has included a reproducible demo.  They can easily fork this JSBin demo: http://jsbin.com/fiqugiq
- [ ] The issue has been associated with a corresponding [milestone](https://github.com/cytoscape/cytoscape.js/milestones).
- [ ] The commits have been incorporated into the corresponding branches.  Bug-fix patches go on
    - [ ] `master` and
    - [ ] `unstable`.
- [ ] The issue has been labelled as a [`bug`](https://github.com/cytoscape/cytoscape.js/labels/bug), if necessary.

ubmarco:
Good question. I never implemented a zoom feature in JS. I'd peek into other packages that handles zoom differently.
(I don't want to advertise other solutions. I just want to take a look to see how things can be improved.)

E.g. the zoom for visjs or G6 works good on my machine:
https://visjs.github.io/vis-network/examples/network/layout/hierarchicalLayout.html
[...]

PR #3491, opened 2026-07-23
Use GCD of initial wheel deltas as inaccurate scroll factor

fedarko:
Hi, and thanks as always for the amazing library! I found a very minor bug (which, arguably, is more of a problem with mouse events on my system than anything else). This PR addresses this bug on my system, and should keep behavior the same for people with other systems.

## Environment info

- Cytoscape.js version: 3.34.0
- OS: Linux (Ubuntu 22.04.5, using Xfce)
- Browser: Firefox 152.0.4 (but, from testing, this issue also impacts Chromium)
- USB Mouse: Logitech MX Ergo S (but, from testing, this issue also impacts another USB mouse -- and even the touchpad on my laptop)

### Background

I have been using Cytoscape.js 3.31.4 on Linux for a while, and I've noticed that -- when using a USB mouse to scroll in a Cytoscape.js drawing -- the scrolling is very "jerky." This is presumably the same issue as #3336.

So, I updated to Cytoscape.js 3.34.0, which works very nicely! It makes scrolling with the mouse smoother, aside from one strange corner case:

### The bug

I noticed that, if the first couple scrolls that I do in a Cytoscape.js instance are very big / fast, then the scrolling from then on in the interface becomes very jerky -- like before. However, if the first couple scrolls are slower, then the scrolling from then on in the interface is smooth and nice.

### Screen recordings

(These use the [Dagre demo](https://cytoscape.org/cytoscape.js-dagre/), like #3336.)

#### Fast initial scrolling (results in future scrolling being jerky)

(This GIF involves a lot of fast-moving images by its nature, so [I am just going to link it here](https://github.com/user-attachments/assets/a47c513a-efab-40c2-bd4a-0d8e35f8804a) instead of pasting it directly here to avoid triggering people with photosensitivity.)

#### Slow initial scrolling (results in future scrolling being smooth)

Image

## Desired behaviour

Ideally, even fast initial scrolling would still result in smooth scrolling.

### Diagnosing the bug

I tried adding the following log statement to [right after this line in the Cytoscape source code](https://github.com/cytoscape/cytoscape.js/blob/9ebf7f0999e5763e89531a8a1b0ee1dca22a69ec/src/extensions/renderer/base/load-listeners.mjs#L1084):

```js
        if (wheelDeltas.length >= wheelDeltaN) {
          // use log to determine if inaccurate
          var wds = wheelDeltas;
          console.log("wds are", wds); // <-- new thing; log the wheel deltas
```

When there is slow initial scrolling that causes smooth future scrolling (i.e. the bug does not occur), I see the following:

```js
wds are Array(4) [ -138, -138, -138, -138 ]
```

But when when there is fast scrolling that causes jerky future scrolling (i.e. the bug occurs!), I see the following or something like it:

```js
wds are Array(4) [ -138, -276, -138, -138 ]
```

Note that 276 = 138 * 2. It looks like the values here are always multiples of `-138` -- another example is `[ -138, -552, -276, -138 ]`, and I was even able to get my laptop's touchpad to give me `[ -138, -138, -1242, -966 ]`.

My guess as to what is happening here is that, sometimes, **"fast" scroll events in a short interval are combined**. This would explain why everything is divisible by `-138`.

## Proposed solution

Under the assumption that the "base" magnitude (i.e. `138`, for my system) will always be present in the initial sampling of wheel deltas, we could replace `allAreSameMagnitude()` by just finding the minimum-magnitude value, and checking to see if all other magnitudes in the array are divisible by it.

However, this assumption may not hold (i.e. maybe `wds` is something dreadful like `[-276, -276, -1242, -966]`). To account for this pathological case, this PR instead finds the greatest common divisor (GCD) of the wheel deltas and sets this as the `inaccurateScrollFactor`.

The GCD is generally only defined for integers, and I wasn't sure if we needed to worry about non-integer wheel deltas; [MDN indicates that we might](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaY). So, this PR includes some logic to gracefully handle non-integer wheel deltas -- if any of the input wheel deltas are not integers, then the GCD function will bail out and not change anything. (I've left `allAreSameMagnitude()` in, so that if the input wheel deltas are all identical large floating-point numbers then they should still be set as the `inaccuratescrollFactor`.)

Anyway, thanks so much for your time, and hope this is helpful!

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wheel-zoom behavior on devices that report inconsistent scroll amounts by using a more robust initial wheel input sampling approach.
  * Wheel zoom now computes the zoom factor more reliably, reducing unexpected zoom speeds and improving consistency across different wheel hardware.

* **Tests**
  * Added new test coverage for the underlying GCD-based math helpers used to interpret sampled scroll input.

### DIFF
--- a/src/extensions/renderer/base/load-listeners.mjs
+++ b/src/extensions/renderer/base/load-listeners.mjs
@@ -1043,16 +1043,6 @@ BRp.load = function(){
   var inaccurateScrollDevice;
   var inaccurateScrollFactor = 100000; // base of inaccurate wheel deltas (e.g. base 5 could yield wheels of 10, 25, 50, etc.)
 
-  var allAreDivisibleBy = function( list, factor ){
-    for( var i = 0; i < list.length; i++ ){
-      if( list[i] % factor !== 0 ){
-        return false;
-      }
-    }
-
-    return true;
-  }
-
   var allAreSameMagnitude = function(list) {
     var firstMag = Math.abs(list[0]);
     for (var i = 1; i < list.length; i++) {
@@ -1081,38 +1071,25 @@ BRp.load = function(){
 
     if (inaccurateScrollDevice == null) {
       if (wheelDeltas.length >= wheelDeltaN) { // use log to determine if inaccurate
+        inaccurateScrollDevice = false;
         var wds = wheelDeltas;
-        inaccurateScrollDevice = allAreDivisibleBy(wds, 5);
-        var isfUnset = true;
 
-        if (!inaccurateScrollDevice) { // check for all large values of same or common-divisor magnitude
-          if (wds[0] > 5) {
-            var factor;
-            // an array of equal integers "x" will have a GCD of x, so in that
-            // case there is no need to have a separate "allAreSameMagnitude"
-            // check. but the GCD function only supports integers, so keeping
-            // both checks allows arrays of equal floating-point numbers
-            if (allAreSameMagnitude(wds)) {
-              factor = wds[0];
-            } else {
-              factor = math.gcdMultipleZeroIfNonInt(wds);
-            }
-            if (factor > 1) {
-              inaccurateScrollDevice = true;
-              inaccurateScrollFactor = factor;
-              // use this flag to indicate that we already have a scroll factor
-              // set -- no need to go through the list again
-              isfUnset = false;
-            }
+        if (wds[0] >= 5) {
+          var factor;
+          // an array of equal integers "x" will have a GCD of x, so in that
+          // case there is no need to have a separate "allAreSameMagnitude"
+          // check. but the GCD function only supports integers, so keeping
+          // both checks allows arrays of equal floating-point numbers
+          if (allAreSameMagnitude(wds)) {
+            factor = wds[0];
+          } else {
+            factor = math.gcdMultipleZeroIfNonInt(wds);
+          }
+          if (factor > 1) {
+            inaccurateScrollDevice = true;
+            inaccurateScrollFactor = factor;
           }
         }
-        
-        if (isfUnset && inaccurateScrollDevice) {
-          for (var i = 0; i < wds.length; i++) {
-            inaccurateScrollFactor = Math.min(wds[i], inaccurateScrollFactor);
-          }
-        }
-
         // console.log('Sampled wheel deltas:', wds);
         // console.log('inaccurateScrollDevice:', inaccurateScrollDevice);
         // console.log('inaccurateScrollFactor:', inaccurateScrollFactor);
--- a/src/extensions/renderer/base/load-listeners.mjs
+++ b/src/extensions/renderer/base/load-listeners.mjs
@@ -1053,23 +1053,6 @@ BRp.load = function(){
     return true;
   }
 
-  // https://en.wikipedia.org/wiki/Greatest_common_divisor#Euclidean_algorithm
-  var gcd = function( a, b ) {
-    if( b === 0 ) {
-      return a;
-    }
-    return gcd(b, a % b);
-  }
-
-  // Finds the GCD of a list of numbers: https://stackoverflow.com/a/4885641
-  var gcdMultiple = function( list ) {
-    var out = list[0];
-    for ( var i = 1; i < list.length; i++ ) {
-        out = gcd(out, list[i]);
-    }
-    return out;
-  }
-
   var wheelHandler = function( e ){
     var clamp = false;
     var delta = e.deltaY;
@@ -1094,7 +1077,7 @@ BRp.load = function(){
 
         if (!inaccurateScrollDevice) { // check for all large values with common divisor magnitude
           if (wds[0] > 5) {
-            var factor = gcdMultiple(wds);
+            var factor = math.gcdMultiple(wds);
             if (factor > 1) {
               inaccurateScrollDevice = true;
               inaccurateScrollFactor = factor;
@@ -1105,7 +1088,7 @@ BRp.load = function(){
           }
         }
         
-        if ( isfUnset && inaccurateScrollDevice ) {
+        if (isfUnset && inaccurateScrollDevice) {
           for (var i = 0; i < wds.length; i++) {
             inaccurateScrollFactor = Math.min(wds[i], inaccurateScrollFactor);
           }
--- a/src/extensions/renderer/base/load-listeners.mjs
+++ b/src/extensions/renderer/base/load-listeners.mjs
@@ -1053,6 +1053,16 @@ BRp.load = function(){
     return true;
   }
 
+  var allAreSameMagnitude = function(list) {
+    var firstMag = Math.abs(list[0]);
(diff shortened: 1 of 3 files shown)


fedarko:
I've added commit 4f6606c, which addresses [the LLM's comment](https://github.com/cytoscape/cytoscape.js/pull/3491#discussion_r3641125978) -- this commit removes the explicit divisibility-by-5 check (and the later loop through `wds` to figure out the minimum magnitude and set it as `inaccurateScrollFactor`) in favor of just figuring out the GCD of the wheel deltas.

So, if `wheelDeltas` is something like `[ 10, 20, 15, 100 ]`, then -- rather than detecting divisibility by 5 and then setting the minimum magnitude (10) as `inaccurateScrollFactor` -- we will now find the GCD of these magnitudes (5) and set this as `inaccurateScrollFactor`.

This makes the code simpler, and I _think_ this matches what was intended in the original design of this part of the code -- but I recognize that it does mean changing the previous behavior. Happy to revert 4f6606c if the change is not desirable :)

maxkfranz:
GCD is basically a factoring and running that potentially every frame doesn't sit right with me.

Have you profiled this in practice?

The existing solution is basically doing a factoring (for common ones) but with a cheap heuristic.

Would this change have potentially negative effects on the majority, who primarily use good trackpads (or a tablet)?  While it's great to have an improved heuristic, it's more important that it doesn't negatively affect the majority.

fedarko:
Thank you!

> GCD is basically a factoring and running that potentially every frame doesn't sit right with me.

First -- I do not think that it would be run every frame. As I understand it, `math.gcdMultipleZeroIfNonInt()` is only called either zero times or one time (per Cytoscape.js instance, if I am understanding `load-listeners.mjs` correctly). This possible call to `math.gcdMultipleZeroIfNonInt()` occurs on the wheelHandler call where:

1. `inaccurateScrollDevice == null`, and
2. `wheelDeltas.length >= wheelDeltaN`.

If both of these conditions are met, then `inaccurateScrollDevice` will then be set to something that is not `null` -- either `true` or `false`. This is visible in [lines 1074-1095 of `load-listeners.mjs` in the proposed changes](https://github.com/fedarko/cytoscape.js/blob/cb2c81e7c6902dcbb2646f57ec20acf54def4712/src/extensions/renderer/base/load-listeners.mjs#L1074-L1095) -- `inaccurateScrollDevice` is set to `false` as soon as we enter this block, and every path through this block either leaves it as `false` or sets it to `true`.

Because `inaccurateScrollDevice` is never `null` again, this means that `math.gcdMultipleZeroIfNonInt()` is only called at most once.

--------


> Would this change have potentially negative effects on the majority, who primarily use good trackpads (or a tablet)?

As far as I am aware, no. Even within that block described above (after we've finished sampling the first 4 wheel deltas), the following conditions must be satisfied for `math.gcdMultipleZeroIfNonInt()` to be called at all:

1. the magnitude of the first wheel delta is at least 5, and
2. the magnitudes of the wheel deltas are not all the same

I don't have quick access to a "good" device to test with, but based on https://github.com/cytoscape/cytoscape.js/issues/3336#issue-2881889434 ("A user on Mac just gets values between -2 and +2 and smooth zooming."), the GCD computation should not be done at all on mac systems -- the first condition would not be met, since none of the wheel deltas' magnitudes would be at least 5.

In general -- under the assumption that "good" devices should produce wheel deltas with uniform magnitudes (and/or magnitudes that are less than 5), the GCD computation will not be run for any "good" devices. I do not know enough to say if this assumption is reasonable.

--------

> Have you profiled this in practice?

I tried adding some quick logging statements using `performance.now()` at:

- the start and end of [the `wheelHandler` callback](https://github.com/fedarko/cytoscape.js/blob/cb2c81e7c6902dcbb2646f57ec20acf54def4712/src/extensions/renderer/base/load-listeners.mjs#L1056-L1173)

- the start and end of [`gcdMultipleZeroIfNonInt()`](https://github.com/fedarko/cytoscape.js/blob/cb2c81e7c6902dcbb2646f57ec20acf54def4712/src/math.mjs#L123-L133)

... and played around with the Dagre demo graph (pointing the Cytoscape.js source to the `cytoscape.min.js` output from `npm run build`).

Some very informal observations:

- Computing the GCD of the four sampled wheel deltas typically takes around 0-2 ms.

- Calls to `wheelHandler` typically take 0-2 ms (mostly 0 or 1 ms), both before and after this PR's changes.

  - Interestingly, the _first_  `wheelHandler` call (caused by the very first nudge on the scroll wheel of a Cytoscape.js instance) seems to take a bit longer for this PR's changes -- it can take up to ~6 ms (but subsequent `wheelHandler` calls are usually still around 0-2 ms).

    - My best guess is that this phenomenon is due to how this PR calls `Math.abs()` while logging a wheel delta? But I am not sure that is a sufficient explanation.

In any case, this benchmarking (plus the fact that GCD computation is a one-time cost, and is only done on certain systems) leads me to believe that this PR does not represent a performance problem. Happy to adjust the code further if desired, though!

-------

### Appendix: how long, in theory, does the one-time GCD computation take to run?

We know that `gcdMultipleZeroIfNonInt(arr)` is called at most once. If it is called, this function then calls `gcd(a, b)` at most `arr.length - 1` times (finding the GCD of `arr[0]` and `arr[1]`, then the GCD of that and `arr[2]`, etc).

`gcd()` implements the [Euclidean algorithm](https://en.wikipedia.org/wiki/Euclidean_algorithm). Wikipedia has [a whole section talking about its efficiency](https://en.wikipedia.org/wiki/Euclidean_algorithm#Algorithmic_efficiency); by [Lamé's theorem](https://en.wikipedia.org/wiki/Lam%C3%A9%27s_theorem), the maximum amount of "steps" performed by GCD(_a_, _b_) is 5 \* _k_ (where _k_ is the number of digits of _b_).

Most of the examples I've seen of wheel deltas are two or three digits. Even if we are pessimistic and assume that _k_ = 4 (all the wheel deltas have four digits), then this implies that each call to `gcd()` takes at most 20 steps. Since there are at most 3 calls to `gcd()`, there would mean that calling `gcdMultipleZeroIfNonInt()` would require there be at most 60 steps taken. (This is ignoring the logic that makes sure that each wheel delta is an integer, and quits early if not.)

Following [the third method Wikipedia gives](https://en.wikipedia.org/wiki/Euclidean_algorithm#Average), we can estimate the average-case runtime of the GCD algorithm -- if we assume that all wheel deltas are uniformly distributed in the range [1, 9999] -- as 
`12/(pi**2) * log(2) * log(9999) + 0.06 = 7.82` steps. If we assume 3 `gcd()` calls , then this means that `gcdMultipleZeroIfNonInt()` would require roughly `23.47` steps on average.

... Of course, wheel delta magnitudes are not uniformly distributed (e.g. something like `[138, 276, 138, 138]` is more realistic on my system), meaning that in practice GCD computation should require relatively few steps :)

maxkfranz:
Excellent points.  Let's merge this in

